PR #15, opened 2018-11-30
Introducing TravisCI.

davidbarsky:
*Description of changes:*

- Added a .travis.yml, which allows us to run CI on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,8 @@ matrix:
   - rust: nightly
     env: ALLOW_FAILURES=true
   - os: osx
-  - env: TARGET=x86_64-unknown-freebsd
-  - env: TARGET=i686-unknown-freebsd
-  - env: TARGET=i686-unknown-linux-gnu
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+  - rust: 1.30.0
+  - rust: stable
+  - rust: beta
+  - rust: nightly
+    env: ALLOW_FAILURES=true
+  - os: osx
+  - env: TARGET=x86_64-unknown-freebsd
+  - env: TARGET=i686-unknown-freebsd
+  - env: TARGET=i686-unknown-linux-gnu
+script:
```
sapessi: Should we have the `x86_64-unknown-linux-gnu` env here? That's the target we should be using for Lambda

davidbarsky:
Uh, yeah. I'll add x86_64-unknown-linux-gnu and the status to the README.md.

.travis.yml
davidbarsky: Done.

PR #16, opened 2018-11-30
Have TravisCI cache cargo between builds.

davidbarsky:
*Description of changes:*

see title.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,11 @@
 language: rust
-cache: cargo
+
+cache:
+  directories:
+    - /home/travis/.cargo
+
+before_cache:
+  - rm -rf /home/travis/.cargo/registry
 
 matrix:
   include:


softprops:
You may want to keep in mind some of the caveats identied of this approach listed in [this post](https://levans.fr/rust_travis_cache.html)

davidbarsky:
Thanks @softprops! That motivated me to not have the the cache initially, but then I noticed Travis took ~30 minutes to run, whereas other projects I worked on (Tokio Trace, for one) took _maybe_ 2 minutes. The only difference was the presence of cached Travis CI. I’m open to re-evaluating this choice, however!

davidbarsky:
Hmm, the issue seems to be related to the fact that we're building for macOS. It might make sense to skip, IMO.

sapessi:
Yeah. No need to build for macos

softprops:
It's not about not caching but what to cache. There's some notes in the post on more focused dirs to cache.

davidbarsky:
Thanks @softprops. I've added the caching strategy that the article recommended.

Issue #13, opened 2018-11-29
gateway bindings


srijs:
Hi!

It's great to see official lambda support for Rust!

Over at https://github.com/srijs/rust-aws-lambda, we have been building a lambda runtime and various auxiliary crates. We will start to redirect users of the runtime parts over to the official runtime, but similar to #12 I'm keen to hear whether there is interest on your side to move some of the other auxiliary crates.

In particular, there are bindings that translate between the API Gateway proxy events and the types provided by the `http` crate. I'm happy to continue to provide these myself, but I'd like to avoid any duplication of effort, so if you'd like to incorporate these please let me know!

davidbarsky:
> Over at https://github.com/srijs/rust-aws-lambda, we have been building a lambda runtime and various auxiliary crates. We will start to redirect users of the runtime parts over to the official runtime, but similar to #12 I'm keen to hear whether there is interest on your side to move some of the other auxiliary crates.

> In particular, there are bindings that translate between the API Gateway proxy events and the types provided by the http crate. I'm happy to continue to provide these myself, but I'd like to avoid any duplication of effort, so if you'd like to incorporate these please let me know!

We'd be super happy to accept any auxiliary crate from rust-aws-lambda! I'd also like to explore how we can support how Lambda accepts HTTP requests as a target of an [Application Load Balancer](https://aws.amazon.com/blogs/networking-and-content-delivery/lambda-functions-as-targets-for-application-load-balancers/). I suspect the interaction would be pretty similar.

phillyfan1138:
So does this official library not support http events yet?

sapessi:
No @phillyfan1138, this library only takes care of interacting with the Runtime APIs and start your handler function. To handle API Gateway proxy events or ELB requests you'd need to provide your structs that mirror the event types. Take a look at [this crate](https://github.com/srijs/rust-aws-lambda/tree/master/aws_lambda_events). We are exploring taking this over in the future (#12)

srijs:
Yay, good job @softprops!

Issue #12, opened 2018-11-29
Event struct definitions


LegNeato:
Hello! I am one of the maintainers of https://github.com/srijs/rust-aws-lambda (Rust interface to the Go runtime). As part of that project, we have lambda event definitions that don't depend on the runtime:

https://github.com/srijs/rust-aws-lambda/tree/master/aws_lambda_events

It would be great to have those event types integrated in this crate, preferably by re-exporting. I'm also fine if you want to take over ownership and make it officially supported, as long as it doesn't take too long for changes to be merged.

Note right now the generated structs take `String` and thus may require more allocations. I plan to also have variants that take `Cow<'a, str>`.

softprops:
Few ideas I wanted to throw out there because this same topic must be opened for each and every type checked language used in lambda in the past and likely every type checked language implemented in the future.

Something I'd considered a few months back was kicking around the idea putting together a repository that collecting machine readable json schemas documenting and describing AWS lambda types. https://github.com/softprops/typed-lambda

It's just an idea at this point and it's incomplete ( as you can see from it's Travis build :) ). The first attempt was to use a go package to read the go structs and output Json schemas that could in theory be used to regenerate the go and any other type checked language AWS lambda structures.

In practice I found the go definitions to produce imprecise information. For instance in go, zero values (nil) are useful and don't translate information to a language like rust. 

My next exploration was going to be translating this via  typescript types which express type information more precisely to Json schemas https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/aws-lambda/index.d.ts

I wanted to drop a heads up in case you folks would be interested in collaborating to generate rust types from Json schemas as well. I see this kind of thing have a benefit to a  wider audience as other language sdks would also benefit.

sapessi:
Hey @softprops, I'll meet with the Lambda team next week to figure out what we have internally and could use for this. I'd try to rely on those definitions since they are the ones the Lambda team is likely to keep up to date for us. I'll update this issue once I have something.

alexreg:
Any progress on this lately?

davidbarsky:
@alexreg It's something that the Lambda team is looking at, but unfortunately, I can't give you a specific date as to when it'll be ready. In the meantime, I suggest using the [`aws_lambda_events`](https://crates.io/crates/aws_lambda_events) crate, as that is the most comprehensive source of typed CloudWatch Events currently available.

We'll update this ticket when official support is available.

[...]

PR #18, opened 2018-11-30
introduce lambda-http module based on lando to support apigateway

softprops:
*Issue #, if available:*

This attempts to address #13 

*Description of changes:*

## Motivation

With this project announced at reinvent, I saw an opportunity to rebase [lando](https://github.com/softprops/lando) a crate providing [http crate](https://crates.io/crates/http) interfaces for api gateway for deployment on lambda. I started working on that then I stumbled on #13. These crates seem to all serve the same purpose and landos goal from the start was to `not to replace tools but instead to work well with them`. I saw this work going down the path of fracturing when I'd rather push it forward.

## Impl

I just took what I was working on in lando's in migrating towards rebalancing on top of this project and moved into a module _in_ this project.

I took some steps to remove a few dependencies lando had that I think we're actually needed, but my first attempt forced a change to the `lambda_runtime::start` api which introduced a single allocation in order to support a serde deserializable => `http::Request` => `http::Response` => serde serializable transformation where the user space dealt only in terms of  `http::{Request, Response}`'s . I couldn't find a way to avoid it with the lambda_runtime's function pointer interface but I'm open to change/suggestions.

The `RequestExt` type contains one opinionated method `payload` which was lando's convenience for transparently deserializing request bodies from form encoded _or_ application/json content. I'm open to pulling this out but I wanted to get the temperature in the room before I did

There's also an opinionated dependency on the `failure` crate. the failure crate mainly facilitated removing Error definition boilerplate for the `payload` methods errors. I'm also open to ditching this dependency on impl the boilerplate by hand to keep the dependency tax low

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.  ✅

### DIFF
--- a/lambda-http/Cargo.toml
+++ b/lambda-http/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "lambda_http"
+version = "0.1.0"
+authors = ["Doug Tangren"]
+description = "Rust API Gateway interfaces for AWS Lambda"
+keywords = ["AWS", "Lambda", "Runtime", "Rust"]
+license = "Apache-2.0"
+homepage = "https://github.com/awslabs/aws-lambda-rust-runtime"
+repository = "https://github.com/awslabs/aws-lambda-rust-runtime"
+documentation = "https://docs.rs/lambda_runtime"
+
+[dependencies]
+http = "0.1"
+serde = "^1"
+serde_json = "^1"
+serde_derive = "^1"
+lambda_runtime = { path = "../lambda-runtime", version = "^0.1" }
+tokio = "^0.1"
+base64 = "0.10"
+failure = "0.1"
+failure_derive = "0.1"
+serde_urlencoded = "0.5"
--- a/lambda-http/examples/basic.rs
+++ b/lambda-http/examples/basic.rs
@@ -1,0 +1,31 @@
+extern crate lambda_runtime as lambda;
+extern crate lambda_http;
+extern crate log;
+extern crate simple_logger;
+
+use lambda_http::{lambda, Request, Response};
+use lambda::{error::HandlerError, Context};
+
+use log::error;
+use serde_derive::{Deserialize, Serialize};
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    simple_logger::init_with_level(log::Level::Debug).unwrap();
+    lambda!(my_handler);
+
+    Ok(())
+}
+
+fn my_handler(e: Request, c: Context) -> Result<Response, HandlerError> {
+    let name = e.query_string_parameters().get("first_name");
+    Ok(match name {
+      Some(first_name) => {
+        Response::new(format!("Hello, {}!", first_name).into())
+      },
+      _ => {
+        error!("Empty first name in request {}", c.aws_request_id);
+        Response::builder(400).body("Empty first name").expect("failed to render response").into()
+      }
+    })
+}
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -1,0 +1,103 @@
+#![warn(missing_docs)]
+//#![deny(warnings)]
+//! Enriches `lambda_runtime` with `http` types targeting API Gateway proxy events
+//!
+//! # Example
+//!
+//! ```rust,no_run
+//! #[macro_use] extern crate lambda_http;
+//! extern crate lambda_runtime as lambda;
+//!
+//! use lambda::{Context, HandlerError};
+//! use lambda_http::{Request, Response, RequestExt};
+//!
+//! fn main() {
+//!   lambda!(handler)
+//! }
+//!
+//! fn handler(
+//!   request: Request,
+//!   ctx: Context
+//! ) -> Result<Response, HandlerError> {
+//!   Ok(
+//!     Response::new(
+//!       format!(
+//!         "hello {}",
+//!         request.query_string_parameters()
+//!           .get("name")
+//!           .unwrap_or_else(|| "stranger")
+//!       ).into()
+//!     )
+//!   )
+//! }
+//! ```
+extern crate base64;
+extern crate failure;
+#[macro_use]
+extern crate failure_derive;
+extern crate http;
+extern crate lambda_runtime;
+extern crate serde;
+#[macro_use]
+extern crate serde_derive;
+extern crate serde_json;
+extern crate serde_urlencoded;
+extern crate tokio;
+
+use lambda_runtime::{self as lambda, Context};
+use lambda_runtime::error::HandlerError;
+use http::{Request as HttpRequest, Response as HttpResponse};
+use tokio::runtime::Runtime as TokioRuntime;
+
+mod ext;
+pub mod request;
+mod response;
+mod body;
+mod strmap;
+
+use request::GatewayRequest;
+use response::GatewayResponse;
+pub use ext::RequestExt;
+pub use strmap::StrMap;
+pub use body::Body;
+
+/// Type alias for `http::Request`s with a fixed `lambda_http::Body` body
+pub type Request = HttpRequest<Body>;
+
+/// Type alias for `http::Response`s with a fixed `lambda_http::Body` body
+pub type Response = HttpResponse<Body>;
+
+/// Functions acting as API Gateway handlers must conform to this type.
+pub type Handler = fn(Request, Context) -> Result<Response, HandlerError>;
+
+/// lifts private internal (de)serializable API Gateway proxy event types into `http` types
+struct Lift(Handler);
+
+impl <'a> lambda::Handler<GatewayRequest<'a>, GatewayResponse> for Lift {
+    fn handle(&self, req: GatewayRequest, ctx: Context) -> Result<GatewayResponse, HandlerError> {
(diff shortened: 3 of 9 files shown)


lambda-http/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+
+/// Creates a new `lambda_runtime::Runtime` and begins polling for API Gateway events
+///
+/// # Arguments
+///
+/// * `f` A type that conforms to the `Handler` interface.
+///
+/// # Panics
+/// The function panics if the Lambda environment variables are not set.
+pub fn start(f: Handler, runtime: Option<TokioRuntime>) {
```
softprops: I took the path of going for module name spacing vs method name spacing here i.e. `lambda_http::start` vs `start_http`

lambda-http/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "lambda_http"
+version = "0.1.0"
+authors = ["Doug Tangren"]
```
softprops: I'm not really particular on what you want to put here.

lambda-http/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+description = "Rust API Gateway interfaces for AWS Lambda"
+keywords = ["AWS", "Lambda", "Runtime", "Rust"]
```
sapessi: We should probably have AWS, Lambda, Rust, APIGateway (or API, Gateway) here.

lambda-http/src/request.rs
```diff
@@ -0,0 +1,259 @@
...
+/// Representation of an API Gateway proxy event data
+#[doc(hidden)]
+#[derive(Deserialize, Debug, Default)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct GatewayRequest<'a> {
+    pub(crate) path: Cow<'a, str>,
+    #[serde(deserialize_with = "deserialize_method")]
+    pub(crate) http_method: Method,
+    #[serde(deserialize_with = "deserialize_headers")]
+    pub(crate) headers: HeaderMap<HeaderValue>,
```
sapessi: You should consider using API Gateway's new multi value headers and query string properties. See announcement [here](https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-api-gateway-adds-support-for-multi-parameters/)

lambda-http/src/response.rs
```diff
@@ -0,0 +1,100 @@
...
+/// Representation of API Gateway response
+#[derive(Serialize, Debug)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct GatewayResponse {
+    pub status_code: u16,
+    #[serde(
+        skip_serializing_if = "HeaderMap::is_empty",
+        serialize_with = "serialize_headers"
+    )]
+    pub headers: HeaderMap<HeaderValue>,
```
sapessi: Same here on multi-value headers.

lambda-http/src/lib.rs
```diff
@@ -0,0 +1,103 @@
...
+pub use body::Body;
+
+/// Type alias for `http::Request`s with a fixed `lambda_http::Body` body
+pub type Request = HttpRequest<Body>;
+
+/// Type alias for `http::Response`s with a fixed `lambda_http::Body` body
+pub type Response = HttpResponse<Body>;
+
+/// Functions acting as API Gateway handlers must conform to this type.
+pub type Handler = fn(Request, Context) -> Result<Response, HandlerError>;
```
davidbarsky: We _just merged a change by that allows people to use closures as lambda handler. Could you rebase on those changes?

lambda-http/examples/basic.rs
```diff
@@ -0,0 +1,31 @@
...
+    simple_logger::init_with_level(log::Level::Debug).unwrap();
+    lambda!(my_handler);
+
+    Ok(())
+}
+
+fn my_handler(e: Request, c: Context) -> Result<Response, HandlerError> {
+    let name = e.query_string_parameters().get("first_name");
+    Ok(match name {
+      Some(first_name) => {
```
davidbarsky: Looks like it’s two spaces. Do you mind running rustfmt on this? I’ll have a PR that runs rustfmt on CI.

lambda-http/Cargo.toml
davidbarsky: Eh, not sure either. This will be fine for now, I think.

lambda-http/src/lib.rs
davidbarsky: Nice!

lambda-http/src/request.rs
softprops: will do, these were just pulled over from lando as-and it didn't support that. adding support should be straight forward. ill push changes up for this soon

davidbarsky: Thanks!

lambda-http/examples/basic.rs
softprops: ok. before I do, are we only doing this for the nightly version? here's what I get when I run on stable


```sh
cargo fmt --all                                                               ✘ 101
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
Warning: can't set `merge_imports = true`, unstable features are only available in nightly channel.
```

I'll run  `cargo +nightly fmt --all` for now. that seemed to only change the http module

lambda-http/src/request.rs
softprops: clarification _headers_ will be straightforward because [http::HeaderMap](https://docs.rs/http/0.1.5/http/struct.HeaderMap.html) already abstracts over that easily. let me think about query_string_values. StrMap, currently used for query_string_parameters and path_parameters was a thing I put into place to make ergonomics of ownership with these maps easier to deal with. let me revisit that with query string parameters.

traveling a bit today. Ill see if I get get to this tonight. otherwise if its not a deal breaker I'd be happy to follow up to add support.

lambda-http/src/lib.rs
softprops: yep! I saw that. taking advantage of that now ill mirror that with a the trait/impl pattern. I'll look at this tonight

lambda-http/examples/basic.rs
softprops: done

davidbarsky: yeah, let's maybe not have nightly only runs. I'm working on a fix for that.

lambda-http/src/lib.rs
davidbarsky: No worries, travel safe!

lambda-http/src/request.rs
davidbarsky: > traveling a bit today. Ill see if I get get to this tonight. otherwise if its not a deal breaker I'd be happy to follow up to add support.

I'm inclined to say that the query string business can be held off for a bit. That can be addressed in a separate PR that'll support query string params. It's not as if we're releasing a 0.2 _right_ now, and anybody using this project via git in cargo knew what they're getting themselves into.

sapessi: Yes. It can probably wait, particularly because we are likely to move the basic proxy request/response types to a separate events crate (see #12)

softprops: something to note here about these types is that they are optimized for use with the `http` crate. example like deserializing directly deserializing directly to `http::{HeaderMap,Method}` rather to Stringly types then copying into `http` crate types

lambda-http/src/lib.rs
softprops: updated

davidbarsky:
Small thing @softprops—can you fix the build issues on https://travis-ci.org/awslabs/aws-lambda-rust-runtime/jobs/462147745#L1072? Otherwise, this looks good to me.

softprops:
Will grab this tonight, Bali time :)

softprops:
All set

phillyfan1138:
I've successfully implemented this PR in my app:

https://github.com/phillyfan1138/option_price_faas

It works great!  One suggestion: would it be possible to re-export the crate http?  I had to add it to the project as a dependency to get the code to compile (see, for example, https://github.com/phillyfan1138/option_price_faas/blob/master/src/bin/output_constraints.rs)

softprops:
@phillyfan1138 will do. I tried to keep this as keep some corners from lifted lando kept conservative initially. I too found this annoying in practice, much as you typically need to pull in `futures` crate for the majority of crates that expose `futures::{Future,Stream}` types in their public API.  In lando I ended up [re-exporting with DX in mind](https://github.com/softprops/lando/blob/22f6fc690fc7e915ad4577b6fde775a876b49ad4/src/lib.rs#L106-L108) I'll do that. I'd suggest we start filing things like this as follow ups. I'm a fan of iteration. I'm already thinking there are other things I'd like to add but I want to iterate on it to see if I'll actually need it before I add it

softprops:
@davidbarsky let me know if there's anything left you'd like to see in this pull. I'd start to.start iterating on top of this in smaller pulls

PR #19, opened 2018-11-30
Support closures as handlers

srijs:
*Description of changes:*
Currently, the `Handler` type does not permit closures. This is a small refactor that enables the use of closures as handlers.


davidbarsky:
(Really happy because this should allow us to support Tower's service in a similar way.)

PR #22, opened 2018-12-01
FunctionArn -> TraceID to get xray_trace_id

anmonteiro:
*Description of changes:*
When getting the `xray_trace_id`, the client is instead looking for the `function_arn` header, which is probably the result of copy pasting and not changing one extra place.


davidbarsky:
Nice catch, thanks!

Issue #23, opened 2018-12-01
Rust ergonomics for error handling


arienmalec:
The examples for using the Rust Lambda runtime require additional calls around error handling that require access to the `Context` object.

Unfortunately, that breaks Rust ergonomics for error handing, that ordinarily would use `?` and map custom errors to `HandlerError` through an `impl From` (which doesn't have a reference to the `Context` struct.

Ideally, I could return my own error types, and either have the runtime take care of supplying context, with perhaps some place to inject logging, etc. into the error handing flow.

In practice, in the short term, I'm breaking out all actions that can return a `Result` into a separate module, because even trivial activities get complicated quickly if you want to inject error handing with access to `Context`.

davidbarsky:
Thanks for the feedback—we really appreciate it! I'm thinking of replacing the current setup with something that leverages (maybe) Failure and (for sure) ErrorKinds, and using something similar to what Sean proposed in https://github.com/tower-rs/tower/issues/131. Would a system like this—behind a breaking change semver bump—be something like you're looking for?

arienmalec:
I think so. In my head, probably poorly thought out, the handler type would be `(U,Context) -> Result<T,<Box<dyn Error>>`, and the lambda macro would have an (optional?) second hook for logging/cleanup.

An example of how I'm doing this, loosely based on `map_err`:

```rust
fn map_handler_err(msg: &str, ctx: &Context, err: impl Error) -> HandlerError {
    error!("{} in request {}",msg, ctx.aws_request_id);
    ctx.new_error(&err.to_string())
}

fn my_handler(event: MsgEvent, ctx: Context) -> Result<MsgResponse, HandlerError> {
    match record::Record::parse_base64(&event.msg) {
        Ok(rec) => {
            let s = rec.summary();
            Ok(MsgResponse{id: s.id, sum: s.sum, longest: s.longest})
        },
        Err(e) => Err(map_handler_err("Message parsing error", &ctx, e))
    }
}
```

Which suggests a non-breaking old-style Rust error handling hack: a trait with a `map_err_ctx` or some such, so that I could rewrite the handler to:
```rust
let r = record::Record::parse_base64(&event.msg).map_err_ctx(my_err_handler)?;
let s = r.summary();
Ok(MsgResponse{...}
```

sapessi:
[...]

PR #26, opened 2018-12-03
Handle any errors

sapessi:
*Issue #, if available:* #23 

*Description of changes:* 
Beginning to address generic error handling in the runtime. Changed the handler to return a `Box<dyn Error>`. This is still a work in progress.


lambda-runtime/src/lib.rs
```diff
@@ -11,8 +11,10 @@
 //! extern crate serde_derive;
 //! #[macro_use]
 //! extern crate lambda_runtime;
+//! #[macro_use]
+//! extern crate simple_error;
```
davidbarsky: I'm not sure how many people use `simple_error`. Should we opt for Failure for popularity reasons?

lambda-runtime/examples/custom_error_failure.rs
```diff
@@ -0,0 +1,47 @@
+extern crate lambda_runtime as lambda;
+extern crate log;
+extern crate serde_derive;
+extern crate simple_logger;
+#[macro_use]
```
davidbarsky: We don't need to have `#[macro_use]` in Rust 1.30 and above—they can be imported like normal macros. Given that the 2018 edition is coming out on Thursday, I think it's a good idea to be forward-facing with regard to these imports.

lambda-runtime/src/lib.rs
sapessi: As per our conversation, I'm keeping the simple error crate in the docs example for simplicity's sake. I've removed the `#[macro_use]` and added an example specific to the failure crate

lambda-runtime/examples/custom_error_failure.rs
sapessi: Fixed all over the place.

lambda-runtime/src/error.rs
```diff
@@ -1,11 +1,35 @@
...
 
-use backtrace;
-use lambda_runtime_client::error;
+use lambda_runtime_client::error::ApiError;
 use serde_json;
 
+/// The `HandlerError` struct can be use to abstract any `Err` of the handler method `Result`.
+/// The `HandlerError` object can be generated `From` any object that supports `Display`,
+/// `Send, `Sync`, and `Debug`. This allows handler functions to return any error using
+/// the `?` syntax. For example `let _age_num: u8 = e.age.parse()?;` will return the
```
softprops: since this comment is likely going to be read from the perspective of someone reading the rustdoc without much context for the example this is lifted from in the repo, you may just want to give a more isolated example like  `"not a u8".parse::<u8>()?`

softprops:
wanted to give you a quick heads up about a pull that may conflict with this but it seems there are a number of merge conflicts already present. In anycase in https://github.com/awslabs/aws-lambda-rust-runtime/pull/50, I'm proposing moving away from stringly values for ErrorResponse#errorType

sapessi:
Closing this PR in favor of the refactor in #63

Issue #24, opened 2018-12-02
Thoughts about deadline API


Sh4rK:
First, there is some inconsistency about what the deadline values mean: here
https://github.com/awslabs/aws-lambda-rust-runtime/blob/109a54770f25830fdd663a2d4cd46504bd72b1d0/lambda-runtime/src/context.rs#L62-L64
and here
https://github.com/awslabs/aws-lambda-rust-runtime/blob/109a54770f25830fdd663a2d4cd46504bd72b1d0/lambda-runtime-client/src/client.rs#L101-L102
the comments state that the timestamp is in nanoseconds, but here (and in the tests below)
https://github.com/awslabs/aws-lambda-rust-runtime/blob/109a54770f25830fdd663a2d4cd46504bd72b1d0/lambda-runtime/src/context.rs#L103-L108
it is used as if it's actually milliseconds. Which one is correct?

Second, I don't think we need `u128` for the deadlines, even if they are in nanoseconds; an `u64` would last until the year 2554, and even an `i64` until the year 2262. Also, we should probably use a signed int (`i64`) for the return value of `get_time_remaining_millis`. What do you think?

I can open a PR for these if we come to any conclusions.

 (Sorry for using "we" when I clearly haven't done anything for the project yet, it was the easiest to phrase it this way.)

sapessi:
Agree on u64. The deadline is actually millis - used to be nanos. I forgot to change the comments. Love the "we" - we've been blown away by the community engagement.

Sh4rK:
Closed accidentally.

Sh4rK:
So with #27 merged, I've been thinking about directly exposing the deadline, either in milliseconds or as a `chrono` datetime value. However, I don't actually need that (right now), it's more like a "why not?" idea, so feel free to close this issue.

coltonweaver:
@Sh4rK let us know if you still feel strongly about this issue. In the meantime I'm going to close this, but feel free to reopen it.

PR #27, opened 2018-12-03
Convert deadline-related values to i64 and fix comments.

Sh4rK:
*Issue #, if available:* #24

*Description of changes:*
Convert deadline-related values to i64 and fix comments.
I chose to use `i64` for both the deadline and the remaining time for ease of use and correctness with subtractions. As a precedent, `chrono::Utc::now().timestamp_millis()` also returns an `i64`.

### DIFF
--- a/lambda-runtime-client/src/client.rs
+++ b/lambda-runtime-client/src/client.rs
@@ -22,7 +22,7 @@ pub enum LambdaHeaders {
     FunctionArn,
     /// The X-Ray trace ID generated for this invocation
     TraceId,
-    /// The deadline for the function execution in nanoseconds
+    /// The deadline for the function execution in milliseconds
     Deadline,
     /// The client context header. This field is populated when the function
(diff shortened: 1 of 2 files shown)


lambda-runtime-client/src/client.rs
```diff
@@ -395,14 +395,13 @@ impl RuntimeClient {
             }
         };
 
-        let deadline: u128 = match headers.get(LambdaHeaders::Deadline.to_string()) {
-            Some(value) => value.to_str()?.to_owned(),
+        let deadline = match headers.get(LambdaHeaders::Deadline.to_string()) {
```
sapessi: We just merged #28 that introduced the `as_str()` method on the headers to avoid unnecessary allocations. Could you switched to as_str here.

Sh4rK: Done (rebased).

PR #41, opened 2018-12-10
`ClientApplication` keys should be snake_case

anmonteiro:
See https://github.com/aws/aws-lambda-go/blob/9e3676ee8ca83aee500682f382e10b9d03660093/lambdacontext/context.go#L43-L48

### DIFF
--- a/lambda-runtime-client/src/client.rs
+++ b/lambda-runtime-client/src/client.rs
@@ -60,19 +60,14 @@ impl fmt::Display for LambdaHeaders {
 #[derive(Deserialize, Clone)]
 pub struct ClientApplication {
     /// The mobile app installation id
-    #[serde(rename = "installationId")]
     pub installation_id: String,
     /// The app title for the mobile app as registered with AWS' mobile services.
-    #[serde(rename = "appTitle")]
     pub app_title: String,
     /// The version name of the application as registered with AWS' mobile services.
-    #[serde(rename = "appVersionName")]
     pub app_version_name: String,
     /// The app version code.
-    #[serde(rename = "appVersionCode")]
     pub app_version_code: String,
     /// The package name for the mobile application invoking the function
-    #[serde(rename = "appPackageName")]
     pub app_package_name: String,
 }
 


softprops:
It may be helpful to prevent future regressions of this sort, to provide an example json payload where you experienced an error, and a unit test validating the change you made to ensure this itself wont introduce new defects

softprops:
Here's an example of how we test this kind of thing from the [lambda-http](https://github.com/awslabs/aws-lambda-rust-runtime/blob/838d130c9a7c2a948f749987292b579b5d9e5ba4/lambda-http/src/request.rs#L221-L227) crate. The unit test loads in an external [flat json file](https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-http/tests/data/proxy_request.json) containing the data we expect to parse and assert that it parses without error

anmonteiro:
Thanks both for your reviews. I just opened this PR after doing the necessary research to define the types in my [OCaml Lambda runtime](https://github.com/anmonteiro/aws-lambda-ocaml-runtime).

I am, however, not a Rust user, and I suspect it would take me some time to get comfortable with the setup and toolchain to provide the tests you suggest. I'd rather close this PR than doing the necessary work to get it through.

Apologies to all for having wasted your time. If you want to pick this where I left it, please do.

davidbarsky:
@anmonteiro You haven't wasted any of our time! I greatly appreciate your assistance here, and the testing is pretty similar to ReasonML with Jest (compared to last I used it...). If you're curious, this is that the test would look like: https://play.integer32.com/?version=stable&mode=debug&edition=2018&gist=3d37b76ae47813ab82614cb6552a84f0.

In the PR, the `fn deserialize_client_application()` test can be placed right next to the `ClientApplication` definition, and running the test consists of `cargo test`.

davidbarsky:
Ah, I see you're using `esy`—the test work is pretty similar. Instead of `esy make test`, it'll `cargo test`. Up to you if you want to land that, but I'm very sorry if we made you feel discouraged.

anmonteiro:
@davidbarsky thanks for the examples. I'll take a stab at it soon.

anmonteiro:
Not going to get to this.

PR #43, opened 2018-12-11
impl PartialEq for HandlerError

softprops:
*Issue #, if available:*

*Description of changes:*

While making some attempts to sketch out what unit testing might look like for this project I quickly realized that the core user interface, [Handler](https://github.com/awslabs/aws-lambda-rust-runtime/blob/70e7301fc7a778d011c95455ba099a9a88d46d32/lambda-runtime/src/runtime.rs#L14-L18), does not lend itself well to the out of the box testing interfaces, like `assert_eq!` because Handler results can yield HandlerErrors which do not implement PartialEq. I realize HandlerError is in a bit [in flux](https://github.com/awslabs/aws-lambda-rust-runtime/pull/26) but I think we should still have a basic store for unit testability with Rust's default toolchain.


davidbarsky:
Thanks! This is a solid fix.

PR #50, opened 2018-12-17
make ErrorResponse#errorType efficient and impossible to mis-represent

softprops:
*Issue #, if available:*

*Description of changes:*

While studying the source of this project I came across of few improvements that I thought would be helpful to spread across independent pull requests. In this case I noticed that `ErrorResponse#errorType` was represented as a string though it's values as it seems today, are enumerable. In cases like these, especially for public fields, it makes it possible to represent invalid values. As such its [error prone to work with](https://github.com/awslabs/aws-lambda-rust-runtime/pull/30/files) :). There's another optimization here is that with enum values, there's no need to allocate new strings when building the errorType value of the ErrorResponse object.

Included in addition to the changes are tests to validated them.

@davidbarsky Here's a few things interesting things I noticed. 

1) By habit I'm used to running `cargo +nightly fmt --all` after making changes. I noticed in this case that increased the side of this diff because it seems it hasnt been run recently and has affected other unrelated files.  I wanted to get your input here. Should I just reformat those in this pull. An alternative that I'd prefer is opening a pull that does that + adds `cargo fmt --all -- check`  to pull to prevent the introduction of adhoc style fmting. Thoughts?

2) `ErrorResponse#stacktrace` serializes to `null` by default. There's a link to the runtime open api spec [here](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html). Though I can't speak to its accuracy and currecness it seems a bit unclear which fields are required and which aren't in the  ErrorResponse definition. 

in the spec it says meaning all fields are optional according to openapi

```yaml
    ErrorResponse:
      type: object
      properties:
        errorMessage:
          type: string
        errorType:
          type: string
```

however... in the same doc there's a comment at the top that describes ErrorResponses as having a `stackTrace` ( with an empty value )

```yaml
# Runtimes should report all errors using Lambda standard error format, in order to integrate with other AWS services:
#
# Content-Type: application/vnd.aws.lambda.error+json:
# {
#     "errorMessage": "...",
#     "errorType": "...",
#     "stackTrace": [],
# }
#
```

You may want to pass this feedback long to the the lambda team if you've got the right communication channels open :)

With serde you can instruct it to not serialize a value when not present or "empty" It's unclear what this should be but I thought it was out of scope for my change. I'm happy to adfress here but just as happy to address in a follow up or track with new gh issue.


softprops:
It's worth calling out I discovered the stackTrace thing while writing a unit test. It makes me think there's value in upping code coverage a bit for the sake of maintaining integrity across changes

lambda-runtime-client/src/error.rs
```diff
@@ -182,3 +182,24 @@ impl RuntimeApiError for ApiError {
...
+
+#[cfg(test)]
+mod tests {
+    use super::ErrorResponse;
+
+    #[test]
+    fn handled_error_response_serializes() {
+        assert_eq!(
+            serde_json::to_string(&ErrorResponse::handled("💀")).expect("failed to serialize ErrorResponse"),
+            r#"{"errorMessage":"💀","errorType":"Handled","stackTrace":null}"#
```
softprops: unrelated, but note the json null ( serde default for Option::None ). we may want to catch this in a follow up if that's not intentional

lambda-runtime-client/src/error.rs
```diff
@@ -61,10 +61,10 @@ impl ErrorResponse {
     ///
     /// # Return
     /// A populated `RuntimeError` object that can be used with the Lambda Runtime API.
-    pub fn unhandled(message: String) -> ErrorResponse {
+    pub fn unhandled<M>(message: M) -> ErrorResponse where M: Into<String> {
```
softprops: happen to back out of this but in general this is fmt is typically more ergonomic in rust as it now accepts anything that implements `Into<String>` like slices and the like

softprops:
addressing the rustfmt thing over in https://github.com/awslabs/aws-lambda-rust-runtime/pull/51

lambda-runtime-client/src/error.rs
```diff
@@ -9,30 +9,30 @@ use hyper;
...
+    /// unexpcted errors.
+    Unhandled,
+}
 
 /// This object is used to generate requests to the Lambda Runtime APIs.
 /// It is used for both the error response APIs and fail init calls.
 /// custom error types should implement the `RuntimeError` trait and return
 /// this object to be compatible with the APIs.
 #[derive(Serialize)]
+#[serde(rename_all="camelCase")]
```
softprops: this removes some potential for error prone manual string formatted by using a feature built into serde that will do the formatting for you resulting in less code and less error-proneness.

davidbarsky: Good call, thanks.

davidbarsky:
Merged #51.

davidbarsky:
> It's worth calling out I discovered the stackTrace thing while writing a unit test. It makes me think there's value in upping code coverage a bit for the sake of maintaining integrity across changes

Yes, agreed. I've been working on a clean-room implementation of the runtime (with unit tests) to see if I can discover any blindspots. I'll have a PR for that up shortly (or at the very least, a separate repo that you can view).

lambda-runtime-client/src/error.rs
davidbarsky: > happen to back out of this but in general this is fmt is typically more ergonomic in rust as it now accepts anything that implements Into<String> like slices and the like.

Nice catch, agreed. we were a little pressed before release, so it's nice that you're catching all these little gotchas that we missed first time around...

lambda-runtime-client/src/error.rs
davidbarsky: Unsure, I've reached out to the lambda runtime team about this.

sapessi:
Hey @softprops, hold off on these changes. Same thing for my PR (#26) - it was just an experiment. We are likely to change the way errors are handled for the next release of the framework. We are discussing this internally with the Lambda team too. I will open an RFC issue about the proposed methodology this week.

sapessi:
See #54

softprops:
closing this pull for now to keep the merge queue closed. thanks for the link!

PR #51, opened 2018-12-17
ensure a consistent style for all changes

softprops:
*Issue #, if available:*

*Description of changes:*

I found in another pull that I wasn't able to fmt sources without incurring a bigger git diff of unrelated files. Formatting files is actually good for contributors because its often a feature of rust's efforts to improve the quality of life tools like rls and its vscode extension ( which auto formats on save ). Consistency is also good for readability. 

Included is 

* an updated travis config adds a nightly check ( the rustfmt config this project uses requires nightly ) for rustfmt 
* an update to contributing file to give contributors a heads up
* some helpful links for those new to the project ( and rustfmt ) what these style configs rules do

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,2 +1,4 @@
+# https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#merge_imports
 merge_imports = true
+# https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#max_width
 max_width = 120
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
 script:
   - |
     if [[ "$TRAVIS_RUST_VERSION" == nightly ]]; then
-      cargo +nightly fmt --all -- --check
+      cargo fmt --all -- --check
     fi
   - cargo build --verbose --all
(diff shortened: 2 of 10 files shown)


softprops:
going to wait on travis before actually applying the style rules to see what happens in the case the style rules _aren't_ honored

softprops:
check seemed to fail without fmting so that worked

screen shot 2018-12-18 at 12 31 16 am

screen shot 2018-12-18 at 12 32 49 am

softprops:
noting here that the ALLOW_FAILURES=true env var probably doesn't do what you think it does

davidbarsky:
> noting here that the ALLOW_FAILURES=true env var probably doesn't do what you think it does

Damn, okay. I'll need to learn how Travis works _one_ of these days.

> I found in another pull that I wasn't able to fmt sources without incurring a bigger git diff of unrelated files. Formatting files is actually good for contributors because its often a feature of rust's efforts to improve the quality of life tools like rls and its vscode extension ( which auto formats on save ). Consistency is also good for readability.

Totally agree. I was using rustfmt 1.0, but Travis, for whatever reason, had 0.99, which had a _slightly_ different formatting config, which would cause failures.

Thanks for setting this up!

PR #58, opened 2018-12-20
impl multi_value_headers

softprops:
*Issue #, if available:*

*Description of changes:*

This is a followup to the previous lambda-http pull to add support for multi value headers. note this has no externally visible changes from a user perspective.  These were previously represented which [http::HeaderMap](https://docs.rs/http/0.1.14/http/struct.HeaderMap.html) which abstracts over these cases already. This changes just adds parsing an tests for cases where aws events provide them.

heads up on what to look for next.

I plan on following up quickly with a pull to add support multi value query string parameters then support for transparent alb events. These are roughly the same with some notable differences outlined [here](https://serverless-training.com/articles/api-gateway-vs-application-load-balancer-technical-details/),  all of which I believe as surmountable. I wanted to get these changes out in iterative chunks go get the merge queue going faster. Expect fast follow ups


lambda-http/tests/data/apigw_multi_value_proxy_request.json
```diff
@@ -0,0 +1,131 @@
...
+    "CloudFront-Is-SmartTV-Viewer":[
+      "false"
+    ],
+    "CloudFront-Is-Tablet-Viewer":[
+      "false"
+    ],
+    "CloudFront-Viewer-Country":[
+      "US"
+    ],
+    "":[
```
softprops: I'm not sure if this is a documentation bug but this is what aws [has documented](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format) I'm not sure what the right external channels are but you may want to leverage internal channels to notify the team that maintains these or clarify in the docs what an empty by present header name means :)

lambda-http/src/request.rs
```diff
@@ -119,11 +165,13 @@ where
         where
             A: MapAccess<'de>,
         {
-            let mut headers = http::HeaderMap::new();
+            let mut headers = map
```
softprops: took a liberty to make a small optimization here. serde's internal map model may provide a size hint for those building up deserialized maps to pre-allocate an appropriately sized map. I'm doing that here.

lambda-http/src/request.rs
```diff
@@ -224,8 +285,24 @@ mod tests {
     fn deserializes_request_events() {
         // from the docs
         // https://docs.aws.amazon.com/lambda/latest/dg/eventsources.html#eventsources-api-gateway-request
-        let input = include_str!("../tests/data/proxy_request.json");
-        assert!(serde_json::from_str::<GatewayRequest<'_>>(&input).is_ok())
+        let input = include_str!("../tests/data/apigw_proxy_request.json");
```
softprops: renamed this file to add an `apigw` prefix for clarity. In a follow up I'll be adding `alb` samples to differentiate in tests

softprops:
just realized multi_value_headers in the response isn't in this diff. putting that in the oven now. will be ready soon

softprops:
pushed a change to this branch to add multi value headers in response as well to complete the logical change

lambda-http/src/response.rs
```diff
@@ -22,23 +22,47 @@ pub(crate) struct GatewayResponse {
...
+        serialize_with = "serialize_multi_value_headers"
+    )]
+    pub multi_value_headers: HeaderMap<HeaderValue>,
     #[serde(skip_serializing_if = "Option::is_none")]
     pub body: Option<Body>,
     #[serde(skip_serializing_if = "Not::not")]
     pub is_base64_encoded: bool,
 }
 
+#[cfg(test)]
```
softprops: to a liberty as well. this is an internal type and its Default impl is only used in tests. By setting the `cfg(test)` attr hear this doesn't not add compile time/size tax to runtime produced artifacts

lambda-http/src/request.rs
davidbarsky: Nice!

lambda-http/tests/data/apigw_multi_value_proxy_request.json
davidbarsky: Good question; will do. Not entirely sure myself...

PR #59, opened 2018-12-20
impl http multi value query str parameters

softprops:
*Issue #, if available:*

*Description of changes:*

this is a follow on to https://github.com/awslabs/aws-lambda-rust-runtime/pull/58 ( which adds support for (de)serializing multi valued headers. since this is a pull against master it will also contain those changes since this was branched from. That pull should be reviewed first and this second.

There will be one final pull to tie the bow and make the lambda-http crate transparently handle both api gateway _and_ alb events. 


lambda-http/src/request.rs
```diff
@@ -171,8 +223,16 @@ impl<'a> From<GatewayRequest<'a>> for HttpRequest<Body> {
                 path
             )
         });
-
-        builder.extension(QueryStringParameters(query_string_parameters));
+        // multi valued query string parameters are always a super
+        // set of singly valued query string parameters,
+        // when present, multi-valued query string parameters are preferred
+        builder.extension(QueryStringParameters(
```
softprops: abstracting over both cases was a conscious decision. I've found multi valued query string parameter cases awkward to work as two separate collections. the thought process behind making `StrMap` represent both cases was inspired by `HeaderMap` which abstracts over single and multi valued headers.  The semantics here are documented in the public api for users

davidbarsky:
Thanks! Looks solid to me.

davidbarsky:
(There are merge conflicts. I think I merged out of order. I can fix them tomorrow, or you can. Up to you.)

softprops:
pushed conflict resolution commit

Issue #53, opened 2018-12-17
[RFC] Runtime crates refactor


sapessi:
At the moment the AWS Lambda rust runtime is composed of two crates: The [Runtime Client](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime-client) and the [Runtime](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime). The runtime client implements a client SDK to the [Lambda runtime APIs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html), taking care of all the communication to the runtime endpoint to receive new events, send responses and errors. The runtime implements the “main loop” of a function's lifecycle: call `/next` to receive the next event, invoke the handler, and then post responses or errors to the relevant endpoints. The runtime client SDK only deals with byte slices and does not understand types; the runtime crate, on the other hand, expects the event and response types to be serializable/deserializable with Serde.

We have received requests to handle multiple event types in the same Lambda function (#29) as well as abstracted implementations that abstract the event into a well-known type in rust, such as the HTTP crate (#18). To make it easier to extend the runtime and support future use-cases, we propose to split the Rust Lambda runtime into three separate crates:

1. The existing runtime Client SDK - follows the current behavior and only deals with byte slices
2. The new **runtime-core** - implements the main event loop of receiving events from the `/next` endpoint and posting responses. Expects a handler that only deals with byte slices.
3. The runtime - the current, typed, implementation of the runtime will depend on the **runtime-core** and wraps the given typed handler into its own byte-slice-based wrapper for the runtime-core.

The runtime-core crate will declare this handler type:

```rust
pub trait Handler<Error> {
where Error: Display + ErrorExt + Send + Sync {
    fn run(&mut self, event: Vec<u8>, ctx: Context) -> Result<Vec<u8>, Error>;
}
```

The typed runtime will declare the same handler type as now and wrap the run method to perform the serialization/deserialization before passing data to the core crate.

srijs:
Quick feedback on this as well, I generally like the idea, but relying on `Vec<u8>` to this extend may be a bit restrictive and might hinder attempts at reducing or re-using allocations.

(How much these types of optimizations are really desirable or necessary, I will leave that up to you. Personally when I need that kind of sustained high-throughput performance I tend to opt for something different than serverless, but I could also imagine folks wanting their lambdas to be as efficient as possible.)

[...]

PR #63, opened 2018-12-27
[Do not merge] Refactor to address RFCs

sapessi:
This is a major refactor of the crate structure to address the comments in RFCs #53. While the changes are major, they should not be breaking for existing code. 

*Description of changes:* Moved all of the "main loop" logic in a `runtime-core` crate that defines a `Vec<u8>`-based handler. The runtime crate now simply wraps the bytes handler into a typed handler method using serde.


lambda-runtime-core/src/runtime.rs
```diff
@@ -1,33 +1,13 @@
...
+use log::*;
 use tokio::runtime::Runtime as TokioRuntime;
 
-use crate::{
-    context::Context,
-    env::{ConfigProvider, EnvConfigProvider, FunctionSettings},
-    error::{HandlerError, RuntimeError},
-};
-
 const MAX_RETRIES: i8 = 3;
```
davidbarsky: If this is defined as a `pub(crate) const`, this _should_ be accessible to other functions within the crate, and as such, don't require threading the variable though a bunch of functions.

lambda-runtime/Cargo.toml
```diff
@@ -19,12 +19,13 @@ maintenance = { status = "actively-developed" }
...
-backtrace = "^0.3"
-lambda_runtime_client = { path = "../lambda-runtime-client", version = "^0.1" }
-chrono = "^0.4"
+lambda_runtime_core = { path = "../lambda-runtime-core", version = "^0.1" }
 
 [dev-dependencies]
 simple_logger = "^1"
+simple-error = "^0.1"
+failure = "^0.1"
+#rusoto_core = "^0.35"
```
davidbarsky: If we're not using these, should the `rusoto_*` crate be removed?

lambda-runtime/examples/custom_error.rs
```diff
@@ -0,0 +1,66 @@
+extern crate lambda_runtime as lambda;
+extern crate log;
+extern crate serde_derive;
+extern crate simple_logger;
```
davidbarsky: As far as I can tell, you've removed the `extern crate` in other files. Perhaps they should be removed here as well?

lambda-runtime/examples/custom_error_failure.rs
```diff
@@ -0,0 +1,47 @@
+extern crate failure;
```
davidbarsky: Ditto for `extern crate` here.

lambda-runtime/src/lib.rs
```diff
@@ -32,55 +32,92 @@
...
+    fn run(&mut self, event: E, ctx: Context) -> Result<O, error::HandlerError> {
+        (*self)(event, ctx)
+    }
+}
+
+/// Wraps a typed handler into a closure that complies with the `Handler` trait
+/// defined in the `lambda_runtime_core` crate. The closure simply uses `serde_json`
+/// to serialize and deserialize the incoming event from a `Vec<u8>` and the output
+/// to a `Vec<u8>`.
+fn wrap<E, O>(mut h: impl Handler<E, O>) -> impl FnMut(Vec<u8>, Context) -> Result<Vec<u8>, error::HandlerError>
```
davidbarsky: I like this!

lambda-runtime-client/src/error.rs
```diff
@@ -135,6 +153,8 @@ impl Error for ApiError {
         None
     }
 }
+unsafe impl Send for ApiError {}
```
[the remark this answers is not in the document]
sapessi: This will go away, with the next iteration all errors will switch to failure. I have quite a bit of legacy error handling code in here. Just wanted to make one big change at a time.

lambda-runtime-core/src/error.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// When a new `RuntimeError` is created the stack trace for the error is collected
+    /// automatically using the `backtrace` crate.
+    ///
+    /// # Arguments
+    ///
+    /// * `msg` The error message to be attached to the error.
+    ///
+    /// # Returns
+    /// A new `RuntimeError` instance with the `recoverable` property set to `false`.
+    pub(crate) fn unrecoverable(msg: &str) -> RuntimeError {
```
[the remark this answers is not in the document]
sapessi: Same comment as above.

lambda-runtime-core/src/error.rs
```diff
@@ -0,0 +1,87 @@
+//! The error module defines the error types that can be returned
+//! by custom handlers as well as the runtime itself.
+use std::{env, error::Error, fmt};
+
+use lambda_runtime_client::error::ApiError;
+
+/// The `RuntimeError` object is returned by the custom runtime as it polls
+/// for new events and tries to execute the handler function. The error
+/// is primarily used by other methods within this crate and should not be relevant
+/// to developers building Lambda functions. Handlers are expected to return
```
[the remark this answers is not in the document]
sapessi: Agree, no need to expose this. I will make the change as I work on the errors refactor.

lambda-runtime-core/src/handler.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn from(e: E) -> HandlerError {
+        HandlerError { msg: format!("{}", e) }
+    }
+}
+
+impl Display for HandlerError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self.msg)
+    }
+}
```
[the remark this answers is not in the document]
sapessi: Nope. It's because I was running on the stable toolchain and quite a few of fmt features don't work automatically there, switched to nightly. I'll push the fmt changes soon

lambda-runtime-core/src/runtime.rs
sapessi: I'm wondering whether I should move this to an env config variable so that it can be easily overridden. The variable was only meant to be internal configuration anyway.

lambda-runtime/Cargo.toml
sapessi: Yup, forgot to do that before pushing

lambda-runtime/examples/custom_error.rs
sapessi: I missed quite a few examples - that's also what broke the build. pushing the fix soon.

lambda-runtime-core/src/runtime.rs
```diff
@@ -252,7 +194,7 @@ where
                                     "Error for {} is not recoverable, sending fail_init signal and panicking",
                                     request_id
                                 );
-                                self.runtime_client.fail_init(&e);
+                                self.runtime_client.fail_init(&ErrorResponse::from(Box::new(e)));
```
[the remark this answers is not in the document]
sapessi: Same as above - haven't touched errors at all

lambda-runtime-core/src/runtime.rs
davidbarsky: I'm not sure I'm a fan of enviroment variables for configuration (see [here](https://twitter.com/copyconstruct/status/983216554539073536)), and I'm not sure this should be customizable. I don't plan to allow a knob when I push the async changes (e.g., https://gist.github.com/seanmonstar/2469a8e544ada2d31d13c8ee54f17d48)

lambda-runtime/src/lib.rs
```diff
@@ -93,10 +97,11 @@ where
 ///
 /// # Panics
 /// The function panics if the Lambda environment variables are not set.
-pub fn start<E, O>(f: impl Handler<E, O>, runtime: Option<TokioRuntime>)
+pub fn start<EV, O, ER>(f: impl Handler<EV, O, ER>, runtime: Option<TokioRuntime>)
 where
-    E: serde::de::DeserializeOwned,
+    EV: serde::de::DeserializeOwned,
     O: serde::Serialize,
+    ER: Fail + LambdaErrorExt + Display + Send + Sync,
```
davidbarsky: I'm of the weak opinion that we shouldn't use two-letter generic names, and go all-in with spelling out the generic parameter. e.g., `Event` instead of `EV` and `Error` instead of `ER`.

sapessi: I had changed it to those initially - ran into all sort of issues with the generic name getting confused with the failure::Error when imported. I'll try again to make the changes.

lambda-runtime/src/lib.rs
```diff
@@ -30,57 +26,194 @@
...
+    Event: serde::de::DeserializeOwned,
+    Output: serde::Serialize,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
+{
+    let wrapped = wrap(f);
+    start_with_config(wrapped, &EnvConfigProvider::default(), runtime)
+}
+
+/// Initializes the Lambda runtime with the given handler. Optionally this macro can
+/// also receive an instance of Tokio runtime for the Hyper HTTP client. If the Tokio
```
softprops: I feel like it would be more useful for readers of these api docs to see something like "receive a customized instance of Tokio runtime to drive internal lambda operations to completion". The mention of Hyper HTTP client is kind of an implementation detail that's not useful for readers.

lambda-runtime-client/Cargo.toml
```diff
@@ -24,4 +24,4 @@ serde_json = "^1"
 serde_derive = "^1" 
 log = "0.4"
 lambda_runtime_errors = { path = "../lambda-runtime-errors", version = "^0.1" }
-failure = "^0.1"
\ No newline at end of file
+failure = { version = "^0.1", features = ["std", "backtrace", "derive"] }
```
davidbarsky: These features should be on by default.

lambda-runtime-client/src/client.rs
```diff
@@ -125,8 +125,9 @@ pub struct EventContext {
 pub struct RuntimeClient {
     _runtime: Runtime,
     http_client: Client<HttpConnector, Body>,
-    endpoint: Uri,
+    next_endpoint: Uri,
     runtime_agent: String,
+    host: String,
```
davidbarsky: Why the addition of `host` and change from `endpoint` to `next_endpoint`?

lambda-runtime-client/src/error.rs
```diff
@@ -146,3 +146,37 @@ impl From<Context<ApiErrorKind>> for ApiError {
...
+
+    #[test]
+    fn does_not_produce_stack_trace() {
+        env::remove_var("RUST_BACKTRACE");
+        let err = format_err!("Test error").compat();
+        let resp_err = ErrorResponse::from(err);
+        assert_eq!(resp_err.stack_trace, None);
+    }
+
+    /*#[test] skip this tests as it's broken right now
```
davidbarsky: if this isn't running, I'd prefer to remove it rather than commenting it out.

lambda-runtime/src/lib.rs
```diff
@@ -30,57 +26,194 @@
...
+    /// Method to execute the handler function
+    fn run(&mut self, event: Event, ctx: Context) -> Result<Output, EventError>;
+}
+
+/// Implementation of the `Handler` trait for both function pointers
+/// and closures.
+impl<Function, Event, Output, EventError> Handler<Event, Output, EventError> for Function
+where
+    Function: FnMut(Event, Context) -> Result<Output, EventError>,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
```
softprops: Few notes here. 

Because [Fail requires  Display + Send + Sync](https://docs.rs/failure/0.1.5/failure/trait.Fail.html). Adding `Display + Send + Sync` on top of `Fail` is redundant.

I'll have to read through a bit more to understand the purpose of `LambdaErrorExt` but my initial impression is this. The core interface should make the trivial easy and the complex possible. If a dependency on `Fail` wasn't enough implementing `LambdaErrorExt` may seem like a tall order for making the simple trivial. Thoughts?

lambda-runtime-client/src/error.rs
```diff
@@ -1,20 +1,25 @@
...
+use std::{
+    fmt::{self, Display},
+    option::Option,
+};
 
 /// Error type description for the `ErrorResponse` event. This type should be returned
 /// for errors that were handled by the function code or framework.
-pub const ERROR_TYPE_HANDLED: &str = "Handled";
+#[allow(dead_code)]
+pub(crate) const ERROR_TYPE_HANDLED: &str = "Handled";
```
softprops: What advantage is gained by managing these through a string interface over an enum? Is there a finite number of these?

lambda-runtime-client/src/error.rs
```diff
@@ -37,148 +42,141 @@ pub struct ErrorResponse {
...
             error_message: message,
-            error_type: String::from(ERROR_TYPE_HANDLED),
+            error_type: err_type,
             stack_trace: Option::default(),
+        };
+        // assume that failure is smart enough to only collect a backtrace
+        // if the env variable is enabled
+        if let Some(stack) = backtrace {
+            trace!("Begin backtrace collection");
+            let trace_string = format!("{:?}", stack)
```
softprops: Can you walk me though how we arrived as this imp? It seems like it's taking the `Debug` representation of stack, splitting that by lines of the debug string, converting those to lines and collecting them? I see there are only a few options available for introspecting [Backtrace](https://docs.rs/failure/0.1.1/failure/struct.Backtrace.html) but depending on something like the stability of something like a toString() impl makes me thing we may want to consider another option.

lambda-runtime-client/src/error.rs
```diff
@@ -37,148 +42,141 @@ pub struct ErrorResponse {
...
+                .lines()
+                .map(|s| s.to_string())
+                .collect::<Vec<String>>();
+            trace!("Completed backtrace collection");
+            err.stack_trace = Option::from(trace_string);
```
softprops: Would `Some(trace_string)` to here?

lambda-runtime-client/src/client.rs
sapessi: `next_endpoint` is the pre-parsed uri for the `/next` endpoint, which remains stable throughout the lifetime of the client. The host is used to re-generate the `Uri` for the response endpoints for each request. I'm caching the `next_endpoint` `Uri` value to avoid parsing it each time.

lambda-runtime-client/src/error.rs
softprops: Secondary comment. I feel like we may just want to call this something like `trace_lines` to be a bit more representative. `trace_string` reads like [hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation) but is lying about its type :)

lambda-runtime-client/src/error.rs
```diff
@@ -37,148 +42,141 @@ pub struct ErrorResponse {
...
-    }
 }
-
-impl fmt::Display for ApiError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        write!(f, "{}", self.msg)
-    }
+/// Failure context for the `ApiError` type. The kind is used to indicate whether the
+/// error is recoverable and should be retried or not.
+#[derive(Clone, Eq, PartialEq, Debug, Fail)]
```
softprops: I'm curious where the `Eq` requirement come into play. Was this actually needed. Typically I've only seem `PartialEq` required for value comparison .

lambda-runtime-client/src/error.rs
```diff
@@ -37,148 +42,141 @@ pub struct ErrorResponse {
...
-        ApiError::new(e.description())
+#[cfg(test)]
+pub(crate) mod tests {
+    use super::*;
+    use failure::format_err;
+    use std::env;
+
+    #[test]
+    fn does_not_produce_stack_trace() {
+        env::remove_var("RUST_BACKTRACE");
```
softprops: I can't say this without proof but I feel like a dependency on modifying global env state inside a unit test is going to reduce the potential for deterministic tests. I think cargo may run tests in parallel by default.

lambda-runtime-core/build.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    // next generate the metadata function for the runtime
+    let dest_path = Path::new(&out_dir).join("metadata.rs");
+    let mut f = fs::File::create(&dest_path).unwrap();
+
+    f.write_all(
+        format!(
+            "
+/// returns metdata information about the Lambda runtime
+pub fn runtime_release() -> &'static str {{
```
softprops: I think this is cool but it seems like a lot of code. an MVP for this may have just been to use [one of the cargo provided env vars](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-crates)  in the source of the crate itself. 

```rust
lazy_static! {
    static ref USER_AGENT: String = format!(
        "{}/{}",
         env!("CARGO_PKG_NAME"),
         env!("CARGO_PKG_VERSION")
    );
}
```

Do we anticipate the rust version and channel a crate was built with having enough differentiating behavior that would require this extra information ( beyond the crate version ) to warrant the added build complexity?

lambda-runtime-core/examples/simple.rs
```diff
@@ -0,0 +1,21 @@
+use failure::format_err;
+use lambda_runtime_core::{lambda, Context, HandlerError};
+use simple_logger;
+use std::error::Error;
+
+fn main() -> Result<(), Box<dyn Error>> {
```
softprops: something I've been meaning to clean up in these examples. Unless we're bubbling up errors with `?` we're not getting any value out of giving main a result type. Examples should be "teaching" what you want so rather than removing the `main()` return type you may want to remove the `.unwrap()` below

lambda-runtime-core/src/env.rs
```diff
@@ -39,8 +40,8 @@ pub trait ConfigProvider {
 /// used by the `start()` method of this module.
 pub struct EnvConfigProvider;
 
-impl EnvConfigProvider {
-    pub fn new() -> Self {
+impl std::default::Default for EnvConfigProvider {
+    fn default() -> Self {
         EnvConfigProvider {}
```
softprops: I'm curious if the `{}` is needed here.

lambda-runtime-core/src/handler.rs
```diff
@@ -0,0 +1,20 @@
...
+/// Functions acting as a handler must conform to this type.
+pub trait Handler<EventError> {
+    /// Run the handler.
+    fn run(&mut self, event: Vec<u8>, ctx: Context) -> Result<Vec<u8>, EventError>;
+}
+
+impl<'ev, Function, EventError> Handler<EventError> for Function
+where
+    Function: FnMut(Vec<u8>, Context) -> Result<Vec<u8>, EventError>,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
```
softprops: I mentioned this above but `Fail` implies `Display + Send + Sync` so we can probably save some visual noise here by removing them.

lambda-runtime-core/src/runtime.rs
```diff
@@ -36,14 +24,16 @@ where
...
 /// The function panics if the Lambda environment variables are not set.
-pub fn start<E, O>(f: impl Handler<E, O>, runtime: Option<TokioRuntime>)
+pub fn start<EventError>(f: impl Handler<EventError>, runtime: Option<TokioRuntime>)
 where
-    E: serde::de::DeserializeOwned,
-    O: serde::Serialize,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
 {
-    start_with_config(f, &EnvConfigProvider::new(), runtime)
+    start_with_config(f, &EnvConfigProvider::default(), runtime)
```
softprops: something just feels off to me about requiring a constructor-less struct to need a default impl and a ref here.

lambda-runtime-core/src/runtime.rs
```diff
@@ -76,11 +66,13 @@ macro_rules! lambda {
...
-    E: serde::de::DeserializeOwned,
-    O: serde::Serialize,
-    C: ConfigProvider,
+pub fn start_with_config<Config, EventError>(
+    f: impl Handler<EventError>,
+    config: &Config,
+    runtime: Option<TokioRuntime>,
+) where
+    Config: ConfigProvider,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
```
softprops: same as above. `Fail` implies `Display + Send + Sync`

lambda-runtime-core/src/runtime.rs
```diff
@@ -123,38 +117,35 @@ where
...
-pub(crate) fn start_with_runtime_client<E, O>(
-    f: impl Handler<E, O>,
+pub(crate) fn start_with_runtime_client<EventError>(
+    f: impl Handler<EventError>,
     func_settings: FunctionSettings,
     client: RuntimeClient,
 ) where
-    E: serde::de::DeserializeOwned,
-    O: serde::Serialize,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
```
softprops: And again... `Fail` implies `Display + Send + Sync` :)

lambda-runtime-core/src/runtime.rs
```diff
@@ -123,38 +117,35 @@ where
...
-    _phan: PhantomData<(E, O)>,
+    _phantom: PhantomData<EventError>,
 }
 
 // generic methods implementation
-impl<F, E, O> Runtime<F, E, O> {
+impl<Function, EventError> Runtime<Function, EventError>
+where
+    Function: Handler<EventError>,
+    EventError: Fail + LambdaErrorExt + Display + Send + Sync,
```
softprops: This trait stack seems referenced often enough that it seems like it warrants a type alias.

lambda-runtime-core/src/runtime.rs
```diff
@@ -299,26 +272,15 @@ where
...
-                match parse_result {
-                    Ok(ev) => {
-                        let mut handler_ctx = Context::new(self.settings.clone());
-                        handler_ctx.invoked_function_arn = invocation_ctx.invoked_function_arn;
-                        handler_ctx.aws_request_id = invocation_ctx.aws_request_id;
-                        handler_ctx.xray_trace_id = invocation_ctx.xray_trace_id;
-                        handler_ctx.client_context = invocation_ctx.client_context;
-                        handler_ctx.identity = invocation_ctx.identity;
-                        handler_ctx.deadline = invocation_ctx.deadline;
+                let mut handler_ctx = Context::new(self.settings.clone());
```
softprops: this wasn't really a change introduced in this pull but I feel like there's something easy to get wrong in this reassignment of values from the settings + the context it makes me wonder if there should just be an associated method on `Context` that takes both objects and manages producing a new `Context`. That would probably have the side benefit of improved potential for unit testability

lambda-runtime-errors-derive/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+edition = "2018"
+description = "Rust runtime errors derive for AWS Lambda"
+keywords = ["AWS", "Lambda", "Runtime", "Rust"]
+license = "Apache-2.0"
+homepage = "https://github.com/awslabs/aws-lambda-rust-runtime"
+repository = "https://github.com/awslabs/aws-lambda-rust-runtime"
+documentation = "https://docs.rs/lambda_runtime"
+
+[dependencies]
+log = "^0.4"
```
softprops: this crate doesn't actually depend on `log`. It's unclear if that's not also the case for `syn` as well.

lambda-runtime-errors/errorgen.py
```diff
@@ -0,0 +1,138 @@
+#!/usr/bin/env python3
+
+# Generates the LambdaErrorExt implementation for all of the Errors
+# in the standard library, excluding unstable APIs and errors that
+# require generics.
+#
+# !! Please note this script is a hacky, short term solution !!
```
softprops: let's hope so!

lambda-runtime-errors/src/error_ext_impl.rs
```diff
@@ -0,0 +1,248 @@
+// Generated code, DO NOT MODIFY!
```
softprops: can we add a comment in the generated code that describes what this is used for and why it is necessary?

lambda-runtime-errors/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// of the error type. For example `MyCustomError`.
+    ///
+    /// # Return
+    /// An alphanumeric identifier for the error
+    fn error_type(&self) -> &str;
+}
+
+impl LambdaErrorExt for Error {
+    fn error_type(&self) -> &str {
+        if let Some(name) = self.find_root_cause().name() {
```
softprops: you can probably avoid the really return and just return an expression 

```rust
self.find_root_cause().name().unwrap_or_else(|| "FailureError")
```

lambda-runtime/examples/custom_error.rs
```diff
@@ -0,0 +1,70 @@
...
+}
+impl Error for CustomError {}
+impl From<std::num::ParseIntError> for CustomError {
+    fn from(i: std::num::ParseIntError) -> Self {
+        CustomError::new(&format!("{}", i))
+    }
+}
+// the value return by the error_type function is included as the
+// `errorType` in the AWS Lambda response
+impl LambdaErrorExt for CustomError {
```
softprops: I can foresee this getting very onerous. Not only for those familiar with rust but for those new to rust dabbling through experimenting transitioning a lambda to rust. Is there any thing the runtime could provide that could provide a default impl that allows for specialization when needed? This goes back to making the trivial easy and the complex possible.

lambda-runtime-core/src/runtime.rs
```diff
@@ -267,27 +240,27 @@ where
 
     /// Invoke the handler function. This method is split out of the main loop to
     /// make it testable.
-    pub(super) fn invoke(&mut self, e: E, ctx: Context) -> Result<O, HandlerError> {
-        (&mut self.handler).run(e, ctx)
+    pub(super) fn invoke(&mut self, e: Vec<u8>, ctx: Context) -> Result<Vec<u8>, EventError> {
```
davidbarsky: I'm forgetful, but we're set on the `Vec<u8>`, right? Not going the route of [`bytes::Buf`](https://carllerche.github.io/bytes/bytes/trait.Buf.html) or anything else? I believe we discarded the `&[u8]` for the lifetime annotations.

lambda-runtime-core/src/runtime.rs
davidbarsky: Can you elaborate?

lambda-runtime-errors-derive/src/lib.rs
```diff
@@ -0,0 +1,40 @@
+use proc_macro2::TokenStream;
+use quote::quote;
+use synstructure::decl_derive;
+
+decl_derive!([LambdaErrorExt, attributes()] => lambda_error_derive);
+
+fn lambda_error_derive(s: synstructure::Structure) -> TokenStream {
+    /*
+    if let Data::Struct(d) = &s.ast().data {
```
davidbarsky: Should this be commented out?

lambda-runtime-errors/errorgen.py
davidbarsky: specialization _when_

lambda-runtime-errors/src/error_ext_impl.rs
davidbarsky: +1

lambda-runtime-errors/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+    /// fn my_handler(_event: Vec<u8>, _ctx: Context) -> Result<Vec<u8>, Compat<Error>> {
+    ///     let age = "hello"; // this will throw an error when we try to parse it into an int
+    ///     age.parse::<u8>().failure_compat()?;
+    ///     Ok(vec!())
+    /// }
+    /// ```
+    fn failure_compat(self) -> Result<OK, Compat<Error>>;
+}
+
+impl<OK, ERR> LambdaResultExt<OK, ERR> for Result<OK, ERR>
```
davidbarsky: I wonder if the all-caps generic params could be replaced with something like `T, E`.

lambda-runtime-errors/src/lib.rs
```diff
@@ -0,0 +1,190 @@
...
+//! AWS Lambda Rust runtime.
+mod error_ext_impl;
+
+pub use crate::error_ext_impl::*;
+
+use failure::{format_err, Compat, Error, Fail};
+use std::fmt;
+
+/// The `LambdaErrorExt` trait defines the `error_type()` method used
+/// by the AWS Lambda runtime client to generate `ErrorResponse`
```
davidbarsky: How should we refer to this library? “AWS Lambda runtime client” seems a bit... long. Been a thing for a while, in fairness.

lambda-runtime-errors/src/lib.rs
```diff
@@ -0,0 +1,190 @@
+//! The Lambda runtime errors crate defines the `LambdaErrorExt` trait
+//! that can be used by libriaries to return errors compatible with the
+//! AWS Lambda Rust runtime.
+mod error_ext_impl;
+
+pub use crate::error_ext_impl::*;
+
+use failure::{format_err, Compat, Error, Fail};
```
davidbarsky: I think Failure can be placed behind a feature flag that's on by default. Do you disagree?

lambda-runtime/examples/custom_error.rs
davidbarsky: Something I'm debating—not really brought up with Stefano—is to point people to just use Failure and be done with it, but I'm not convinced that's the right approach.

lambda-runtime/examples/custom_error_failure.rs
```diff
@@ -0,0 +1,51 @@
...
+#[fail(display = "Custom Error")]
+struct CustomError;
+impl LambdaErrorExt for CustomError {
+    fn error_type(&self) -> &str {
+        "CustomError"
+    }
+}
+impl From<std::num::ParseIntError> for CustomError {
+    fn from(_i: std::num::ParseIntError) -> Self {
+        CustomError {}
```
davidbarsky: @softprops alluded to this a while back, but it might be nice to have custom error wrap the error like this:

```rust
#[derive(Fail, Debug)]
#[fail(display = "Error: {:?}", _0)]
struct LambdaError(std::num::ParseIntError);
```

...and have the `CustomError` wrap `_i`.

(I'll go through the examples and try go away from the “Custom*” naming scheming—it feels like it takes the perspective of the library author, not the end customer _reading_ the code. This parenthetical is not blocking, however.)

lambda-runtime-core/build.rs
softprops: It wouldn't be a complete replacement for everything here but it may get you pretty far. Thing I'm poking at is if we actually _need_ everything here :)

lambda-runtime-core/src/runtime.rs
softprops: Had my eyes on these error types. The trait stack seems duplicated in signatures enough that in thought and alias may make change more manageable.

lambda-runtime/examples/custom_error.rs
softprops: Failure sounds like a same middle ground as it plays nice with the ecosystem what already exists. The complexity seems to come from an implementation detail but if feels like too much weight to on users which won't benefit as much/appreciate the feature of the named errors inside lambda responses. It's likely user logging errors to cloudwatch logs will provide the same debugging benefit at no extra interface cost. 

That said do we gain anything from failure over Box<Error> beside the back trace? Std Errors are really the least common denominator and work well with ? Chaining

lambda-runtime-errors/src/lib.rs
softprops: Crowbar did that with error_chain. It's actually a nice feature of rust that fits the 0 cost tagline. Only pay for what you use, and make it possible to opt out of what you dont

lambda-runtime-errors/src/lib.rs
softprops: No need to deviate from stdlib conventions https://doc.rust-lang.org/std/result/enum.Result.html

sapessi:
> Posted a handful of comments. Notably I'm wondering if this change could been broken down into digestable chunks (pulls). This felt like a lot to absorb in one pull.

Apologies for the huge number of changes. We worked on bite-sized chunks with @davidbarsky as he reviewed each commit during development. The nature of the changes (refactor) made them hard to publish them in separate PRs without potentially breaking stuff for anyone depending on master. Hopefully it won't (need to) happen again.

lambda-runtime-client/src/error.rs
```diff
@@ -37,148 +42,141 @@ pub struct ErrorResponse {
...
-impl RuntimeApiError for ApiError {
-    fn to_response(&self) -> ErrorResponse {
-        let backtrace = format!("{:?}", self.backtrace);
-        let trace_vec = backtrace.lines().map(|s| s.to_string()).collect::<Vec<String>>();
-        let mut err = ErrorResponse::unhandled(self.msg.clone());
-        err.stack_trace = Option::from(trace_vec);
-
-        err
-    }
+    /*#[test] skip this tests as it's broken right now
```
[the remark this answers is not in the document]
sapessi: I removed this as per @davidbarsky's suggestion. We will re-introduce this once we finally figure out what is going on with backtraces in failure... #!@?!!&

lambda-runtime-client/Cargo.toml
```diff
@@ -21,6 +21,7 @@ tokio = "0.1"
 http = "0.1"
 serde = "^1"
 serde_json = "^1"
-serde_derive = "^1"
+serde_derive = "^1"
```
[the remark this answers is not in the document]
sapessi: Done - will push once I've gone through all the changes.

lambda-runtime/src/lib.rs
sapessi: The reason why we have Display, Send, and Sync there is that I don't trust the failure crate to maintain those in the `Fail` trait. This library has burned me more than once recently (scorched me to death is more appropriate). Since in the next iteration of the library we plan to make the core async we wanted to introduce a constraint of our own in this, even if failure removes it.

Lambda gives us the ability to return an error type field which we plan to use to specify the error's root cause. The LambdaErrorExt trait defines the `error_type(&self) -> &str` to provide this. In the crate we have included implementations of the trait for almost all errors in the standard library as well as failure types. With these default implementations, I'm hoping to still make the trivial easy (+1 on this turn of phrase by the way, very nicely put).

lambda-runtime-client/src/error.rs
sapessi: Went back and checked - these are no longer required since we only ever send "Unhandled" in the error type header. I've removed both constants and made that value a static string in the one method that generates error requests. I honestly expect that header to disappear too in the next iteration of the runtime APIs.

lambda-runtime/src/lib.rs
sapessi: Like it. Made the change

lambda-runtime/src/lib.rs
davidbarsky: > The reason why we have Display, Send, and Sync there is that I don't trust the failure crate to maintain those in the Fail trait. This library has burned me more than once recently (scorched me to death is more appropriate). Since in the next iteration of the library we plan to make the core async we wanted to introduce a constraint of our own in this, even if failure removes it.

You're not _wrong_ to distrust Failure given your experience, but the way [the Failure is evolving](https://github.com/rust-lang-nursery/failure/issues/287) is to _just_ provide a derive for std::error::Error + Display + Send + Sync bounds. Additionally, the [standard library encourages](https://doc.rust-lang.org/std/error/trait.Error.html) implementing Debug + Display on errors, and I can think of a _single_ error in the standard library ([`std::sync::PoisonError`](https://doc.rust-lang.org/beta/std/sync/struct.PoisonError.html)) that doesn't implement Send/Sync. Most errors I've seen implement Send/Sync, and if they don't, that's often grounds for opening a PR/issue.

(With regard to `std::sync::PoisonError`—I believe that most people just panic on a poisoned mutex. I'll be _shocked_ if a Lambda customer has to worry about poisoned mutexes.

lambda-runtime-client/src/error.rs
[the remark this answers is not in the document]
davidbarsky: yeah, it might've been an older commit.

lambda-runtime-client/src/error.rs
sapessi: This was simply the easiest way I could think of - with my limited knowledge of rust - to get a readable stack trace in the format the Lambda runtime APIs expect: a list of strings.

lambda-runtime/src/lib.rs
davidbarsky: I think this might've been addressed partially in https://github.com/awslabs/aws-lambda-rust-runtime/pull/63#discussion_r249605686.

lambda-runtime-client/src/error.rs
sapessi: Yup :-) didn't even know I could do that.

lambda-runtime/src/lib.rs
softprops: Im thinking then may it's a good idea to put the Failure dependency behind a feature flag to maintain a more stable API. Scorch free!

lambda-runtime-client/src/error.rs
sapessi: It's all gone away in favor of:
```rust
err.stack_trace = Some(
                format!("{:?}", stack)
                    .lines()
                    .map(|s| s.to_string())
                    .collect::<Vec<String>>(),
            );
```

lambda-runtime-client/src/error.rs
sapessi: I don't believe it's needed - I think it's something I stuck in there while I was experimenting with another change and forgot to roll back. Good catch, will remove.

lambda-runtime/examples/custom_error.rs
softprops: I'm a customer as well!

lambda-runtime-core/src/env.rs
sapessi: Because I didn't know it wasn't...

lambda-runtime-core/src/runtime.rs
sapessi: I believe this recommendation came from clippy - instead of having a 0-params constructor implement the `Default` trait.

lambda-runtime-core/src/runtime.rs
sapessi: Agree on error types. Any name suggestions? `HandlerResultError`?

lambda-runtime-errors-derive/Cargo.toml
sapessi: Done

lambda-runtime-errors/src/error_ext_impl.rs
sapessi: Will do

lambda-runtime/examples/custom_error.rs
sapessi: That's why I had included the error derive crate, to allow developers to just:
```rust
#[derive(LambdaErrorExt)]
struct BasicCustomError;
```

Do you feel we need more? What should our generic be for the trait implementation be, any `Fail`?

lambda-runtime-errors-derive/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+        if let Fields::Named(fs) = &d.fields {
+            for f in fs.named.iter() {
+                let field_name = f.ident.clone();
+                println!("!!Field name: {}", field_name.expect("Field must be named"));
+            }
+        }
+    }*/
+
+    for v in s.variants() {
+        println!("Variant: {}", v.ast().ident);
```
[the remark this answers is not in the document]
sapessi: whoopsie - some careful debugging there on my part.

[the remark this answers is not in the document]
sapessi: The derive crate is very much a work in progress. Few more things I'd like to do.

lambda-runtime-errors/src/lib.rs
sapessi: I literally meant the client SDK - could be shortened to "runtime client"

lambda-runtime-core/src/runtime.rs
davidbarsky: We could rip off what Go's using internally.

lambda-runtime/examples/custom_error.rs
davidbarsky: @softprops 

> I'm a customer as well!

That's true, sorry! I just remember him telling me that people are creating Dashboards on specific error strings.

lambda-runtime-errors/src/lib.rs
davidbarsky: yeah, i know. it's not strictly related to this PR. Runtime client could work.

lambda-runtime/examples/custom_error.rs
sapessi: One more thing. We implemented errors this way to avoid forcing the failure crate on users of the runtime - it's purely a utility for us. When they take the runtime as a dependency their errors automatically become fails because we bring the failure crate. Once the standard library errors include back traces we can remove failure from the runtime implementation without it being a breaking change

lambda-runtime-errors-derive/src/lib.rs
sapessi: The derive crate is a work in progress. Need to do some cleaning up on this code and implement a few more things for enums

lambda-runtime/src/lib.rs
sapessi: Responded in another comment too (perhaps we should switch to the main PR thread for this conversation). With the current implementation, we don't force the failure crate or the `Fail` trait on developers building Lambda functions, their errors become Fail for us automatically because we include failure. This gives us the ability to switch away from failure in the future without making a breaking change.

lambda-runtime-client/src/error.rs
sapessi: Added comment

lambda-runtime-core/src/runtime.rs
sapessi: Resolving this. let's continue this conversation in the main thread

lambda-runtime-core/src/runtime.rs
sapessi: uh? What do you mean?

lambda-runtime/examples/custom_error_failure.rs
sapessi: Go ahead, examples shouldn't block this.

lambda-runtime-errors/src/lib.rs
sapessi: See my answer in the other comments - why put it behind a feature flag if it has no impact on developers using the library? They do not have to use failure, it's just a utility for us.

lambda-runtime-core/src/runtime.rs
sapessi: Yes, we definitely dropped slices (`&[u8]`). I'll look into using Buf today.

lambda-runtime-errors/src/lib.rs
davidbarsky: The feature flag allows you conditionally disable compilation for libraries that you're not using.

I think we should use Failure internally, but customers shouldn't need to pay the compilation cost if they're not using Failure in their application. The conditional compilation/feature flag might make some the conversions easier.

sapessi:
Moving the `failure` conversation to a thread here to make it more manageable.

I spoke with @davidbarsky this morning. The dependency on `failure` is something we decided to take because it makes our life easier when accepting any kind of error and also allows us to collect backtraces. Because of how the `Fail` trait works, we do not need developers to include failure as a dependency in their functions. Their errors - as long as they implemented `Error` from the standard library - automatically implement `Fail` because we include `failure`. 

Both David and I agree that this is a temporary dependency that we will remove in the future, once backtraces are available in the standard library. Since we do not require `failure` to be included in functions the change should be non-breaking. 

Other than compilation time, and on the basis that we plan to remove failure in the future once backtraces are stable in the standard library, do you have any other concern with releasing `0.2` with `failure` @softprops? Unless we have a serious issue with the dependency, I'd rather not block the release, it's been a long time coming.

softprops:
No concerns here. I didn't realize all you need is std Error. That sounds great. I'd love a release. I've got a growing number of serverless framework templates that depend on git urls for the http lambda module. I'd love to update those to a release

sapessi:
Huzza! I'll merge and push 0.2 tomorrow. The http crate will go out as a 0.1. I'll probably hold off on the errors_derive crate until I've had time to clean it up properly.

PR #65, opened 2018-12-29
[dont merge] support both apigw and alb events as http::{Request,Response} (cargo features)

softprops:
*Issue #, if available:*

*Description of changes:*

bizzaro branch off  of https://github.com/awslabs/aws-lambda-rust-runtime/pull/64 with cargo feature flag toggles for conditional compilation

Have not checked the documentation checkbox yet because Im not sure this will look like the end state


lambda-http/src/request.rs
```diff
@@ -21,50 +23,135 @@ use crate::{
...
+        api_id: String,
+        identity: Identity,
+    },
+    /// ALB request context
+    #[serde(rename_all = "camelCase")]
+    Alb { elb: Elb },
+}
+
+/// ALB request context
+#[cfg(all(feature = "alb", not(all(feature = "apigw", feature = "alb"))))]
```
softprops: TIL learned cfg attrs are like a smaller dialect of lisp!

lambda-http/Cargo.toml
```diff
@@ -30,3 +30,8 @@ serde_urlencoded = "0.5"
 [dev-dependencies]
 log = "^0.4"
 simple_logger = "^1"
+
+[features]
+default = ['apigw', 'alb']
```
softprops: big drawback here. if a consumer sets default-flags to false you'll get neither apigw nor alb! I don't think there's a way to express "there can be at least one!" in cargo

softprops:
gonna fix up these merge conflicts...

softprops:
with the 0.2 release looming what do you think about fold on this for now until we find its actually useful to enable feature flagging these. it seems the cost of maintenance added may not be worth the benefit. I'm kind of curious about hearing if this is an actual pain point _before_ investing in an ointment to relieve it. Thoughts?

davidbarsky:
> I'm kind of curious about hearing if this is an actual pain point before investing in an ointment to relieve it. Thoughts?

I think that's a good point. Feature flagging it _seems_ cleaner, but it might not be worth the complexity.

softprops:
Cool. I'll let you folks make the call. I'd be fine to close this with commitment to revisit this if there's demand for it

softprops:
I havent seen any demand for this so I'm gonna close this for now to make sure the pull queue is only populated with pulls that spark joy

PR #68, opened 2019-01-06
Handlers can be async

### LINKED ISSUE #14: async support
Given that the runtime is built on top of tokio, I think it would be useful to be able to provide async handler functions.

Happy to contribute this if there is interest from your side!

EDIT: Support for `tower` via the [`tower-service`](https://crates.io/crates/tower-service) crate could also be interesting.

illicitonion:
Fixes #14 

Handlers now return an `IntoFuture` (which `Result` implements), so that async handlers can be written. In particular, this is useful for using `rusoto`, because it means that its Futures will be executed on the tokio runtime.

This is mostly backwards compatible, because `Result` implements `IntoFuture`, but it adds a `Send + 'static` bound to a few types (notably `Handler`, `Event`) so should probably be considered a breaking change.

It also stops us calling `wait()` everywhere except the top-level, but it tries to make minimally invasive changes. At some point I'd be tempted to change all of the inline `panic!` calls to be Future error returns, and handle them at the top-level, rather than `panic!`ing all over the place.

Each commit is useful to separately review, as they are separated out building blocks. Sorry that the example I added is so contrived...!

- [✔] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [✔] I confirm that I've made a best effort attempt to update all relevant documentation.


sapessi:
Hi @illicitonion, thanks for the PR. We also planned to make the whole runtime async eventually. We were holding off for now because we have a pretty significant refactor in progress in #63. I'll @davidbarsky chime in here since he owns the async RFC.

davidbarsky:
Thanks for cutting this CR! I think this is an excellent start from my cursory overview. I was also looking at replacing the panics with a top-level panic handler, but fell down a rabbit hole with Tower integration.

README.md
```diff
@@ -174,18 +174,21 @@ For error reporting to the runtime APIs the library defines the `RuntimeApiError
...
-        event: E,
-        ctx: Context
-    ) -> Result<O, HandlerError>;
+    fn run(&mut self, event: E, ctx: Context) -> Self::IntoFuture;
 }
 ```
 
 `Handler` provides a default implementation that enables you to provide a Rust closure or function pointer to the `lambda!()` macro.
 
+If your handler is synchronous, you can just return a `Result` from it; if your handler is asynchronous, you can return a `Future` from it.
```
davidbarsky: Thanks for these docs! Small phrasing nit—I'd prefer avoiding using “just” in documentation. I'd instead emphasize that due to the `IntoFuture` implementation on `Result`, most synchronous handlers won't need to make any code changes.

lambda-http/src/lib.rs
```diff
@@ -70,14 +70,14 @@ use crate::{request::LambdaRequest, response::LambdaResponse};
 pub type Request = http::Request<Body>;
 
 /// Functions serving as ALB and API Gateway handlers must conform to this type.
-pub trait Handler<R> {
+pub trait Handler<R>: Send {
```
davidbarsky: Is `Send` not inferred on Handler?

lambda-http/src/lib.rs
```diff
@@ -92,14 +92,14 @@ where
 ///
 /// # Panics
 /// The function panics if the Lambda environment variables are not set.
-pub fn start<R>(f: impl Handler<R>, runtime: Option<TokioRuntime>)
+pub fn start<R>(f: impl Handler<R> + 'static, runtime: Option<TokioRuntime>)
```
davidbarsky: Two things: 

- I'm not completely sure that `impl Handler` needs to be `&'static`. I wonder if `start<R>` can be refactored to take an `impl Future` instead of an `impl Handler<R>`.
- If we're able to get rid of all `.waits()` inside the client, we can remove the optional `TokioRuntime` parameter.

lambda-runtime-client/src/client.rs
```diff
@@ -120,49 +120,44 @@ pub struct EventContext {
...
-    _runtime: Runtime,
     http_client: Client<HttpConnector, Body>,
     endpoint: String,
 }
 
 impl RuntimeClient {
     /// Creates a new instance of the Runtime APIclient SDK. The http client has timeouts disabled and
     /// will always send a `Connection: keep-alive` header.
-    pub fn new(endpoint: String, runtime: Option<Runtime>) -> Result<Self, ApiError> {
+    pub fn new(endpoint: String, task_executor: TaskExecutor) -> Result<Self, ApiError> {
```
davidbarsky: The signature can be changed to just return `Self`—we're not returning `ApiError` anywhere now (nice job that!)

lambda-runtime-client/src/client.rs
```diff
@@ -211,41 +207,44 @@ impl RuntimeClient {
...
-                        "Error from Runtime API when posting response for request {}: {}",
-                        request_id,
-                        resp.status()
-                    );
-                    return Err(ApiError::new(&format!(
-                        "Error {} while sending response",
-                        resp.status()
-                    )));
+        let http_client = self.http_client.clone();
+        uri.into_future()
```
davidbarsky: I didn't know that `Uri` had an `IntoFuture` implementation. Can I ask why you opted for this?

lambda-runtime-client/src/client.rs
```diff
@@ -211,41 +207,44 @@ impl RuntimeClient {
...
-                    error!(
-                        "Error from Runtime API when posting response for request {}: {}",
-                        request_id,
-                        resp.status()
-                    );
-                    return Err(ApiError::new(&format!(
-                        "Error {} while sending response",
-                        resp.status()
-                    )));
+        let http_client = self.http_client.clone();
```
davidbarsky: `http_client` might be able to be wrapped in an Arc<T>, which would make the clones cheaper, I believe. It might be an unnecessary micro-optimization.

lambda-runtime/src/runtime.rs
```diff
@@ -36,12 +55,16 @@ where
 ///
 /// # Panics
```
davidbarsky: Doc nit—can this be a non-H1 case?

lambda-runtime/src/runtime.rs
```diff
@@ -73,10 +96,14 @@ macro_rules! lambda {
...
-pub(crate) fn start_with_config<E, O, C>(f: impl Handler<E, O>, config: &C, runtime: Option<TokioRuntime>)
+pub(crate) fn start_with_config<E, O, C>(
+    f: impl Handler<E, O>,
+    config: &C,
+    task_executor: TaskExecutor,
+) -> impl Future<Item = (), Error = String> + Send
 where
-    E: serde::de::DeserializeOwned,
-    O: serde::Serialize,
+    E: serde::de::DeserializeOwned + Send + 'static,
```
davidbarsky: I'm think I'm missing why the event needs to be static—aren't we taking ownership of it here?

lambda-runtime/src/runtime.rs
```diff
@@ -179,7 +205,7 @@ impl<F, E, O> Runtime<F, E, O> {
         Ok(Runtime {
             runtime_client: client,
             settings: config,
-            handler: f,
+            handler: Arc::new(Mutex::new(f)),
```
davidbarsky: I'm not the biggest fan of having a mutex here. Can you elaborate why it's necessary?

lambda-runtime/src/runtime.rs
```diff
@@ -191,134 +217,169 @@ impl<F, E, O> Runtime<F, E, O> {
 impl<F, E, O> Runtime<F, E, O>
 where
     F: Handler<E, O>,
-    E: serde::de::DeserializeOwned,
-    O: serde::Serialize,
+    E: serde::de::DeserializeOwned + Send + 'static,
```
davidbarsky: See earlier comments about the `'static` bound w.r.t. ownership.

lambda-runtime/src/runtime.rs
```diff
@@ -191,134 +217,169 @@ impl<F, E, O> Runtime<F, E, O> {
...
-                                Ok(_) => info!("Response for {} accepted by Runtime API", request_id),
-                                // unrecoverable error while trying to communicate with the endpoint.
-                                // we let the Lambda Runtime API know that we have died
+
+        let max_retries = self.max_retries;
+        let runtime_client = self.runtime_client.clone();
+        let settings = self.settings.clone();
+        let handler = self.handler.clone();
+
+        loop_fn((), move |()| {
```
davidbarsky: We might have the opportunity to refactor this chunk of code into a bunch of smaller and more methods & futures.

lambda-runtime/src/runtime.rs
```diff
@@ -191,134 +217,169 @@ impl<F, E, O> Runtime<F, E, O> {
...
-                    None => {
-                        self.runtime_client.fail_init(&err);
+    /// A `Future` resolving to the next `Event` object to be processed.
+    pub(super) fn get_next_event(
+        max_retries: i8,
+        runtime_client: RuntimeClient,
+        settings: FunctionSettings,
+    ) -> impl Future<Item = (E, Context), Error = String> {
+        loop_fn(
+            (0, None),
```
davidbarsky: I was thinking of using `tower-retry` for retries in a slightly less ad-hoc way (e.g., https://gist.github.com/seanmonstar/2469a8e544ada2d31d13c8ee54f17d48). I'll talk to the Tower folks about getting them on crates.io. I've been doing that locally with some success.

softprops:
Futures retry exists and doesn't depend on anything but the futures crate https://mexus.gitlab.io/futures-retry/futures_retry/

lambda-runtime-client/src/client.rs
softprops: I think uri here is a Result and all Result types get a free `IntoFuture` impl. Since the return type changed the ? needed to be removed above

README.md
illicitonion: Rephrased

lambda-http/src/lib.rs
illicitonion: Because `Handler` is a `trait`, there's no way for the type system to know that an implementation of it doesn't have some non-`Send` fields. We can either add the bound here to require that implementors are `Send`, or we can add a `+ Send` bound in each of the places that requires `Handler` be `Send`, but doing the latter would require adding the bound to several core functions, and would lead to more confusing error messages if you had a non-`Send` trait.

lambda-http/src/lib.rs
illicitonion: I'll hold off on looking at this until #63 merges, if that's ok

lambda-runtime-client/src/client.rs
illicitonion: Done

lambda-runtime-client/src/client.rs
illicitonion: Yeah; I could change this to `future::done(uri)` if that's more clear?

lambda-runtime-client/src/client.rs
illicitonion: Right now cloning a `Client` is 3 `Arc` clones and ~6 word copies, so yeah, probably a bit micro for now, but I can do it if you'd like :)

lambda-runtime/src/runtime.rs
illicitonion: Sorry, I'm not quite sure what you're asking for here...

lambda-runtime/src/runtime.rs
illicitonion: We're taking ownership of the instance of `E` - the `'static` here is stating that that instance's lifetime doesn't depend on any other (non-static) lifetimes; otherwise the `E` may have a reference to some other data in it which could be dropped, which would mean the `E` instance was no longer valid. It's a bound on whether `E` itself can have borrows, rather than on how long the `E` instance actually lives.

lambda-runtime/src/runtime.rs
illicitonion: Because `run` takes a `&mut self`, we're only allowed to call `run` from one thread at a time. This `Mutex` is how we guarantee that multiple threads don't call `run`.

If we wanted to drop the `Mutex`, we would need to either:
1. Change `run` from taking `&mut self` to just taking `&self` (this would mean we could only implement it for `Fn` not for `FnMut` like we currently do). This would be easy, but an API change.
2. Somehow guarantee that we only have one thread which calls a `Handler`. This is what we were doing before (by blocking the executions); if we wanted to do this, we could probably use some kind of same thread executor, but I imagine that would be awkward for actually writing handlers, because they couldn't necessarily rely on the runtime they submit tasks to being able to do more than one thing at a time.

lambda-runtime/src/runtime.rs
illicitonion: Yeah, this could very reasonably become a series of nicely tightly scoped `map`s and `and_then`s in a follow-up change :)

lambda-runtime/src/runtime.rs
illicitonion: I'll happily give this a go when the crate is published! :)

davidbarsky: Sounds good!

saks:
Any progress on this?

nappa85:
Sorry, didn't saw this PR before making mine...

davidbarsky:
@nappa85 you're fine! don't worry about it.

illicitonion:
> Sorry, didn't saw this PR before making mine...

No worries @nappa85! I haven't had time to fix this up, and something that's done and works is definitely better than something that's part-finished! :)

softprops:
I think this can be closed now. First class async support now exists in master

PR #80, opened 2019-01-28
README: Update example in readme to match basic.rs example

Luke-Nukem:
Also removes reference to `new_error()` in client.rs `event_error`
doc comment as this function is removed from `Context`

- [ x ] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [ x ] I confirm that I've made a best effort attempt to update all relevant documentation.


davidbarsky:
Thanks for the catch! I've disabled skeptic a while ago; can you verify that the changes still work?

Luke-Nukem:
Yeah changes work fine, tested example before PR. You can also propagate errors like this:

``` rust
return Err("A new error")? // returns as an `Unknown` error with string.
```
or

``` rust
Error::new(ErrorKind::Other, "Failed to sign up user") // Error and ErrorKind from relevant std
```

The first example means you don't really need the `simple-error` crate at all. The `?` operator calls the `From` trait functions for us.

davidbarsky:
Perf, okay. I'll re-introduce skeptic for the examples in the readme once I merge this.

Issue #89, opened 2019-02-20
Panic at 'Could not retrieve next event'


LeakyBucket:
I thought maybe it was something in my code but I get the same panic with the example here:  https://docs.rs/crate/lambda_runtime/0.2.0

```
START RequestId: 967aad3f-eaaa-4946-bc51-dc717f81aa71 Version: $LATEST
thread 'main' panicked at 'Could not retrieve next event', /Users/redacted/.cargo/registry/src/github.com-1ecc6299db9ec823/lambda_runtime_core-0.1.0/src/runtime.rs:266:17
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at src/libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at src/libstd/sys_common/backtrace.rs:59
             at src/libstd/panicking.rs:211
   3: std::panicking::default_hook
             at src/libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at src/libstd/panicking.rs:491
   5: std::panicking::begin_panic
   6: <lambda_runtime_core::runtime::Runtime<Function, EventError>>::get_next_event
   7: <lambda_runtime_core::runtime::Runtime<Function, EventError>>::get_next_event
   8: <lambda_runtime_core::runtime::Runtime<Function, EventError>>::get_next_event
   9: <lambda_runtime_core::runtime::Runtime<Function, EventError>>::get_next_event
  10: <lambda_runtime_core::runtime::Runtime<Function, EventError>>::get_next_event
  11: lambda_runtime_core::runtime::start_with_config
  12: bootstrap::main
  13: std::rt::lang_start::{{closure}}
  14: std::panicking::try::do_call
             at src/libstd/rt.rs:59
             at src/libstd/panicking.rs:310
  15: __rust_maybe_catch_panic
             at src/libpanic_unwind/lib.rs:102
  16: std::rt::lang_start_internal
             at src/libstd/panicking.rs:289
             at src/libstd/panic.rs:398
             at src/libstd/rt.rs:58
  17: main
END RequestId: 967aad3f-eaaa-4946-bc51-dc717f81aa71
[...]

PR #91, opened 2019-02-26
Panic when X-Ray header not present

sapessi:
Addressing panic when the X-Ray header is not present. This can happen when X-Ray is not available in the region - for example GovCloud.

*Issue #, if available:* #89 

*Description of changes:*
* Changed the context objects in the library to use `Option<String>` for the trace id
* Changed parsing of header value in the runtime client
* Added basic unit tests.


lambda-runtime-client/src/client.rs
```diff
@@ -461,3 +471,42 @@ fn header_string(value: Option<&HeaderValue>, header_type: &LambdaHeaders) -> Re
...
+    use chrono::{Utc, Duration};
+
+    fn get_headers() -> HeaderMap<HeaderValue> {
+        let mut headers: HeaderMap<HeaderValue> = HeaderMap::new();
+        headers.insert(LambdaHeaders::RequestId.as_str(), HeaderValue::from_str("req_id").unwrap());
+        headers.insert(LambdaHeaders::FunctionArn.as_str(), HeaderValue::from_str("func_arn").unwrap());
+        headers.insert(LambdaHeaders::TraceId.as_str(), HeaderValue::from_str("trace").unwrap());
+        let deadline = Utc::now() + Duration::seconds(10);
+        headers.insert(LambdaHeaders::Deadline.as_str(), HeaderValue::from_str(&deadline.timestamp_millis().to_string()).unwrap());
+        //headers.insert(LambdaHeaders::ClientContext.as_str(), HeaderValue::from_str("{}").unwrap());
```
davidbarsky: Meant to be commented out?

sapessi: Yes. I do not need them for these tests. Left them there in case we need to set them up for future tests. Should I remove?

davidbarsky: Yeah, I'm generally of the opinion that we should remove commented out code—we can always get it back via git.

PR #106, opened 2019-05-10
async support

nappa85:
Issue #14

The main loop has been converted into a future::LoopFn, who takes ownership on the Runtime.
Now it is possible to use async code inside lambdas.


davidbarsky:
@nappa85: Thanks for opening this! I think allowing folks to write async Lambda functions is a good idea, but given the upcoming stabilization of async/await and std::future, I was thinking of introducing a clean break and enforcing async/await first: https://paper.dropbox.com/doc/Lambda-Runtime-for-Rust-0.3--Ac2u75PLCibbXervTP2Rv5z_AQ-jLrDOiKk815YyIHLWva7M. Any thoughts?

cc: @softprops @iliana @sapessi

softprops:
The vision looks super dope! 

What I like about this pull is its relatively transparent for users to upgrade to the next version of the interface in this pull returning something that implements IntoFuture which gets the best of both worlds because todays Result interface should just work in most cases. That enables new future returning funcs and for sync funcs it should "just work" and no added syntactic cost.

I think the vision in the linked document could be possible even with this implementation as a graduating step. What I really liked in the doc was that it touched on a more painful part of todays interface, a less strict error representation! Satisfying HandlerError types have been a much more economic pain point than block_on calls.

I'm a fan of where the async/await syntax is going but I'm an even bigger fan of using a compromise that works on stable Rust today ;)

davidbarsky:
Oh, I’m really sorry! I’m in favor of merging this change, and I have no idea why I said “introducing async/await first!” I must’ve not been paying attention when typing. Really sorry!


________________________________

nappa85:
Sorry for late reply.
Yes, my intent was to offer a drop-in upgrade, so existing implementations should not be affected, while giving everyone the ability to use futures-powered-lambdas.
I know about async/await stabilization will probably happen in July with 1.36, you can release this as a version bump now and start working on a refactory that will end up in a new major version to be released after July.

lambda-runtime/src/lib.rs
```diff
@@ -115,16 +142,18 @@ where
 #[macro_export]
 macro_rules! lambda {
     ($handler:ident) => {
-        $crate::start($handler, None)
+        use tokio;
+        tokio::run($crate::start($handler))
```
iliana: This requires a user to add `tokio` to their Cargo.toml even though they might not otherwise use it directly. Is there a way to make this experience better here?

saks: Good point @iliana! 
There is a trick that will let us avoid forcing user to have explicit dependency here. There are two steps:
1. re-export `tokio` form this crate:
```rust
#[doc(hidden)]
pub use tokio;
```
2. then use tokio in the macro body like:
```rust
$crate::tokio::run($crate::start($handler))
```

@nappa85, what do you think?

softprops:
I think that's a common tactic for macro libs that depend on third party crates

lambda-runtime/src/lib.rs
softprops: I think that's a common tactic for macro libs that have third party crate dependencies

nappa85: Yes, you're right, my bad for not thinking about it first, I'll fix it immediately.

lambda-runtime/src/lib.rs
```diff
@@ -99,14 +126,16 @@ where
...
-    Event: serde::de::DeserializeOwned,
+    Event: serde::de::DeserializeOwned + Send,
     Output: serde::Serialize,
     EventError: Fail + LambdaErrorExt + Display + Send + Sync,
+    I: IntoFuture<Item=Output, Error=EventError> + Send,
+    I::Future: Send + 'static,
 {
     let wrapped = wrap(f);
-    start_with_config(wrapped, &EnvConfigProvider::default(), runtime)
+    start_with_config(wrapped, &EnvConfigProvider::default()/*, runtime*/)
```
davidbarsky: Small nit—can the commented out runtime be removed?

lambda-runtime/src/lib.rs
```diff
@@ -40,29 +40,34 @@ use lambda_runtime_core::{start_with_config, EnvConfigProvider, HandlerError, La
 use serde;
 use serde_json;
 use std::fmt::Display;
-use tokio::runtime::Runtime as TokioRuntime;
+// use tokio::runtime::Runtime as TokioRuntime;
```
davidbarsky: Nit: can you remove this commented out code?

lambda-runtime/src/lib.rs
```diff
@@ -40,29 +40,34 @@ use lambda_runtime_core::{start_with_config, EnvConfigProvider, HandlerError, La
...
+use tokio::prelude::future::{self, Future, IntoFuture};
 
 /// The error module exposes the HandlerError type.
 pub mod error {
     pub use lambda_runtime_core::{HandlerError, LambdaErrorExt, LambdaResultExt};
 }
 
 /// Functions acting as a handler must conform to this type.
-pub trait Handler<Event, Output, EventError> {
+pub trait Handler<Event, Output, EventError, I>: Send {
```
davidbarsky: Instead of `I`, can this be called `Fut` to align with the other generic parameters?

lambda-runtime/src/lib.rs
davidbarsky: Thanks for this!

lambda-runtime/src/lib.rs
nappa85: Sure.
I had left the commented out runtime code as a trace of the edit, for debugging purposes.
I renamed the "I" into Fut everywhere, so it's coherent

nappa85:
Hello, sorry to bother you all...
Are there still open tasks blocking this merge?
Thank you

davidbarsky:
Not a bother! I don’t have any additional concerns merging this and would be happy if @sapessi or @softprops don’t have any concerns.


________________________________

nappa85:
Bump
A month has passed, still no news?

lvicentesanchez:
hi! what's the status of this PR? I would really like to start using async lambdas.

lvicentesanchez:
@davidbarsky  @sapessi  @softprops ?

mies:
any update on this @davidbarsky @nappa85 ?

sbditto85:
bump. update?

drusellers:
Just checking in. :)

davidbarsky:
You should be able to pull from the master branch and take this for a spin. It doesn’t suck.

softprops:
I think this can be closed now

brainstorm:
Yes, this can be closed, async works very well on rust-lambda for a while now 🚀

coltonweaver:
Going to close this PR as it seems all is well here!

PR #111, opened 2019-06-28
Lambda with Async/Await

davidbarsky:
This PR is a simplification of the Lambda Runtime for Rust. You can now write a Lambda function similar to the following:

```rust 
use lambda::lambda;
use serde_json::Value;

#[lambda]
#[tokio::main]
async fn main(event: Value) -> Result<Value, Error> {
    let _ = lambda::context();
    Ok(event)
}
```

These are the high-level changes:

- Async/await is the primary interface within a Lambda function. While Lambda does not support concurrent invocations in a single container (concurrency is the responsibility of Lambda-the-service), async/await _does_ allow for concurrent calls to other services.
- Functions no longer need to return a `LambdaError`, as the the stabilization of [`std::any::type_name`](https://doc.rust-lang.org/std/any/fn.type_name.html) removed the need for a Lambda-specific error type.
- Lambda functions can be created and started through the `#[lambda]` annotation. Note that starting a futures executor is still necessary. In the above example, this is handled through the `#[tokio::main]` macro.
- The execution context is not an argument to the handler. Instead, it is a task-local, which is similar to a [thread local](https://doc.rust-lang.org/std/thread/struct.LocalKey.html), but specific to a future instead of a thread. The task local can be fetched through free function `lambda::context()`.

### The Handler Trait

For greater control over a handler, this PR provides a `Handler` trait. It is defined as:

```rust
pub trait Handler<A, B> {
    /// Errors returned by this handler.
    type Err;
    /// The future response value of this handler.
    type Fut: Future<Output = Result<B, Self::Err>>;
    /// Process the incoming event and return the response asynchronously.
    fn call(&mut self, event: A) -> Self::Fut;
}
```

This is _not_ implemented in this PR yet, but I'd like to provide a blanket implementation of [`tower::Service`](https://docs.rs/tower/0.3.0/tower/trait.Service.html), which would provide compatibility with Tower's rich ecosystem of middleware. In particular, this would enable the use of [`tracing`](https://github.com/tokio-rs/tracing) for rich logging and trace reporting.

### Without Macros

For those that prefer to not make use of the `#[lambda]` macro, a Lambda function can be started with the following code:

```rust
use lambda::handler_fn;

#[tokio::main]
    let func = handler_fn(func);
    lambda::run(func).await?;
    Ok(())
}

async fn func(event: String) -> Result<String, Error> {
    Ok(event)
}
```

### Remaining Work

For this PR to land, there are some additional bits of work needed: 
- Update documentation to show the usage of this runtime, especially some of the more advanced features.
- Guide customers to _only_ build against the `x86_64-unknown-linux-musl` target. The glibc in the Lambda execution environment is too old to run Rust executables.
- Convert `Executor::run` (see `lambda/src/lib.rs`) to a stream so that Lambda functions can be easily unit-tested _in-memory_. By converting it to a stream, limiting execution via [`SteamExt::take`](https://docs.rs/futures/0.3.1/futures/stream/trait.StreamExt.html#method.take) or [stream-cancel](https://docs.rs/stream-cancel/0.5.0/stream_cancel/index.html) becomes _much_ easier.
- Provide examples (beyond this library's unit tests) of how to test a Lambda function.
- Provide examples of using the runtime with [tokio-compat](https://github.com/tokio-rs/tokio-compat) and Rusoto, if Rusoto doesn't support async/await prior to this PR landing.
- Figure out how to support `http`-based Lambdas through the `#[lambda]` macro. I think an attribute like `#[lambda::http]` dispatching to an `Executor::run_http` method is the best approach until specialization is stable.
- Release `tracing`/AWS X-Ray integration.

cc: @softprops @iliana @sapessi.


davidbarsky:
Wrote a draft of the Handler that supports the `http` crate and Serde in a single trait: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3e6b636e053eb59b7fef67bd29834770

davidbarsky:
Minor update: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=7ada5aa87c5a5a2c9d597ab7feeeaad2. Still thinking this through.

lambda-attributes/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+version = "0.1.0"
+authors = ["David Barsky <dbarsky@amazon.com>"]
+edition = "2018"
+
+[lib]
+proc-macro = true
+
+[dependencies]
+syn = { version = "0.15.33", features = ["full"] }
+proc-macro2 = { version = "0.4.29", features = ["nightly"] }
```
softprops: Does the `nightly` feature prevent building and deploying binaries on a `stable` channel. If so is there a way we could avoid that. I really appreciate stable builds when it comes to things others depend on :)

lambda-attributes/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+quote = "0.6.12"
+runtime = "0.3.0-alpha.5"
```
softprops: Is there a less than alpha release coming around the corner. For the same reason I like shipping on the stable channel, I feel less at easy with something annotated with with `alpha`.

lambda-attributes/Cargo.toml
```diff
@@ -0,0 +1,18 @@
...
+bytes = "0.4.12"
+
+[dev-dependencies]
+runtime = "0.3.0-alpha.5"
```
softprops: if `runtime` is an the main dependency path, I believe you get it for free as a dev-dependency

lambda/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "lambda"
+version = "0.1.0"
+authors = ["David Barsky <dbarsky@amazon.com>"]
+edition = "2018"
+
+[dependencies]
+futures-preview = "0.3.0-alpha.16"
```
softprops: more alpha dependencies. are these going to have a short alpha shelf live and with quick follow up stable release versions?

lambda/examples/hello-with-ctx.rs
```diff
@@ -0,0 +1,11 @@
+#![feature(async_await)]
+
+use lambda::{lambda, LambdaCtx};
+type Err = Box<dyn std::error::Error + Send + Sync + 'static>;
```
softprops: Can we export this definition at the top level of the  `lambda` crate? Also I have a slight preference for `Error` because its a bit more common and `Err` clobbers the result `Err` variant which makes it working with types in the standard scope more awkward that it needs to be.

lambda/examples/hello-with-ctx.rs
```diff
@@ -0,0 +1,11 @@
+#![feature(async_await)]
```
softprops: Is this a nightly only feature still. If so is there a way to make this optional. I _really_ like stable systems and reliable ci builds for things other depend on.

lambda/examples/hello-without-macro.rs
```diff
@@ -0,0 +1,15 @@
...
+type Err = Box<dyn std::error::Error + Send + Sync + 'static>;
+
+#[runtime::main]
+async fn main() -> Result<(), Err> {
+    let func = handler_fn(func);
+    lambda::run(func).await?;
+    Ok(())
+}
+
+async fn func(event: String, _: Option<LambdaCtx>) -> Result<String, Err> {
```
softprops: I feel like `Option<LambdaCtx>` isn't much of an improvement over `_: LambdaCtx` api-wise. I feel like you can slim the api and let rust standard idioms kick in over for cases where you want to ignore an arg

lambda/src/client.rs
```diff
@@ -0,0 +1,167 @@
...
+            Ok(res)
+        };
+
+        fut.boxed()
+    }
+}
+
+/// The `Stream` implementation for `EventStream` converts a `Future`
+/// containing the next event from the Lambda Runtime into a continuous
+/// stream of events. While _this_ stream will continue to produce
```
softprops: super clean approach. nice

lambda/src/lib.rs
```diff
@@ -0,0 +1,241 @@
+#![feature(async_await)]
```
softprops: does this demand prevent being able to ship lambdas on stable release channels?

lambda/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+use serde::{Deserialize, Serialize};
+use std::{convert::TryFrom, env};
+
+mod client;
+/// Mechanism to provide a custom error reporting hook.
+pub mod error_hook;
+/// Types availible to a Lambda function.
+mod types;
+
+type Err = Box<dyn std::error::Error + Send + Sync + 'static>;
```
softprops: Having read further it seems you could just `use lambda::Err;`  in the examples. Same deal hear. I have a strong preference for not clobbering predef type names like `Result`s `Err` variant. What would it take to convince you to try `Error`? I feel like that's more common for crate apis.

lambda/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+
+#[derive(Debug)]
+/// A string error, which can be display
+pub(crate) struct StringError(pub String);
+
+impl std::error::Error for StringError {}
```
softprops: Isn't it lovely this is all it takes to impl std `Error` these days? :)

lambda/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// Errors returned by this handler.
+    type Err: Into<Err>;
+    /// The future response value of this handler.
+    type Fut: Future<Output = Result<Output, Self::Err>>;
+    /// Process the incoming event and return the response asynchronously.
+    ///
+    /// # Arguments
+    /// * `event` - The data received in the invocation request
+    /// * `ctx` - The context for the current invocation
+    fn call(&mut self, event: Event, ctx: Option<LambdaCtx>) -> Self::Fut;
```
softprops: Still not super sold on lambdas that don't have contexts when in practice they all do vs functions just ignoring the ctx arg as `func(ev: Event, _: LambdaCtx)`.  If its gotta be optional we may as well document why for those that chose to implement `Handler`

lambda/src/lib.rs
```diff
@@ -0,0 +1,241 @@
...
+///     Ok(())
+/// }
+///
+/// async fn func(event: String, _ctx: LambdaCtx) -> Result<String, Err> {
+///     Ok(event)
+/// }
+/// ```
+pub async fn run<Function, Event, Output>(
+    mut handler: Function,
+) -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>>
```
softprops: It could maybe clean this up by re-usng `lambda::Err`. It would also help for refactoring to update in on place.

lambda/src/types.rs
```diff
@@ -0,0 +1,377 @@
...
+    /// empty unless the function is invoked using an AWS mobile SDK.
+    pub client_context: Option<ClientContext>,
+    /// The Cognito identity that invoked the function. This field is empty
+    /// unless the invocation request to the Lambda APIs was made using AWS
+    /// credentials issues by Amazon Cognito Identity Pools.
+    pub identity: Option<CognitoIdentity>,
+    /// Lambda function configuration from the local environment variables.
+    /// Includes information such as the function name, memory allocation,
+    /// version, and log streams.
+    pub env_config: Config,
```
softprops: I haven't seen this info expose this way. It benefits users who move around lambdas using different runtimes to work with types of roughly the same shape. 

https://docs.aws.amazon.com/lambda/latest/dg/python-context-object.html
https://docs.aws.amazon.com/lambda/latest/dg/nodejs-prog-model-context.html
https://docs.aws.amazon.com/lambda/latest/dg/java-context-object.html
https://docs.aws.amazon.com/lambda/latest/dg/go-programming-model-context.html

if env_config is an impl detail you could keep this `pub(crate)` and expose `LambdaCtx` methods that provide a more std interface to the context object. What makes that awkward and would give me pause is that I like having the ability to construct these api types in unit tests so I don't know what the constructor for `LambdaCtx` would look like `LambdaCtx::from_config(Config::default())`??? may be worth some though before making it `pub`

lambda/src/types.rs
```diff
@@ -0,0 +1,377 @@
...
+    use crate::types::{
+        ClientApplication, ClientContext, CognitoIdentity, FunctionArn, InvocationDeadline,
+        MobileClientContext, MobileClientIdentity, RequestId, XRayTraceId,
+    };
+    use bytes::Bytes;
+    use headers::{HeaderMap, HeaderMapExt};
+    use http::Response;
+    use proptest::{collection, option, prelude::*, strategy::Strategy, string::string_regex};
+
+    fn gen_request_id() -> impl Strategy<Value = RequestId> {
```
softprops: This reminds me that I to carve some time and look into proptest. I'm a fan of scalacheck from the scala universe. Having this in rust is super appealing to me and to my internal company sales pitch on the awesomeness of rust

lambda/tests/compile-fail/no-args.stderr
```diff
@@ -0,0 +1,11 @@
+error: The #[lambda] macro can accept one or two arguments.
+ --> $DIR/no-args.rs:3:1
```
softprops: super awesome! I've never seen tests for failing compiles. I'm add this to my tool belt the next time I need something like this.

lambda-attributes/Cargo.toml
davidbarsky: It shouldn't—this is to mostly support async/await until 1.38 is available in Beta.

lambda-attributes/Cargo.toml
davidbarsky: Not sure. I intend to ship _this_ async/await-enabled runtime as alpha/beta on crates.io until async/await is available on the stable channel of Rust.

lambda/Cargo.toml
davidbarsky: That's the plan, yes—most of the async/await-enabled libraries have `-alpha` attached to them.

lambda-attributes/Cargo.toml
davidbarsky: That sounds correct—I don't think the `runtime` should be a regular dependency of `lambda-attributes` anyways.

lambda/examples/hello-with-ctx.rs
davidbarsky: I'm happy to rename it to `Error`. I'm not 100% sure on exporting this type alias, but I can go either way on it!

lambda/src/client.rs
davidbarsky: Thank you!

lambda/examples/hello-without-macro.rs
davidbarsky: Not really, but I'm not sure how represent an optional `LambdaCtx` arg in a proc macro anyhow.

lambda/examples/hello-with-ctx.rs
davidbarsky: Yes, `async/await` is nightly-only, as the feature is being stabilized as part in Rust 1.38. IWe might be able to remove this requirement, but a lot of the ecosystem surrounding `std::future::Future` assume the presence of `async/await`.

In practice, I've found that async/await have been _pretty_ stable for me, but I did approach it with _some_ trepidation.

lambda/src/lib.rs
davidbarsky: See my response above—technically, no, but in practice, due to where the broader ecosystem is, yes.

lambda/src/lib.rs
davidbarsky: Would be happy to rename to `Error`! I just like `Err` in my personal code, but I'm not attached to this typealias whatsoever :)

lambda/src/lib.rs
davidbarsky: The error reform RFC was _great_. I can't wait until we can get backtraces in `std` too, removing the need for `Failure`~

lambda/src/lib.rs
davidbarsky: I'm not a fan either, to be honest—that's why the task-local approach feels so appealing to me.

lambda/src/lib.rs
davidbarsky: Thanks, will do!

lambda/src/types.rs
davidbarsky: Yeah, for sure—@sapessi suggested throwing the environment config throw into here, but I think customers keen on getting _that_ environment config should be able to do so themselves.

Broadly, I expect to _reduce_ how many fields are `pub`, this one included.

lambda/src/types.rs
davidbarsky: Proptest is great—I'm trying to write a proc macro for proptest to more closely mimic hypothesis that looks like:

```rust
#[proptest(gen_function_arn())]
fn function_arn(arn: FunctionArn) {
    let mut headers = HeaderMap::new();
    headers.typed_insert(req.clone());
    prop_assert_eq!(headers.typed_get::<FunctionArn>(), Some(req));
}
```

...because I'm not a fan of the `macro_rules!`-style macros, for very dumb reasons :)

lambda/tests/compile-fail/no-args.stderr
davidbarsky: It's very nice! My biggest complaint is that `trybuild` _runs_ proc macros, not just checks if they compile correctly—is results in tests failing because the Lambda runtime APIs aren't up-and-running :).

davidbarsky:
> I'm hesitant to expose and apigw/alb interface in the default crate. It increases the surface area and crate weight where for cases where it's not needed. In lambda it always seem to benefit users to pack light. Lifting to a sibling crate makes more sense to me.

I'm coming around to that. Unfortunately, I'm sure of a way that allows both:
- `lambda-attributes` to be agnostic of ALB/API Gateway events,
- Allow customers to use the `http::{Request, Response}` types directly.

If had to pick one or the other, I'd prioritize allowing customers to use `http::Request` and `http::Response` directly.

davidbarsky:
I accidentally changed a prior commit and couldn't find a way to undo it, hence the force push. Sorry!

lambda/examples/hello-with-ctx.rs
davidbarsky: Renamed!

lambda/src/lib.rs
davidbarsky: Renamed!

lambda/src/types.rs
davidbarsky: Push this macro—I should share it upstream to proptest.

lambda-attributes/Cargo.toml
davidbarsky: Fixed, kinda.

lambda/examples/hello-with-ctx.rs
softprops: I understand that's the place the bleeding edge community is at. For hobby projects that's find. From an organizational use context this makes rust a harder sell when an applications state of deployability can change from one day to the next.

We've been bit in the past where we've needed to deploy something and a nightly dependency had us hostage. It's not a good place to be in when you're making pitches for using rust in an org context :/

If there's any way to retain a stable version of the rust runtime with nightly features behind cargo feature flags that would be great! I could really the new error api but don't yet _need_ async/await. If there is a way to separate those two features it seems everyone wins!

davidbarsky: Gotcha, thanks for explaining! I think that the proposed changes in this PR really consist of _three_ distinct breaking changes. They are:

1. The `#[lambda]` attribute macro, which replaces the `lambda!` `macro_rules!`-based macro.
2. The new error API.
3. Support for `std::future::Future` and async/await.

I've ordered those changes in increasing difficulty of backporting to the existing, non-concurrent Lambda. In order:

1. The `#[lambda]` macro is the easiest change to backport.
2. The new error API is slightly more difficult to backport, but it would likely have the greatest quality of life increase for existing Lambda users. I'll try to backport it in the next two weeks (I'm traveling quite a bit), but feel free to do it yourself & remove the LambdaError-related crates. This implies a breaking change, but that doesn't worry me—we're nowhere near as foundational as other libraries.
3. While `std::future::Future` will be stabilized on Thursday, July 4, I'm prefer to avoid writing any sort of asynchronous Rust without async/await, as the compiler errors _really_ not great. To wit, I abandoned my 0.1 Futures-based asynchronous rewrite _because_ the diagnostics from `rustc` were so poor. Therefore, I think support for `std::future::Future` and async/await should come in the same release, about three months from now.

In summary, I'm not 100% confident that we can support nightly behind a feature flag, but I _do_ think its feasible to backport _some_ of the changes onto stable Rust. If I'm not mistaken, the easiest changes to backport are the ones you're most interested in.

davidbarsky:
Another possible take: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5b370e1dd9ecd9c87a4c15ce50be954a

I'm coming around to the idea that we'll need specialization to tackle both `A` and `http::Request<A>`  in using the trait system, but we might be able to sidestep the specialization requirement with an `Into<Param>`, where Param support both `A` and `http::Request<A>`.

lambda/src/lib.rs
sapessi: Apologies for the delay - agree on this. We should get rid of the `Option`.

lambda/src/types.rs
sapessi: I've thought about this for a while now. We want to make sure our runtime is friendly to Rust developers by following the standard language semantics. However, in this case these "APIs" are owned by the Lambda service and they have set a convention with all the other runtimes of exposing the config values as a property in the context object. I don't think we should break that.

sapessi:
The last high-level question we should answer is whether the http handler type should be bundled in this crate or not. Thoughts @softprops, @davidbarsky?

softprops:
I think it should be opt in to keep the surface area small to enable cases where you want to distribute light artifacts. Preferably as another crate for discoverability or behind a cargo feature flag. Feature flags in practice make for harder to maintain systems for ci build matrix combos. Feature flags tend to be less user friendly for those coming from non rust backgrounds but are familiar with the concept of npm and jar dependencies.

We are already asking users to bundle the runtime as part of the application which most other runtimes don't. I would consider it a disservice to be asked to bundle more than what I actually need to run a simple function on top of that. My ci system turnaround time also  appreciates giving less source for rustc to compile!

If we do bundle HTTP abstractions in core, it begs other questions like: why _aren't_ we bundling abstractions for every other triggerable event? Which makes the case for those that what to build small artifacts even harder obtain.

Not including it now means we can move forward _today_ with all of the other concrete problems this branch solves. I'm a fan of this because my biggest pain point has been the current error api.

davidbarsky:
Thanks for writing that up, @softprops. I'll try to respond point by point.

> We are already asking users to bundle the runtime as part of the application which most other runtimes don't. I would consider it a disservice to be asked to bundle more than what I actually need to run a simple function on top of that. My ci system turnaround time also appreciates giving less source for rustc to compile!

The thing about introducing an `http`-specific handler is that the `http` crate is _already_ bundled as part of the Hyper dependency. Introducing a top-level `http` handler should have a minimal, if any, impact on overall builds, but I'd be happy to benchmark that to verify that hypothesis.

> If we do bundle HTTP abstractions in core, it begs other questions like: why aren't we bundling abstractions for every other triggerable event? Which makes the case for those that what to build small artifacts even harder obtain.

In this lambda function, API, I'd like to be able to support `A` and `http::Request<A>`, where `A: for<'de> serde::Derserialize`. There are several obstacles:

- `http::Request<_>` cannot and _should not_ implement AWS-specific (e.g., API Gateway/ALB) deserialization logic. 
- While `http::Request<_>` _might_ implement `for<'de> serde::Derserialize` in the future, it does not at the moment. This requires us to write an type constraint similar to the following pseudo-code:

```rust
T<U>
where
    T: !for<'de> serde::Derserialize,
    U: for<'de> serde::Derserialize`
```

That API requires the usage of negative trait bounds, which will not land on stable Rust this year, and possibly several _other_ advanced trait features that depend on [rust-lang/chalk](https://github.com/rust-lang/chalk) landing in the stable compiler.

Given that `http::{Request, Response}` is the most common, non-directly (de)serializable type—appearing in about a third of all Lambda invocations—I believe that there _is_ value in providing first-class support for _this_ non-(de)serializable type in this runtime. We _are_ able to approximate a similar API using the `#[lambda]` procedural macro, which requires _some_ sort of coupling _somewhere_ in this repository. If I were to propose a rule for what event types are bundled-in by default, it'd be:

A specialized event type will be bundled as part of the core Lambda crate if and only if the following hold true:
- The idiomatic crate providing that event type _cannot_ be described using a `A: for<'de> Deserialize`.
- The trait system features required to support this are unavailable on stable Rust (for instance, if this were a Haskell runtime, I'd rely on `GeneralizedNewtypeDeriving` or `DerivingVia` for the `http` event type).
- The event type is a commonly used event type, exceeding—for instance—20% of all Lambda invocations.
- There exists no reasonable default (de)serialize implementation for the event type in question.

At the moment, only `http::{Request, Response}` fit that criteria. No other event type that I can think of has that property.

> Not including it now means we can move forward today with all of the other concrete problems this branch solves. I'm a fan of this because my biggest pain point has been the current error api.

I don't want to block release of the error API on deciding whether we want to support the `http` crate within the main runtime, nor do I think we _should_. At the very least, the error API is can be backported to the existing 0.2 API, but I haven't had the time to open a PR do so.

softprops:
@davidbarsky these were helpful responses. I just want to make I understand with some concrete examples.

> http::Request<_> cannot and should not implement AWS-specific (e.g., API Gateway/ALB) deserialization logic.

For apigw events, does that mean we'd end up with something like `http::Request<ApiGWRequest<UserDefinedBody>>` vs todays `http::Request<UserDefinedBody>>`  + RequestExt to expose ApiGw specific features like preparsed path and query parameters? The same for `http::Response<ApiGWResponse<UserDefinedBody>>`.

I feel like the latter is more ergonomic and possible with todays layered approach where the interface exposed wraps an internal delegation to the to core interface. 

Something worth considering is what if instead of a single proc macro that requires knowledge of dealing all possible event types to dispatch logic on, there could be specialized proc macros that dealt with specific cases `#[xxx::lambda]`. This may remove some of the type system limitations your bumping into as well as the requirements for the core crate to have detailed knowledge of specific event types. This comes at the cost of the boiler plate that comes along with implementing attributes which may be a heavy enough tax to scrap this idea.

If the api did look like `http::Request<ApiGWRequest<UserDefinedBody>>` would that make other non apigw events representable as `http::Request<CloudWatchEvent>` and is that desirable? I can see how it would be possible but from a user and api perspective, it doesn't seem to align with any of the other runtimes' programming models. I'm wondering if the `http::Response<OtherEventResponse>` would muck with that event type's trigger response when serializing the response back to lambda for sync event triggers

> At the very least, the error API is can be backported to the existing 0.2 API, but I haven't had the time to open a PR do so.

I could chip in here! Do you feel like that's enough of a breaking change on its own to bump to `0.3`?

sapessi:
> * The idiomatic crate providing that event type cannot be described using a A: for<'de> Deserialize.
> * The trait system features required to support this are unavailable on stable Rust (for instance, if this were a Haskell runtime, I'd rely on GeneralizedNewtypeDeriving or DerivingVia for the http event type).
> * There exists no reasonable default (de)serialize implementation for the event type in question.

This is partly why in the current version of the runtime we have a core handler that simply deals with `Vec<u8>` and allows customers to implement their own marshalling/unmarshalling logic. At the moment Lambda only supports JSON events, that doesn't mean that in the future it won't support other encodings as well as streaming. Not sure I want to walk down that one-way door.

> * The event type is a commonly used event type, exceeding—for instance—20% of all Lambda invocations.

I see this as a good reason for us to provide an official AWS crate to support the event type. That does not mean it needs to be bundled with the core runtime.

> For apigw events, does that mean we'd end up with something like http::Request<ApiGWRequest<UserDefinedBody>> vs todays http::Request<UserDefinedBody>> + RequestExt to expose ApiGw specific features like preparsed path and query parameters? The same for http::Response<ApiGWResponse<UserDefinedBody>>.

Agree. I'd really like to support `http::Request<UserDefinedBody>`, without going through the service-specific event type. 

>> We are able to approximate a similar API using the #[lambda] procedural macro, which requires some sort of coupling somewhere in this repository. 

The proc macro is a nice simplification for users but I would not make a change to the core APIs we are not happy with for the sake of maintaining the proc macro - as @softprops  said, we can easily have the separate http crate define its own macro.

> 
> > At the very least, the error API is can be backported to the existing 0.2 API, but I haven't had the time to open a PR do so.
> 
> I could chip in here! Do you feel like that's enough of a breaking change on its own to bump to `0.3`?

**Yes**, if you have some spare time it would be super-helpful @softprops.

davidbarsky:
@softprops:

> For apigw events, does that mean we'd end up with something like `http::Request<ApiGWRequest<UserDefinedBody>>` vs todays `http::Request<UserDefinedBody>> + RequestExt` to expose ApiGw specific features like pre-parsed path and query parameters? The same for `http::Response<ApiGWResponse<UserDefinedBody>>`.

> I feel like the latter is more ergonomic and possible with today's layered approach where the interface exposed wraps an internal delegation to the to core interface.

I agree! I would prefer today's approach, which would be `http::Request<UserDefinedBody>`. However, I forgot that an `A` in `http::Request<A>` might not implement `for<'de> Deserialize`. I think we might need to rely on the existing `Body` struct instead:

```rust
pub enum Body {
    /// An empty body
    Empty,
    /// A body containing string data
    Text(String),
    /// A body containing binary data
    Binary(Vec<u8>),
}
```

I'll also cover how I expect (de)serialization would work from API Gateway's JSON types. Given a JSON blob like this (from https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-create-api-as-simple-proxy-for-lambda.html):

```
{
  "message":"Good evening, John of Seattle. Happy Thursday!", 
  "input":{
    "resource":"/helloworld",
    "path":"/helloworld",
    "httpMethod":"POST",
    "headers":{"Accept":"*/*",
    "content-type":"application/json",
    "day":"Thursday",
    "Host":"r275xc9bmd.execute-api.us-east-1.amazonaws.com",
    "User-Agent":"curl/7.64.0",
    "X-Amzn-Trace-Id":"Root=1-1a2b3c4d-a1b2c3d4e5f6a1b2c3d4e5f6",
    "X-Forwarded-For":"72.21.198.64",
    "X-Forwarded-Port":"443",
    "X-Forwarded-Proto":"https"},
    "multiValueHeaders":{"Accept":["*/*"],
    "content-type":["application/json"],
    "day":["Thursday"],
    "Host":["r275xc9bmd.execute-api.us-east-1.amazonaws.com"],
    "User-Agent":["curl/0.0.0"],
    "X-Amzn-Trace-Id":["Root=1-1a2b3c4d-a1b2c3d4e5f6a1b2c3d4e5f6"],
    "X-Forwarded-For":["11.22.333.44"],
    "X-Forwarded-Port":["443"],
    "X-Forwarded-Proto":["https"]},
    "queryStringParameters":{"city":"Seattle",
    "name":"John"
  },
  "multiValueQueryStringParameters":{
    "city":["Seattle"],
    "name":["John"]
  },
  "pathParameters":null,
  "stageVariables":null,
  "requestContext":{
    "resourceId":"3htbry",
    "resourcePath":"/helloworld",
    "htt* Connection #0 to host r275xc9bmd.execute-api.us-east-1.amazonaws.com left intact pMethod":"POST",
    "extendedRequestId":"a1b2c3d4e5f6g7h=",
    "requestTime":"20/Mar/2019:20:38:30 +0000",
    "path":"/test/helloworld",
    "accountId":"123456789012",
    "protocol":"HTTP/1.1",
    "stage":"test",
    "domainPrefix":"r275xc9bmd",
    "requestTimeEpoch":1553114310423,
    "requestId":"test-invoke-request",
    "identity":{
      "cognitoIdentityPoolId":null,
      "accountId":null,
      "cognitoIdentityId":null,
      "caller":null,
      "sourceIp":"test-invoke-source-ip",
      "accessKey":null,
      "cognitoAuthenticationType":null,
      "cognitoAuthenticationProvider":null,
      "userArn":null,
      "userAgent":"curl/0.0.0","user":null
    },
    "domainName":"r275xc9bmd.execute-api.us-east-1.amazonaws.com",
    "apiId":"r275xc9bmd"
  },
  "body":"{ \"time\": \"evening\" }",
  "isBase64Encoded":false
  }
}
```

The `input` blob would be placed into `http::Request`'s [Parts](https://docs.rs/http/0.1.17/http/request/struct.Parts.html), with the json blob's `headers` object being placed into `Parts::headers`, the JSON method's `httpMethod` being placed into `Parts::method`, and so on. I'm papering over _some_ details, but I hope that the _gist_ of what I am attempting to communicate is clear.

I'll also note that as I was writing the above paragraph, I realized that we might not want to introduce a handler like:

```rust
pub trait Handler<A, B>
where
    A: for<'de> Deserialize<'de>,
    B: Serialize
{
/// redacted implementation
}
```

I think we might want something closer to:

```rust
pub trait Handler<A, B> {
/// redacted implementation
}
```

....and have _dedicated_ handlers that add `A: for<'de> Deserialize<'de>`/`B: Serialize` or `TryInto<http::Request<lambda_http::Body>>`/`TryFrom<http::Response<lambda_http::Body>>` bounds. This might allow us to break the `http`-specific handler into a dedicated crate and _optionally_ re-export behind a feature flag.

> If the api did look like `http::Request<ApiGWRequest<UserDefinedBody>>` would that make other non apigw events representable as `http::Request<CloudWatchEvent>` and is that desirable?

I don't think we should to make `http::Request<CloudWatchEvent>` representable. The only circumstance where I can think that can exist is if a customer creates a (questionable?) API Gateway/CloudWatch integration, but @sapessi should be able to correct me. If this _is_ possible, we should ask customers "goodness, _why_ would you do that?".

> Something worth considering is what if instead of a single proc macro that requires knowledge of dealing all possible event types to dispatch logic on, there could be specialized proc macros that dealt with specific cases #[xxx::lambda]. This may remove some of the type system limitations your bumping into as well as the requirements for the core crate to have detailed knowledge of specific event types. This comes at the cost of the boiler plate that comes along with implementing attributes which may be a heavy enough tax to scrap this idea.

I'm increasingly coming around to that idea. I don't think I ever spelled this out on GitHub, but I was thinking that the procedural macro would dispatch to a JSON-based or HTTP-based runner depending on whether `#[lambda(type = JSON)]` or `#[lambda(type = HTTP)]` is provided (note: please feel free to quibble with specific syntax. It's nowhere near final).

One thing I'll mention is that I'd like to avoid dedicated `lambda-attributes` and `lambda-http-attribute` crates to distinguish between JSON and HTTP-based events. For this reason, I'm in favor of relying on cargo's feature flags to support this functionality despite that cargo's feature flags being harder to test & support in CI and it being a somewhat unfamiliar concept to npm/Maven users.

> I could chip in here! Do you feel like that's enough of a breaking change on its own to bump to 0.3?

Please do so! I'm not 100% sure it requires a breaking change, but if it does, what's why we have semver!

---

I apologize for not making my intent more clear! I think I'm in agreement with you with what interface we ought expose, but we are disagreeing on _how_ we can expose that interface.

softprops:
Thanks for the extra clarifications @davidbarsky.  I feel much more aligned now!

I can take a stab at the error api backport as early as next weekend.

softprops:
starting to look into back the error api from this branch to 0.2 and noticed 

today we have

```rust
#[derive(Serialize)]
pub struct ErrorResponse {
    /// The error message generated by the application.
    #[serde(rename = "errorMessage")]
    pub error_message: String,
    /// The error type for Lambda. Normally, this value is populated using the
    /// `error_type()` method from the `LambdaErrorExt` trait.
    #[serde(rename = "errorType")]
    pub error_type: String,
    /// The stack trace for the exception as vector of strings. In the framework,
    /// this value is automatically populated using the `backtrace` crate.
    #[serde(rename = "stackTrace")]
    pub stack_trace: Option<Vec<String>>,
}
```

this branch has

```rust
#[derive(Serialize, Deserialize, Debug, PartialEq)]
pub struct ErrorReport {
    /// The type of the error passed to the Lambda APIs.
    pub name: String,
    /// The [std::fmt::Display] output of the error.
    pub err: String,
}
```

Two questions

1) according to [the runtime docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-invokeerror) there should be `errorType` and `errorMessage` fields. Should new `ErrorReport` api be using those field names?

2) I noticed `stackTrace` is absent from the new api, presumably because its not something that comes for free in std lib which is fine. I just wanted to bring it up in case I was missing anything.

davidbarsky:
> 1. According to the runtime docs there should be errorType and errorMessage fields. Should new ErrorReport api be using those field names?

Yep! That's an oversight on my part.

> 2.  I noticed stackTrace is absent from the new api, presumably because its not something that comes for free in std lib which is fine. I just wanted to bring it up in case I was missing anything.

We should probably support stack traces behind a (on-by-default) feature flag until https://github.com/rust-lang/rust/pull/60852 lands in the stable release channel, which ought be August 15th.

Folyd:
This PR is great! However, any plan to finish and merge this PR in this year? We nearly reach **2020** soon.

davidbarsky:
@lvicentesanchez: Yes, sorry for the delay. I think we should. I deeply apologize for not focusing enough on #106. We've been stretched somewhat thin over the past few months: all of us has things come up in our personal lives.

@Folyd: yep, I think we're near the home stretch now. This PR wasn't _particularly_ useful since Rusoto didn't support working with `std::future::Future`/Tokio 0.2, but since https://github.com/rusoto/rusoto/pull/1498 is nearing completion, this PR took on a bit of urgency. I think I'll close this PR and re-open it as a new one in order to make a rebase easier (and the fact that some assumptions and designs changed.)

As an aside, I've _nearly_ gotten https://github.com/tokio-rs/tracing working with AWS X-Ray, which means this Rust Runtime will work with `tracing` and AWS X-Ray, out of the box. The work Eliza and I did on https://github.com/tokio-rs/tracing/pull/420 was a big enabler in that.

davidbarsky:
@skorobkov Yep, good catch. The path _does_ need to be be that.

lambda/Cargo.toml
davidbarsky: we're off the alphas now.

softprops:
What's the context around the glibc note and running rust about?

lambda/src/types.rs
```diff
@@ -0,0 +1,145 @@
...
+    assert_eq!(expected, actual);
+
+    Ok(())
+}
+
+/// The request ID, which identifies the request that triggered the function invocation. This header
+/// tracks the invocation within the Lambda control plane. The request ID is used to specify completion
+/// of a given invocation.
+#[derive(Debug, Clone, PartialEq)]
+pub struct RequestId(pub String);
```
softprops: Do these wrapper types ever get used?

softprops:
I love how far this has come, seeing how much was simplified, and how much you were able to remove. I'm wondering it would do the project well to just merge this ship an alpha and iterate.

lambda/src/types.rs
davidbarsky: I don't think so. They can be removed, I think.

davidbarsky:
@softprops 
> What's the context around the glibc note and running rust about?

A lot of issues crop when applications using this project _don't_ target `x86_64-unknown-linux-musl`. I'm not sure I've ever seen this project successfully be used by applications when they target `x86_64-unknown-linux-gnu`.

> I love how far this has come, seeing how much was simplified, and how much you were able to remove.

Thank you so much! 

> I'm wondering it would do the project well to just merge this ship an alpha and iterate.

Maybe? Of the open items, I think the stream conversion, the notice about targeting _only_ `x86_64-unknown-linux-musl`, and examples using `tokio-compat` are the critical blockers to an alpha release. I'd like to get the stream conversion in this PR because it'd reduce churn in subsequent releases and the latter two are the primary impediments to using this runtime in production.

The other open items can be resolved in subsequent releases, IMO.

softprops:
Good stuff I can't wait!

.rustfmt.toml
```diff
@@ -1,3 +1,4 @@
+edition = "2018"
```
softprops: Didn't know this was a thing but heads up, [rustfmt picks up the edition from your crate if invoked via cargo](https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#edition) so you might not need this

davidbarsky: Yep, it normally should, but I think rust-analyzer didn't pick it up, but I think that was fixed.

bryantbiggs:
this is amazing @davidbarsky! it has been really interesting to follow along with this PR and I can't wait for the final result, great job!

davidbarsky:
@bryantbiggs thanks!

lambda-attributes/src/lib.rs
```diff
@@ -0,0 +1,68 @@
...
+                #asyncness fn main() {
+                    async fn actual(#arg_name: #arg_type) #ret {
+                        #body
+                    }
+                    let f = lambda::handler_fn(actual);
+                    lambda::run(f).await.unwrap();
+                }
+            }
+        }
+        _ => {
```
sapessi: I can see someone who just wants to schedule a task through CloudWatch and is not interested in any input. We can simply drop the event and ctx and start their code.

lambda/src/client.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    fn call(&mut self, req: Request<Body>) -> Self::Future {
+        let fut = async move {
+            let path = req.uri().path_and_query().unwrap().as_str();
+            let rsp = if path.ends_with("next") {
+                next_event(req).await
+            } else if path.ends_with("response") {
+                complete_event(req).await
+            } else {
```
sapessi: We should support the [init fail method](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html#runtimes-api-initerror) too. Errors in the "construction" of a handler method should be sent to the init error endpoint.

lambda/src/client.rs
```diff
@@ -0,0 +1,287 @@
...
+    assert_eq!(req.uri().path_and_query().unwrap(), &PathAndQuery::from_static(path));
+    let body = json!({"message": "hello"});
+
+    let rsp = NextEventResponse {
+        request_id: "8476a536-e9f4-11e8-9739-2dfe598c3fcd",
+        deadline: 1542409706888,
+        arn: "arn:aws:lambda:us-east-2:123456789012:function:custom-runtime",
+        trace_id: "Root=1-5bef4de7-ad49b0e87f6ef6c87fc2e700;Parent=9a9197af755a6419",
+        body: serde_json::to_vec(&body)?,
+    };
```
sapessi: uh?

lambda/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+where
+    F: Fn(A) -> Fut,
+    Fut: Future<Output = Result<B, Err>> + Send,
+    Err: Into<Box<dyn std::error::Error + Send + Sync + 'static>> + fmt::Debug,
+{
+    type Err = Err;
+    type Fut = Fut;
+    fn call(&mut self, req: A) -> Self::Fut {
+        // we pass along the context here
+        (self.f)(req)
```
sapessi: The doc at the top of the file says that I can optionally receive a ctx. I'm probably missing the function but where is the call method that passes the ctx?

lambda/src/requests.rs
```diff
@@ -0,0 +1,156 @@
...
+impl IntoRequest for InitErrorRequest {
+    fn into_req(self) -> Result<Request<Body>, Err> {
+        let uri = format!("/2018-06-01/runtime/init/error");
+        let uri = Uri::from_str(&uri)?;
+
+        let req = Request::builder()
+            .method(Method::POST)
+            .uri(uri)
+            .header("lambda-runtime-function-error-type", "unhandled")
+            .body(Body::empty())?;
```
sapessi: Should this also be: `serde_json::to_vec(&self.diagnostic)?;`

davidbarsky: Yes, thank you for the catch.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os:
+          - ubuntu-latest
+          - macOS-latest
+        rust:
+          - 1.40.0
```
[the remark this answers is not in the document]
davidbarsky: It depends: do we intend to have an minimum supported rust version?

lambda-attributes/src/lib.rs
davidbarsky: Solid point, I'll add that.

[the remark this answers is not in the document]
davidbarsky: Oh, as part of the macro invocation? Good point, I think that can be a decent option.

> ...but it feels less idiomatic than receiving the content as part of the call since the context is technically related to the invocation.

On a somewhat related note: to sum up my thoughts on this matter—now that I've had a few months to think about it—I've seen three types of APIs in Rust for "context":

1. The thread local approach. Both [`std`](https://doc.rust-lang.org/std/thread/struct.LocalKey.html), [`async-std`](https://docs.rs/async-std/1.4.0/async_std/macro.task_local.html), and in the coming future, Tokio, take this approach. For ambient context that you need _some_ of the time, this is the approach that seems to be preferred. 
2. The "put it in an extensions typemap". This is what Tide and `http` do, but it does complicate trait bounds a bit and it requires _every_ invocation to have an associated typemap. I don't think that's the correct approach here, especially since we're not in control of generating typed bindings for events. Plus, this would be too much churn IMO.
3. Pass the context on _every_ invocation. This is what we do in [`tracing-subscriber`'s Layer](https://tracing.rs/tracing_subscriber/layer/trait.layer#provided-methods) (an interface for building your own `tracing` exporter, among other things), but the context there is different: you need it for nearly _every_ method to determine your current position within an execution graph.

My mental model of Lambda makes me think this approach fits best, but we can simulate how _other_ languages supported by Lambda do it through macros.

lambda/src/client.rs
davidbarsky: True, yes—I dropped this for the initial pass. Thanks for the reminder.

lambda/src/client.rs
davidbarsky: This lets us test the handler + client in memory. I'd like to replace these hard-coded values with something randomly generated, perhaps something in the style of: https://github.com/hyperium/http/blob/master/tests/header_map_fuzz.rs.

lambda/src/lib.rs
davidbarsky: There isn't one—it's an ambient "global", local only to the Future. The context is set in the `Executor::run` method.

.github/workflows/build.yml
```diff
@@ -0,0 +1,71 @@
...
+        run: cargo build --all --verbose
+        env:
+          TARGET: ${{ matrix.target }}
+        continue-on-error: ${{ matrix.allow_failure }}
+      - name: Run tests
+        run: cargo test --all --verbose
+        env:
+          TARGET: ${{ matrix.target }}
+        continue-on-error: ${{ matrix.allow_failure }}
+  fmt:
```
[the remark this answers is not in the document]
davidbarsky: Yeah, I think we should add Clippy.

lambda/src/client.rs
sapessi: Thought so, thanks. Perhaps we should change the function name to make it clear that it's used for testing, or move it to the tests module.

lambda/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+//!    decorated using `#[lamdba]`
+//! 2. A type that conforms to the [`Handler`] trait. This type can then be passed
+//!    to the the `lambda::run` function, which launches and runs the Lambda runtime.
+//!
+//! An asynchronous function annotated with the `#[lambda]` attribute must
+//! accept an argument of type `A` which implements [`serde::Deserialize`] and
+//! return a `Result<B, E>`, where `B` implements [`serde::Serializable`]. `E` is
+//! any type that implements `Into<Box<dyn std::error::Error + Send + Sync + 'static>>`.
+//!
+//! Optionally, the `#[lambda]` annotated function can accept an argument
```
sapessi: Should we remove this then?

lambda-attributes/src/lib.rs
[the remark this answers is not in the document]
sapessi: The thread local approach makes sense for all of the function configuration that is stored in the environment. Would you use the same approach for values that are specific to an invocation such as the timeout - I mean the values passed by the runtime APIs in the response headers?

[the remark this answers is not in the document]
bryantbiggs: just wanted to offer an outside perspective - I've worked extensively with lambda in a number of languages over the past couple of years and they all offer (at minimum) the standard function signature of `fn handler(event, context):`. When I first saw this in rust where context was exported to a macro, it seemed a bit odd (IMO). Are there any benefits to diverging from this norm of packing the event context into the function signature vs a macro?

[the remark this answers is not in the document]
davidbarsky: > The thread local approach makes sense for all of the function configuration that is stored in the environment. Would you use the same approach for values that are specific to an invocation such as the timeout - I mean the values passed by the runtime APIs in the response headers?

Yep. You only care about that information _some_ of the time, so the timeout would be placed into the `lambda::context()`. Its usage isn't driven on a _per instantiation_-basis, but on a _as-needed_ basis.

> Are there any benefits to diverging from this norm of packing the event context into the function signature vs a macro?

I think the big disagreement between @sapessi and I that we haven't decided _which_ grain we should go against in designing this API—Rust's or other Lambdas. I believe we should opt for aligning with _Rust's_ grain, while Stefano has argued (and still might!) for aligning with _Lambda's_ grain. I believe this for a few reasons:

1. Rust is sufficiently different from those other languages that opting for a (slightly, in fairness) unidiomatic Rust API is a reduction in customer experience for those coming to this project from other Rust projects. I think this difference is _most_ acutely noticed in Rust since Rust does not support optional arguments like _all other_ officially supported languages on Lambda (see https://github.com/rust-lang/rfcs/issues/323 for details).
2. This reduces the (already small) effort needed to bridge to other ecosystems like [`tower::Service`](https://docs.rs/tower/0.3.0/tower/trait.Service.html) or [`http_service::HttpService`](https://docs.rs/http-service/0.4.0/http_service/trait.HttpService.html). The pain might be most noticeable when working with generics, but this will likely be small.
3. For those coming from other languages to this project, I expect they _won't_ be implementing a `Handler` by hand—they'll be using the `#[lambda]` macro, which can simulate variadic arguments like other languages. If someone is implementing a `Handler` by hand, I expect that the individual would already be an intermediate or advanced Rust user who already has some expectations about grabbing context from a thread local.

lambda/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+    let mut handler = handler;
+    let config = Config::from_env()?;
+    let client = Client::with(&config.endpoint, hyper::Client::new())?;
+    let mut exec = Executor { client };
+    exec.run(&mut handler, false).await?;
+
+    Ok(())
+}
+
+/// Runs the lambda function almost entirely in-memory. This is meant for testing.
```
softprops: To keep a smaller surface area, if this is intended  for tests, thoughts on scoping it to `#[cfg(test)]`?

lambda/src/lib.rs
```diff
@@ -0,0 +1,269 @@
...
+        Ok(conf)
+    }
+}
+
+thread_local! {
+    static TASK_LOCAL: RefCell<Option<types::LambdaCtx>> = RefCell::new(None);
+}
+
+#[pin_project::pin_project]
+struct WithTaskLocal<F> {
```
softprops: Naming suggestion, take it it leave it. `WithLambdaCtx` is really what this is isn't it?

softprops: Naming suggestion, take it or leave it. `WithLambdaCtx` describes what this is a little more clearly.

vultix:
[This](https://github.com/rusoto/rusoto/pull/1664) PR just landed in Rusoto.  I can't wait for Rust to support async AWS on all fronts, so thank you everyone for all of your work!

ayax79:
Looking forward to seeing this merged!

myfreeweb:
Thanks everyone, I just async-ified [my project](https://github.com/myfreeweb/imgroll) using this. A couple notes:

> `#[lambda] #[tokio::main]`

This seems to break error locations?? As in `rustc` doesn't show where in that annotated function errors are if they exist. (I'm on `rustc 1.42.0-nightly (c2d141df5 2020-01-24)`) Only using `#[tokio::main]` with `lambda::run` is fine.

> The glibc in the Lambda execution environment is too old to run Rust executables

Maybe for the `rustup` official distributions of Rust, but [building with the Rust 1.39.0 that's in CentOS7/EPEL](https://github.com/myfreeweb/imgroll/commit/17ed79e7580a1b16cab8141411b27eac366aa8e1) (and using shared libs from there) still works fine :)

iliana:
> This seems to break error locations?? As in `rustc` doesn't show where in that annotated function errors are if they exist. (I'm on `rustc 1.42.0-nightly (c2d141df5 2020-01-24)`) Only using `#[tokio::main]` with `lambda::run` is fine.

I ran into this issue yesterday as well.

davidbarsky:
> This seems to break error locations?? As in rustc doesn't show where in that annotated function errors are if they exist. (I'm on rustc 1.42.0-nightly (c2d141df5 2020-01-24)) Only using #[tokio::main] with lambda::run is fine.

I _think_—but I am not sure—that this is a limitation of proc macro hygiene in Rust at the moment. I think we might need to drop the `tokio::main` to recover better error messages.

davidbarsky:
This PR has been open for a while now and it's time to rip off the bandage—I'm going to merge this PR as-is and start cutting beta releases. Customers have been using this branch without issue. We'll be making a few breaking changes to those betas, but they'll be _relatively_ minor and well-documented in the changelog.

PR #112, opened 2019-06-30
Add missing backtick in doc string

nicholasbishop:


davidbarsky:
D'oh, I think I may fixed this on another branch! Thanks for the fix, though.

PR #188, opened 2019-12-25
Use default HTTP scheme for endpoint like "127.0.0.1:9001"

skorobkov:
AWS_LAMBDA_RUNTIME_API environment is equal to "127.0.0.1:9001".
http::Uri can't "guess" scheme from that string, so it's None and function Client::set_origin() always fails.


davidbarsky:
thank you!

PR #214, opened 2020-03-22
internal: internal cleanup; correctness

davidbarsky:
*Issue #, if available:* N/A

*Description of changes:*

This branch has a few changes:
1. Since [AWS X-Ray is now availible in GovCloud](https://aws.amazon.com/about-aws/whats-new/2020/02/aws-x-ray-now-available-in-aws-govcloud-us-regions/), we can make the trace id in the context non-optional.
2. The (internal) runtime client is now takes a `&self` instead of a `&mut self`, reducing how many `.clone()`s we need need to write. This also means that the client no longer implements `tower::Service`, but given that functionality wasn't used _anyways_, I don't consider this to much of a loss.
3. We're now modeling the incoming events as a stream using the (awesome!) [genawaiter](https://docs.rs/genawaiter/0.99.1/genawaiter/) crate. This can be replaced with a manual stream implementation + [pin_project_lite](https://docs.rs/pin-project-lite/0.1.4/pin_project_lite/) in the future. Side note: the Lambda Runtime server-side APIs (for those unfamiliar, the Firecracker boxes) suspend the function once they receive the `next_event` call, _not_ the response handler. This allows for races to be cleanly-side stepped.

Still upcoming:

1. `tracing` support for logs, metrics, and distributed tracing (e.g., AWS X-Ray Integration).
2. Docs!
3. Refactoring the test support to not rely sending requests over localhost as per https://github.com/davidbarsky/stuff. Props to Dylan Marriner on the Lambda team and @gardnervickers for the tip and approach.


lambda/src/lib.rs
```diff
@@ -34,32 +34,27 @@
 //! }
 //! ```
 pub use crate::types::LambdaCtx;
+use anyhow::Error;
```
softprops, line 37: I'm a fan of `anyhow` but do you think exposing this is the outward facing api is going to last? 

I asking because I feel like this is a trap the use error crates in libraries in the rust ecosystem in the past. On year is `error_chain`, the next year it's `failure`, the next year it's `anyhow`. I think these are all really good for use in applications but for library crates it's problematic when the community shifts its opinions forcing  applications to deal with breaking library changes when that happens.

lambda/src/lib.rs
```diff
@@ -152,104 +147,85 @@ where
...
-    F: Future,
-{
-    type Item = Result<Response<Body>, S::Error>;
-
-    fn poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
-        unimplemented!()
-    }
+fn incoming(client: &Client) -> impl Stream<Item = Result<http::Response<hyper::Body>, Error>> + '_ {
+    gen!({
+        let req = NextEventRequest.into_req().expect("Unable to construct request");
```
softprops, line 186: well that was neat!

lambda/src/lib.rs
davidbarsky: Yeah, it shouldn't be exposed, I'm just very lazy! _At the very most_, we should depend on [`thiserror`](https://docs.rs/thiserror/1.0.15/thiserror), which was designed to be safely used by libraries by _just_ expanding to `std::error::Error` with _no_ trace of `thiserror`. A future removal of `thiserror` will be equivalent to hand-written `std::error::Error` implementation.

---

Now that I think of it, I think that I can write _all_ of the errors using `thiserror`, but only publish the slightly cleaned up `cargo expand` output.

lambda/src/lib.rs
davidbarsky: I agree!

lambda/src/lib.rs
sapessi: +1 we've been down this error path before. We should try and expose only std errors without forcing an opinion/dependency on function developers

davidbarsky: Yeah, I'll drop it in the next PR and use `std::error::Error` without any derives!. I primarily want to merge this _as is_ so that some _other_ folks can build on these changes more easily.

PR #217, opened 2020-04-21
adapt lambda-http to new lambda runtime api

softprops:
*Issue #, if available:*

*Description of changes:*

Kicking the tires on reviving lambda-http crate. This is not yet finalized but I wanted to communicate progress is being made. Basic things work with the new much more simplified lambda runtime api. Still trying to figure out what good looks like.

- [x] make compile
- [x] tests pass
- [x] update examples
- [x] update dependencies
- [x] adapt to new apigw http api events 
- [x] define a proc macro for http mains


lambda-http/src/lib.rs
```diff
@@ -33,99 +34,110 @@
...
 /// Type alias for `http::Request`s with a fixed `lambda_http::Body` body
 pub type Request = http::Request<Body>;
 
-/// Functions serving as ALB and API Gateway handlers must conform to this type.
-pub trait Handler<R> {
-    /// Run the handler.
-    fn run(&mut self, event: Request, ctx: Context) -> Result<R, HandlerError>;
+/// Functions serving as ALB and API Gateway REST and HTTP API handlers must conform to this type.
+///
+/// This can be viewed as a `lambda::Handler` constained to `http` crate `Request` and `Response` types
```
softprops, line 68: Looking for input on what constitutes agreeable here. The general idea is that I'd like to `lambda_http::Handler` that mentally maps to `lambda::Handler` but whose input and output's are `http` crate type based. Since those types are not (de)serializable. We can't use `lambda::Handler` directly. I also like the idea of have an interface that pins output types to input types, something that would come in handy for triggers with contracts that define a well defined input/output shapes.  

I'd like to keep the same level of flexibility has `lambda::Handler` in that you can define your own or "just use a closure".

lambda-http/src/lib.rs
```diff
@@ -33,99 +34,110 @@
...
+    type Err;
+    /// The type of Response this Handler will return
+    type Response: IntoResponse;
+    /// The type of Future this Handler will return
+    type Fut: Future<Output = Result<Self::Response, Self::Err>> + 'static;
+    /// Function used to execute handler behavior
+    fn call(&mut self, event: Request) -> Self::Fut;
+
+    /// Consumes this Handler into an Adapter type which implements `lambda::Hander`
+    fn to_adapter(self) -> Adapter<Self> {
```
softprops, line 80: This was a bit annoying but I found I couldn't just implement `lambda::Handler` for any `lambda_http::Handler` directly. I needed to add a wrapping locally defined type. http://smallcultfollowing.com/babysteps/blog/2015/01/14/little-orphan-impls/ explains why. In particular the bits about "covered" types. The local `Adapter` type is "covering" handler impls. Open to suggestions on alternatives

lambda-http/src/response.rs
```diff
@@ -15,7 +15,7 @@ use crate::body::Body;
 /// Representation of API Gateway response
 #[derive(Serialize, Debug)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct LambdaResponse {
+pub struct LambdaResponse {
```
softprops, line 18: because `handler` exports a wrapper type, these now need to be pub

lambda-http/src/lib.rs
davidbarsky: > Looking for input on what constitutes agreeable here.

I think what you proposed is a the most viable approach, but I need to give this a more thorough review. The only alternative I can think of that _might_ work doing the duck-typed dispatch that the `write!` macro does under the hood: https://doc.rust-lang.org/std/macro.write.html

> I also like the idea of have an interface that pins output types to input types, something that would come in handy for triggers with contracts that define a well defined input/output shapes.

That's interesting, but I'm not really picturing _how_ this might work. It's not urgent, but I'd love to hear more details.

lambda-http/src/lib.rs
davidbarsky: Thanks for the note. Yeah, this sits at a _bad_ intersection of trait coherence, specialization, and what rustc would might allow. It's a pickle!

lambda-http/src/lib.rs
```diff
@@ -33,99 +34,110 @@
...
-/// * `f` A type that conforms to the `Handler` interface.
-///
-/// # Panics
-/// The function panics if the Lambda environment variables are not set.
-pub fn start<R>(f: impl Handler<R>, runtime: Option<TokioRuntime>)
+#[doc(hidden)]
+pub struct TransformResponse<R, E> {
+    is_alb: bool,
+    fut: Pin<Box<dyn Future<Output = Result<R, E>>>>,
+}
```
davidbarsky, line 110: Two thoughts:

1. Can this be converted a standalone async/await block rather than a manual future implementation?
2. If not, use `pin_project` + add an additional generic argument + a PhantomData for the future?

lambda-http/src/ext.rs
```diff
@@ -1,6 +1,5 @@
 //! ALB and API Gateway extension methods for `http::Request` types
 
-use failure::Fail;
```
davidbarsky, line 3: Yay!

lambda-http/src/lib.rs
```diff
@@ -1,10 +1,10 @@
 #![warn(missing_docs)]
 //#![deny(warnings)]
-//! Enriches the `lambda_runtime` crate with [http](https://github.com/hyperium/http)
-//! types targeting ALB and API Gateway proxy events.
+//! Enriches the `lambda` crate with [http](https://github.com/hyperium/http)
+//! types targeting ALB, API Gateway REST and HTTTP API proxy events.
```
davidbarsky, line 4:
```suggestion
//! types targeting ALB, API Gateway REST and HTTP API proxy events.
```

Note that I don't _think_ your changes handle the schema of the new HTTP APIs yet.

lambda-http/src/ext.rs
```diff
@@ -23,13 +22,13 @@ pub(crate) struct PathParameters(pub(crate) StrMap);
...
 pub enum PayloadError {
     /// Returned when `application/json` bodies fail to deserialize a payload
-    #[fail(display = "failed to parse payload from application/json")]
+    //#[fail(display = "failed to parse payload from application/json")]
     Json(serde_json::Error),
     /// Returned when `application/x-www-form-urlencoded` bodies fail to deserialize a payload
-    #[fail(display = "failed to parse payload application/x-www-form-urlencoded")]
+    //#[fail(display = "failed to parse payload application/x-www-form-urlencoded")]
     WwwFormUrlEncoded(SerdeError),
 }
```
davidbarsky, line 33: Few notes:

- Does this implement `std::error::Error`? If not, do you want to implement it by hand or do you want to `thiserror`, whose removal would be a non-breaking change?
- Mark this enum as `#[non_exhaustive]`?
- Add a `PartialEq` implementation?

lambda-http/Cargo.toml
```diff
@@ -16,18 +16,16 @@ travis-ci = { repository = "awslabs/aws-lambda-rust-runtime" }
...
 serde_derive = "^1"
-lambda = { path = "../lambda" }
-tokio = "^0.1"
-base64 = "0.10"
-failure = "0.1"
-failure_derive = "0.1"
-serde_urlencoded = "0.5"
+serde_json = "^1"
+serde_urlencoded = "0.6"
+tokio = "^0.2"
```
davidbarsky, line 26: I haven't though through this yet, _but_—what parts of Tokio _need_ to be pulled into `lambda_http` as a project dependency?

davidbarsky:
Thanks so much for doing this. I posted a review/questions kicking the tires on _your_ tire kicking :)

lambda-http/src/lib.rs
softprops: I pushed a change that makes this slightly more transparent.

lambda-http/src/lib.rs
softprops: getting there...

lambda-http/src/ext.rs
softprops: Currently re-evaluating what I actually need. Left this commented out as a highlighted reminder for a todo.

lambda-http/Cargo.toml
softprops: This list just got resorted. I don't think there are any actually. I'll double check and remove if so. I only needed a dev-dependency for `#[tokio::main]` bits but I don't think we're actually referencing tokio types directly yet which shines a light on some nice design choices for the new api. Thanks for that.

softprops:
> I posted a review/questions kicking the tires on your tire kicking

Thanks. I'll keep this in draft until I'm ready so expect more change. I'm just using the draft to communicate progress being made. It's still baking in bread in the oven.

davidbarsky:
Sounds good, I'll hold off on making additional comments until this is moved out of a draft state.

lambda-http/src/lib.rs
softprops: 1. I'd would have liked to. Since this is in `Handler::call` but ```rustc --explain E0706
`async fn`s are not yet supported in traits in Rust.
...
```  explains this a bit better than I could.

2. Ill try the `pin_project` thing. 

This is essentially just a `futures::future::Map` transformer. I was aiming to avoid adding the extra dep but what pushed me over the edge I got caught embarrassingly not being able to type out associated type for `lambda::Handler::Fut`. 

Making this a concrete type allowed me to just write out`type Fut = TransformResponse<H::Response, Self::Err>;`

File: lambda-http/src/request.rs [hunk 9]
```diff
@@ -19,90 +19,148 @@ use crate::{
...
-}
-
-/// Event request context as an enumeration of request contexts
-/// for both ALB and API Gateway http events
-#[derive(Deserialize, Debug, Clone)]
+#[derive(Deserialize, Debug)]
 #[serde(untagged)]
-pub enum RequestContext {
-    /// Api Gateway request context
+pub enum LambdaRequest<'a> {
```
softprops, line 30: Changing the strategy here. previously there was one struct made to look look like both apigw rest api proxy and alb events. With the addition of http api events that approach was starting to look long in the tooth and untenable. Going back to the drawing board on this answering "what is this even?" question. I came up with roughly

```typescript
type LambdaRequest = ApiGatewayRequest | ApiGatewayV2Request | AlbRequest;
```

Unionizing types rather than masquerading types had some tradeoffs and some useful benefits. Updating/evolving one variant doesn't require additional consideration in how another gets deserialized. Transforming 3 similar variants incurs some boiler plate but I could add some helpers to make that less onerous.  In retrospect this feels like a better approach given the tradeoffs.

lambda-http/src/request.rs
```diff
@@ -19,90 +19,148 @@ use crate::{
...
-        identity: Identity,
+        is_base64_encoded: bool,
+        request_context: ApiGatewayV2RequestContext,
     },
-    /// ALB request context
     #[serde(rename_all = "camelCase")]
-    Alb { elb: Elb },
-}
-
-impl Default for RequestContext {
```
softprops, line 80: Will come back to this.

lambda-http/src/request.rs
```diff
@@ -241,83 +311,210 @@ where
...
+                stage_variables,
+                body,
+                is_base64_encoded,
+                request_context,
+            } => {
+                let builder = HttpRequest::builder()
+                    .method(request_context.http.method.as_ref())
+                    .uri({
+                        format!(
+                            "{}://{}{}",
```
softprops, line 336: note to self: do some testing to see what `raw_query_string` apigw@v2 values look like with and without query params then add then to the [request uri](https://docs.rs/http/0.2.1/http/uri/struct.Uri.html). Query string params are part of that type despite being accessible via request extension for query_string_parameters

lambda-http/src/request.rs
```diff
@@ -111,6 +169,18 @@ pub struct Elb {
...
+/// Http information captured API Gateway v2 request context
+#[derive(Deserialize, Debug, Default, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct Http {
+    #[serde(deserialize_with = "deserialize_method")]
+    pub method: Method,
+    pub path: String,
+    pub protocol: String,
+    pub source_ip: String,
+    pub user_agent: String,
```
softprops, line 181: Not all clients send a user agent header. Need to confirm how apigw represents this in those cases

lambda-http/src/lib.rs
davidbarsky: > `async fn`s are not yet supported in traits in Rust.

Ah, gotcha. Don't treat this as a blocker. Boxing and pinning is fine for now.

> This is essentially just a futures::future::Map transformer. I was aiming to avoid adding the extra dep but what pushed me over the edge I got caught embarrassingly not being able to type out associated type for lambda::Handler::Fut.

It's a bit of a nightmare to write out some of the accociated types. I've gotten uncomfortably good at writing them and its a skill I wish I didn't have.

I was thinking that this contaminator could be replaced _wholesale_ with an async fn, but that's a refactoring that we can consider down the line.

lambda-http/src/lib.rs
davidbarsky: Oops! My bad.

lambda-http/src/ext.rs
davidbarsky: Gotcha, thanks!

lambda-http/Cargo.toml
davidbarsky: >  I only needed a dev-dependency for #[tokio::main] bits but I don't think we're actually referencing tokio types directly yet which shines a light on some nice design choices for the new api. Thanks for that.

Glad you like it :)

lambda-http/src/ext.rs
softprops: I went impl'ing `std::error::Error` since at this point that only really means impl `fmt::Display` . 

At this point this enum is exhaustive. It's an error that represents not being about to deserialize a type given a  `application/json` or `application/x-www-form-urlencoded` request body

Looks like I don't get an easy `derive(PartialEq)` since `serde_json::error::Error` doesn't impl it. Let me know if you really want that. 

Let me know if you want to see other changes here

davidbarsky: > I went impl'ing `std::error::Error` since at this point that only really means impl `fmt::Display`.

Sounds good!

> At this point this enum is exhaustive. It's an error that represents not being about to deserialize a type given a application/json or application/x-www-form-urlencoded request body

My concern is that a non-`#[non_exhaustive]` enum would become a forwards compatibility hazard if API Gateway (or some other networking product) added support for non-HTTP 1.1 protocols like HTTP/2 or HTTP/3. By saying this error _is_ exhaustive, we'd need to cut a breaking change to support a new protocol. It's not the end of the world, but if it means that customers are _slightly_ better prepared for breaking changes, I'll weakly press for `#[non_exhaustive]`.

> Looks like I don't get an easy derive(PartialEq) since serde_json::error::Error doesn't impl it. Let me know if you really want that.

Oh, bummer. Yeah, not worth it then.

lambda-http/src/ext.rs
```diff
@@ -205,7 +219,10 @@ impl RequestExt for HttpRequest<super::Body> {
     }
 
     fn request_context(&self) -> RequestContext {
-        self.extensions().get::<RequestContext>().cloned().unwrap_or_default()
+        self.extensions()
+            .get::<RequestContext>()
+            .cloned()
+            .expect("Request did not contain a request context")
```
softprops, line 225: I opted to remove a Default impl here for two reasons

1) aws not providing a request context should be an error
2) a context-specific defaulting can not clearly be derived from a context free fn

softprops:
Ok @davidbarsky. I'm pretty happy with this now. I also tested a few deployments to check for regressions. Let me know what you think.

lambda-http-attributes/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "lambda-http-attributes"
```
davidbarsky, line 2: Do you think this functionality could be a part of `lambda-attributes`? In `tracing`, the `#[instrument]` macro can instrument std::future::Future/async functions if another crate (`tracing_futures`) is in the dependency graph. See here: https://docs.rs/tracing/0.1.13/tracing/#spans-1. 

I'd like it if we can do something similar here to reduce how many crates customers might need and how many crates we need to release.

lambda-http/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "lambda_http"
 version = "0.1.1"
 authors = ["Doug Tangren"]
 edition = "2018"
-description = "Rust API Gateway proxy event interfaces for AWS Lambda"
+description = "Rust ALB and API Gateway proxy event interfaces for AWS Lambda"
```
davidbarsky, line 6:
```suggestion
description = "Application Load Balancer and API Gateway event types for AWS Lambda"
```

lambda-http/Cargo.toml
```diff
@@ -3,7 +3,7 @@ name = "lambda_http"
 version = "0.1.1"
```
davidbarsky, line 3: Bump to `0.2.0-beta.1`?

lambda-http/examples/basic.rs
```diff
@@ -1,26 +1,19 @@
...
 
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    lambda::run(handler(func)).await?;
     Ok(())
 }
 
-fn my_handler(e: Request, c: Context) -> Result<impl IntoResponse, HandlerError> {
-    Ok(match e.query_string_parameters().get("first_name") {
+async fn func(event: Request) -> Result<impl IntoResponse, Error> {
```
davidbarsky, line 11: Question—why an `impl IntoResponse` rather than http::Response<Body>`?

(More for customers than for me.)

lambda-http/src/body.rs
```diff
@@ -68,6 +68,22 @@ pub enum Body {
     Binary(Vec<u8>),
 }
 
+impl Body {
+    /// Decodes body, if needed, based aws provided contract
```
davidbarsky, line 72:
```suggestion
    /// Decodes body, if needed.
```

lambda-http/src/ext.rs
```diff
@@ -22,36 +20,51 @@ pub(crate) struct PathParameters(pub(crate) StrMap);
...
+            PayloadError::Json(json) => writeln!(f, "failed to parse payload from application/json {}", json),
+            PayloadError::WwwFormUrlEncoded(form) => writeln!(
+                f,
+                "failed to parse payload from application/x-www-form-urlencoded {}",
+                form
+            ),
+        }
+    }
+}
+impl Error for PayloadError {}
```
davidbarsky, line 46: Implement [`Error::source`](https://doc.rust-lang.org/std/error/trait.Error.html#method.source)?

lambda-http/src/ext.rs
```diff
@@ -22,36 +20,51 @@ pub(crate) struct PathParameters(pub(crate) StrMap);
...
 /// features.
 ///
 /// # Examples
 ///
-/// You can also access a request's body in deserialized format
-/// for payloads sent in `application/x-www-form-urlencoded` or
-/// `application/json` format.
+/// You can access a request's body in deserialized format
+/// for http request body payloads sent in `application/x-www-form-urlencoded` or
+/// `application/json` formats.
```
davidbarsky, line 57:
```suggestion
/// A request's body can be deserialized if its correctly encoded as per  
/// the request's `Content-Type` header. The two supported content types are
/// `application/x-www-form-urlencoded` and `application/json`.
```

lambda-http/src/ext.rs
davidbarsky: Makes sense!

lambda-http/src/ext.rs
```diff
@@ -263,91 +279,57 @@ mod tests {
...
-            headers,
-            body: Some(r#"{"foo":"bar", "baz": 2}"#.into()),
-            ..LambdaRequest::default()
-        };
-        let actual = HttpRequest::from(lambda_request);
-        let payload: Option<Payload> = actual.payload().unwrap_or_default();
+        let request = HttpRequest::builder()
+            .header("Content-Type", "application/json")
+            .body(Body::from(r#"{"foo":"bar", "baz": 2}"#))?;
+        let payload: Option<Payload> = request.payload().unwrap_or_default();
```
davidbarsky, line 312: Nit: can `HttpRequest` be called as `http::Request`? I'd rather prefer the full module path for all `hyperium/http` crates.

lambda-http/src/ext.rs
```diff
@@ -263,91 +279,57 @@ mod tests {
...
+    }
+
+    #[test]
+    fn requests_omiting_content_types_do_not_support_parseable_payloads() -> Result<(), Box<dyn Error>> {
+        #[derive(Deserialize, PartialEq, Debug)]
+        struct Payload {
+            foo: String,
+            baz: usize,
+        }
+        let request = HttpRequest::builder().body(Body::from(r#"{"foo":"bar", "baz": 2}"#))?;
```
davidbarsky, line 330: Same nit.

lambda-http/src/ext.rs
```diff
@@ -263,91 +279,57 @@ mod tests {
...
                 foo: "bar".into(),
                 baz: 2
             })
-        )
+        );
+        Ok(())
+    }
+
+    #[test]
+    fn requests_omiting_content_types_do_not_support_parseable_payloads() -> Result<(), Box<dyn Error>> {
```
davidbarsky, line 324: Let's not return errors in tests; can we `.expect()` fallible operations instead? The error messages on failure are better.

lambda-http/src/lib.rs
```diff
@@ -1,29 +1,70 @@
...
-//! abstracts over both using standard [http](https://github.com/hyperium/http) types allowing
-//! you to focus more on your application while giving you to the flexibility to
-//! transparently use whichever http trigger suits your application's needs best.
+//! This crate abstracts over all of these trigger events using standard [`http`](https://github.com/hyperium/http) types minimizing the mental overhead
+//! of the implementation details allowing you to focus more on your application while also giving you to the maximum flexibility to
+//! transparently use whichever lambda trigger suits your application's needs best.
 //!
 //! # Examples
 //!
+//! ## Hello world
```
davidbarsky, line 12:
```suggestion
//! ## Hello World, Without Macros
```

lambda-http/src/lib.rs
```diff
@@ -1,29 +1,70 @@
...
+//!
+//! lambda_http handlers adapt to the standard `lambda::Handler` interface using the [`handler`](fn.handler.html) function.
+//!
+//! The simplest case http handler is a closure of `http::Request` to a type that can be lifted into an `http::Response`.
```
davidbarsky, line 16: I'm not parsing this sentence. Can you reword?

lambda-http/src/lib.rs
```diff
@@ -1,29 +1,70 @@
...
-//! transparently use whichever http trigger suits your application's needs best.
+//! This crate abstracts over all of these trigger events using standard [`http`](https://github.com/hyperium/http) types minimizing the mental overhead
+//! of the implementation details allowing you to focus more on your application while also giving you to the maximum flexibility to
+//! transparently use whichever lambda trigger suits your application's needs best.
 //!
 //! # Examples
 //!
+//! ## Hello world
+//!
+//! lambda_http handlers adapt to the standard `lambda::Handler` interface using the [`handler`](fn.handler.html) function.
```
davidbarsky, line 14:
```suggestion
//! `lambda_http` handlers adapt to the standard `lambda::Handler` interface using the [`handler`](fn.handler.html) function.
```

lambda-http/src/lib.rs
```diff
@@ -1,29 +1,70 @@
...
+//! type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
+//!
+//! #[tokio::main]
+//! async fn main() -> Result<(), Error> {
+//!     lambda::run(handler(|request| async { Ok("👋 world!") })).await?;
+//!     Ok(())
+//! }
+//! ```
+//!
+//! ## Hello world (simpler)
```
davidbarsky, line 30:
```suggestion
//! ## Hello World
```

Let's also show the simple example first?

lambda-http/src/lib.rs
```diff
@@ -1,29 +1,70 @@
...
+//!
+//! For the simple cases you may not need much if any bootstrapping. To make life simpler
+//! you can add an `#[lambda_http]` attribute to your `main` function and `lambda::run` machinery
+//! will be wired in for you.
```
davidbarsky, line 34:
```suggestion
//! Adding an `#[lambda_http]` attribute to a `#[tokio::run]`-decorated `main` function will setup and run the Lambda function.
```

lambda-http/src/request.rs
```diff
@@ -19,90 +19,148 @@ use crate::{
...
+#[derive(Deserialize, Debug, Clone)]
+#[serde(untagged)]
+pub enum RequestContext {
+    /// API Gateway v2 request context
+    ApiGatewayV2(ApiGatewayV2RequestContext),
+    /// API Gateway request context
+    ApiGateway(ApiGatewayRequestContext),
+    /// ALB request context
+    Alb(AlbRequestContext),
+}
```
davidbarsky, line 162: Mark this as `#[non_exhaustive]`.

lambda-http/src/request.rs
```diff
@@ -19,90 +19,148 @@ use crate::{
...
+
+#[derive(Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
+pub struct AlbRequestContext {
+    pub elb: Elb,
+}
+
+/// Event request context as an enumeration of request contexts
+/// for both ALB and API Gateway and HTTP API events
```
davidbarsky, line 152:
```suggestion
/// Support invocations from Application Load Balancers and API Gateway. Supported API Gateway invocations
/// include both [REST](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-rest-api.html) and [HTTP](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api.html) APIs.
```

lambda-http/tests/data/apigw_v2_proxy_request.json
```diff
@@ -0,0 +1,39 @@
+{
```
davidbarsky, line 1: For later: let's create a JSON schema based off this file (https://app.quicktype.io/ is great) and fuzz/quickcheck the handler based off the schema?

lambda-http-attributes/Cargo.toml
softprops: The way this works now is that the attribute crates are transient dependencies users don't have to manually manage these.

I have some GitHub workflows for publishing crates easy as git pushing a tag for automating releases I could share with this project.

I was trying to find a way to do this with lambda attributes but couldn't find a way do to it without adding a unnatural and cyclical dependency on lambda-http in the lambda crate.

One idea I haven't excersized is compilation flags set by the lambda and lambda-http crates respectively. Thoughts on that approach?

I really like the procmacro-main pattern but it's notably a bit boilerplatey to set up which makes it seems like the yet-another-crate-to-publish feelings are strong. I can sympathize with that.

lambda-http/Cargo.toml
softprops: That's an improvement. What I'm going for is a way to way to use any off the off the shelf http triggers adapted to rust's http crate types. The event types are in here but they are more of an implementation detail than a feature.

lambda-http/examples/basic.rs
softprops: Rust convenience and extensibility pattern. 

For some common cases you'll likely just want to yield a string or serde json value. IntoResponse is a convince here, similar the new http apis interface for implicit json  responses. In other cases where you want a more application specific way of structuring responses for consistency, you can just return your own type and impl into response to avoid the conversion inside the hander. Side effect, it makes that a bit wiring more unit testable. 

In short make the simple cases easy and complex cases possible.

Concretely in this case there are two match arms one for the happy path which returns a string and one for the user error case which builds a response with an error status. The common interface for the result type becomes some type impl IntoResponse.

lambda-http/tests/data/apigw_v2_proxy_request.json
softprops: Interesting you mentioned that. I was actually think about starting a fuzzer project for lambdas, a cli, similar to Sam's templated events  but random and standalone. But dully noted for unit testing as well.

lambda-http/src/ext.rs
softprops: Your call, will do. 

The reason I run with this approach in tests is that when Ive observed others new to rust learn , they start out with cut and paste approach to learning. Often in tests and example bins you'll find unwrap and expects instead of results. I found this a bad habit make for rust new comers. This and or a surprise to see a rust program panic :)

lambda-http/tests/data/apigw_v2_proxy_request.json
davidbarsky: (note: this isn't a blocker for this PR; tracking this in an issue is fine.)

One of the reasons the `simulated.rs` module exists is to support this use-case _without_ something like SAM. I haven't spoken to the SAM team about this, however.

lambda-http/src/ext.rs
softprops: >  I'd rather prefer the full module path for all `hyperium/http`

I'm going to interpret that as including the non-top level `http::header` type symbols as well.

softprops: I just realized I can snip some of these since `http::HeaderMap` as a default type parameter set to `http::header::HeaderValue`. I'll do that to cut down the qualification noise this adds.

softprops: I just noticed there are number of other cases in the `lambda` module that use a `use http::Request`. I'm a big fan of consistency. Should we also move those to local `http::XXX` name qualification to be consistent?

lambda-http-attributes/Cargo.toml
softprops: I took a look at what the `tracing` crate does. It seems it [uses "asyncness'  as a cue](https://github.com/tokio-rs/tracing/blob/5a971befacc233612ec6577c9bae6188c79926ad/tracing-attributes/src/lib.rs#L273) to generate code using tracing_futures  It's not actually aware if tracing_futures is on the dependency graph or not. There's a little less luxury here. 

I'll see if I can use the type as the first argument to main as a cue to generate lambda http references then fallback on the cargo feature flag if not.

dbasedow:
I think `cookies` should be an Option<...> in LambdaRequest::ApiGatewayV2 in my tests the lambda event did not contain that key and i got a "data did not match any variant of untagged enum LambdaRequest" error.

Great work on the PR. It's exactly what I was looking for!

lambda-http-attributes/Cargo.toml
davidbarsky: > I took a look at what the tracing crate does. It seems it uses "asyncness' as a cue to generate code using tracing_futures It's not actually aware if tracing_futures is on the dependency graph or not. There's a little less luxury here.

Ah yeah, I forgot how we (I?) wrote that.

> I'll see if I can use the type as the first argument to main as a cue to generate lambda http references then fallback on the cargo feature flag if not.

Trying to figure out a type in a macro expansion isn't easy, least of all because I ran tried to resolve poor interactions with the `async_trait` and tracing's `#[instrument]`. You'd need something like specialization to properly distinguish between values in a macro, and short of that, autoref specialization: https://github.com/dtolnay/case-studies/blob/master/autoref-specialization/README.md. I'm not sure it's worth tackling now, but if you come up with solution, please, I'm all ears!

softprops:
> I think cookies should be an Option<...> in LambdaRequest::ApiGatewayV2 in my tests the lambda event did not contain that key and i got a "data did not match any variant of untagged enum LambdaRequest" error

Nice catch. I deployed a lambda with the v2 payload enabled and grabbed an example of an "echo" of its event payload then added a unit test to make sure the minimal case is covered in https://github.com/awslabs/aws-lambda-rust-runtime/pull/217/commits/94b18242474ffee45c18cc544de94143974cb65c

softprops:
rebased on master and found some changed to the Handler::Err which is now Handler::Error. Working through accommodating those

lambda-http-attributes/Cargo.toml
softprops: I realized that procmaco doesn't really know about types, only syntax so distinguishing a type seems less feasible. I'll try a cargo feature flag something like a feature for raw handlers and http handlers.

```toml
[features]
raw = []
http = []
defaults = [raw]
```

```
[dependencies]
lambda-attributes = { path = "../lambda-attributes", version = "0.1", default_features = false, features = [http] }
```

then do `#[cfg_attr(feature = "raw")]` vs `#[cfg_attr(feature = "http")]` checks

softprops: hit a small snag here I wanted to share early feedback on.  I added the features mentioned above to lambda-attributes then in its impl

```
if cfg!(feature = "http") {
 // handler code for http events
} else {
 // handler code for raw events
}
```

I'm able to successfully run `cargo test -p lambda` and `cargo test -p lambda_http` separately but when running `cargo test` the block for lambda_http gets emitted from the proc-macro presumably because cargo pins one version of the lambda-attribute dependency with one set of features for all crates in your workspace. This seems like a bug in cargo or it could be a feature. Either way its a bit problematic for the cargo feature flag approach.

```
error[E0433]: failed to resolve: use of undeclared type or module `lambda_http`
 --> lambda/examples/hello.rs:6:1
  |
6 | #[lambda]
  | ^^^^^^^^^ use of undeclared type or module `lambda_http`
```

softprops: I believe this behavior relates to this (open) cargo issue https://github.com/rust-lang/cargo/issues/4463.  I expect workspaces projects for rust lambdas be common with a mix of http handlers and some coordinating (sqs, s3, ect) handlers in the same project which makes me weary of the cargo feature flag approach.

softprops: I found a way out that I actually like better using item attr args that could set the stage for other strictly typed handlers

`#[lambda]`  does what it does today

`#[lambda(http)]` emits http handler code

softprops:
@davidbarsky I think I addressed the comments above. Let me know if there are any more blockers. 

a few side question/comment which should maybe be a separate gh issue.

I've love to see the versions of theses crates keep a shared version making it easier for users to manage. It's subjective but the cost of publishing all crates when on changes is less than that of a user's mental overhead of keeping track of the combinations of versions of crates all published from the same project.

What are you're thoughts on making all versions the same in this project.

softprops:
Is there any objective reason the minimum supported version in ci can't be `stable`? I noticed a ci failure when I introduces a `matches!` call introduced in [1.42](https://blog.rust-lang.org/2020/03/12/Rust-1.42.html)

davidbarsky:
> @davidbarsky I think I addressed the comments above. Let me know if there are any more blockers.

No, nothing that should block merging of this PR, I think. We can capture it in followups.

> I've love to see the versions of theses crates keep a shared version making it easier for users to manage. It's subjective but the cost of publishing all crates when on changes is less than that of a user's mental overhead of keeping track of the combinations of versions of crates all published from the same project.
> What are your thoughts on making all versions the same in this project.

Yeah, I think I'd be in favor of publishing everything/re-exporting everything through the Lambda crate and enabling subcrates through feature flags. I'd diverge from Tokio and have _everything_ enabled by default, though. I think a blocker to this would be to collapse the two macro crates into a single one and try to have a generic `Into` bound on the handler definition, if that makes sense.

> Is there any objective reason the minimum supported version in ci can't be stable? I noticed a ci failure when I introduces a matches! call introduced in 1.42?

I didn't know that was a thing! Let's make our MSRV the current stable Rust release until we hit 1.0. Our versioning policy should allow for bug fix releases to rely on a recent Rust release, but we should document this in a prominent location.

davidbarsky:
I'll merge this in and we catch any fallout of issues later on.

Issue #84, opened 2019-02-14
AWS CLI Instructions README Typo


dbettin:
There is a typo within the AWS CLI deploy instructions in the README.

`--zip-file file://./lambda.zip \` should be `  --zip-file fileb://./lambda.zip \`

Darrenmeehan:
This is address in #219

PR #219, opened 2020-04-28
Documentation fixes with some linting

Darrenmeehan:
*Description of changes:*

- Fixed documentation in `README.md` to work with existing code.
- Update awscli commands to work with version 2.
- Ensure files built when following `README.md` are ignored from `git`
- Remove Travis button as it's no longer used.
- Resolves https://github.com/awslabs/aws-lambda-rust-runtime/issues/84

Some code changes. I can revert this to make this PR cleaner if deemed more appropriate.

- Fixed minor typos in docstring.
- Minor linting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ Now that we have a deployment package (`lambda.zip`), we can use the [AWS CLI](h
 ```bash
 $ aws lambda create-function --function-name rustTest \
   --handler doesnt.matter \
-  --zip-file file://./lambda.zip \
+  --zip-file fileb://./lambda.zip \
   --runtime provided \
   --role arn:aws:iam::XXXXXXXXXXXXX:role/your_lambda_execution_role \
   --environment Variables={RUST_BACKTRACE=1} \
@@ -87,6 +87,7 @@ You can now test the function using the AWS CLI or the AWS Lambda console
 
 ```bash
 $ aws lambda invoke --function-name rustTest \
+  --cli-binary-format raw-in-base64-out \
   --payload '{"firstName": "world"}' \
   output.json
 $ cat output.json  # Prints: {"message":"Hello, world!"}
--- a/README.md
+++ b/README.md
@@ -85,11 +85,13 @@ You can now test the function using the AWS CLI or the AWS Lambda console
 
 ```bash
 $ aws lambda invoke --function-name rustTest \
-  --cli-binary-format raw-in-base64-out \
   --payload '{"firstName": "world"}' \
   output.json
 $ cat output.json  # Prints: {"message":"Hello, world!"}
 ```
+
+**Note:** `--cli-binary-format raw-in-base64-out` is a required
+  argument when using the AWS CLI version 2. [More Information](https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-binaryparam)
 
 #### Serverless Framework
 
--- a/README.md
+++ b/README.md
@@ -62,13 +62,13 @@ There are currently multiple ways of building this package: manually, and the [S
 To deploy the basic sample as a Lambda function using the [AWS CLI](https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-welcome.html), we first need to manually build it with [`cargo`](https://doc.rust-lang.org/cargo/). Since Lambda uses Amazon Linux, you'll need to target your executable for an `x86_64-linux` platform.
 
 ```bash
-$ cargo build -p lambda_runtime --example basic --release
+$ cargo build -p lambda --example hello --release
 ```
 
 For a custom runtime, AWS Lambda looks for an executable called `bootstrap` in the deployment package zip. Rename the generated `basic` executable to `bootstrap` and add it to a zip archive.
 
 ```bash
-$ cp ./target/release/examples/basic ./bootstrap && zip lambda.zip bootstrap && rm bootstrap
+$ cp ./target/release/examples/hello ./bootstrap && zip lambda.zip bootstrap && rm bootstrap
 ```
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -87,6 +85,7 @@ You can now test the function using the AWS CLI or the AWS Lambda console
 
 ```bash
 $ aws lambda invoke --function-name rustTest \
+  --cli-binary-format raw-in-base64-out \
```
softprops, line 88: It might be worth calling out you need the aws  v2 cli for this to work. Some are using that but as it's relatively new, many are not.

lambda/Cargo.toml
```diff
@@ -2,6 +2,7 @@
 name = "lambda"
 version = "0.1.0"
 authors = ["David Barsky <dbarsky@amazon.com>"]
```
softprops, line 4: I might drop the "client" part of this. That's an implementation detail most users won't likely care or care to know about. Just calling this "AWS Lambda Rust runtime" might get the point across and be a little bit more inline with https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

README.md
softprops: For clearer context https://docs.aws.amazon.com/cli/latest/userguide/cliv2-migration.html#cliv2-migration-binaryparam

Darrenmeehan: > It might be worth calling out you need the aws v2 cli for this to work. Some are using that but as it's relatively new, many are not.

Great call. Should I add the invokcation command for both versions? Or highlight the need for `--cli-binary-format` with v2? I'm learning towards the latter as it's less verbose.

lambda/Cargo.toml
Darrenmeehan: Thanks for the feedback. I've addresses this in f3337dbd35194865f8d476fa4cf3448265864dc4

README.md
Darrenmeehan: I've went with the following but happy to change! 56a745f0c9b87b9dc2ef369846f60a771da0bc17

davidbarsky: Yeah, let’s prefer version two of the CLI. I’d be surprised if the audience for this crate aren’t also early adopters of the newer CLI.

Darrenmeehan: My apologies it was late when I made the above commit https://github.com/awslabs/aws-lambda-rust-runtime/commit/56a745f0c9b87b9dc2ef369846f60a771da0bc17. It should say `most users are using v1`. 

I think as it stands the `README.md` does a good job of documenting the command for both v1 & v2. Is there any change you would like here @davidbarsky ?

davidbarsky: Oh, yeah. No change then. Thanks for doing this!

lambda/src/lib.rs
```diff
@@ -6,10 +6,10 @@
 //! There are two mechanisms of defining a Lambda function:
 //! 1. The `#[lambda]` attribute, which generates the boilerplate needed to
 //!    to launch and run a Lambda function. The `#[lambda]` attribute _must_
-//!    be placed on an asynchronous main funtion. However, asynchronous main
-//!    funtions are not legal valid Rust, which means that a crate like
+//!    be placed on an asynchronous main function. However, asynchronous main
+//!    functions are not legal valid Rust, which means that a crate like
 //!    [Runtime](https://github.com/rustasync/runtime) must be used. A main function
-//!    decorated using `#[lamdba]`
+//!    decorated using `#[lambdas]`
```
davidbarsky, line 12:
```suggestion
//!    decorated using `#[lambda]`s
```

Darrenmeehan: Thanks for the suggestion.

I love this new GitHub feature :+1:

lambda/src/lib.rs
```diff
@@ -6,10 +6,10 @@
 //! There are two mechanisms of defining a Lambda function:
 //! 1. The `#[lambda]` attribute, which generates the boilerplate needed to
 //!    to launch and run a Lambda function. The `#[lambda]` attribute _must_
-//!    be placed on an asynchronous main funtion. However, asynchronous main
-//!    funtions are not legal valid Rust, which means that a crate like
+//!    be placed on an asynchronous main function. However, asynchronous main
+//!    functions are not legal valid Rust, which means that a crate like
 //!    [Runtime](https://github.com/rustasync/runtime) must be used. A main function
```
davidbarsky, line 11: (Sorry for the back and forth on this; I reviewed this before my coffee hit.)

Can we reword this to say something along the lines of "Tokio is required; we suggest using `#[tokio::main]`"?

davidbarsky: Thanks! I only noticed this docs issue now since I wrote that; the two big issues are:

1. `runtime` is deprecated.
2. This runtime has a hard Tokio dependency (for now). Cross-runtime support is too hard, and until Rusoto supports runtimes other than Tokio, I don't think its worth _attempting_ to support other runtimes in here.

Darrenmeehan:
@davidbarsky Could this PR be linked with the following issue please? https://github.com/awslabs/aws-lambda-rust-runtime/issues/84

davidbarsky:
Yup! If you’re to add a resolves [issue number] to the description of this PR, it’ll link and close that issue.

lambda/src/lib.rs
Darrenmeehan: I'm not totally familiar with `tokio` but I've attempted to clear this section up. Anything you would like changed? Sorry if I'm misunderstanding something!

Darrenmeehan:
Hi @davidbarsky is there anything you would like changed?

davidbarsky:
@Darrenmeehan No, this looks great! Sorry for the delay and thanks for the fix!

Darrenmeehan:
Awesome thank you for the quick feedback. Looking forward to contributing some more.

Issue #218, opened 2020-04-28
Latest Master Causes Runtime Errors


jatsrt:
Updated just a little bit ago to the latest master, after the "cleanup".  All lambda jobs that I have running now are erroring out with runtime errors:

`Process exited before completing request`

Simple example that causes this...

```rust
#[tokio::main]
async fn main() -> Result<(), Box<dyn std::error::Error + Send + Sync + 'static>> {
    let func = handler_fn(|event| handler(event));
    lambda::run(func).await?;
    Ok(())
}

async fn handler(_event: Value) -> Result<()> { Ok(()) }
```

I am not getting any error to look at if I log all errors that I can.  I can also confirm that it does run the handler to completion, but then exits right away.

jatsrt:
@davidbarsky any thoughts on what change could be causing this?  Except of the introduction of genawaiter, doesn't seem like too much changed.

jatsrt:
Not positive, but is it possible that this is just terminating the thread if there is not a `NextEventRequest` ready to be processed right when it completes.  So the gen/yield will yield the first execution, and will continue if the "queue is full", but terminates when it should be waiting for the next item to show up.

jatsrt:
I believe this is all that is needed.  Otherwise you only ever yield on value, then the generate completes and the runtime closes down.

```rust
    gen!({
        loop {
            let req = NextEventRequest.into_req().expect("Unable to construct request");
            yield_!(client.call(req).await)
        }
    })
```

davidbarsky:
@jatsrt Thanks for the catch and the fix. I've merged your PR.

PR #220, opened 2020-04-28
Loop the generator to continue yeilding values

jatsrt:
*Issue #218*

*Loop the generator to continue to yield values, otherwise a single value is yielded and the runtime is shutdown*


softprops:
Just curious, how hard would it be to derive a unit test for this?

davidbarsky:
Not difficult at all. We’d need to simulate multiple invocations. Us moving to an Rng-driven UUID generation mechanism would make this far easier.

PR #222, opened 2020-05-01
Wrap the handler in order to catch panics

### LINKED ISSUE #221: Handler panics hang the runtime
If a handler panics it will get picked up by the executor but it will just hang.

Example:
```rust

#[tokio::main]
    lambda::run(lambda::handler_fn(handler)).await?;
    Ok(())
}

#[derive(Debug)]
enum MyError {
    Unknown,
}

impl std::fmt::Display for MyError {
        write!(
            f,
            "{}",
            match self {
                MyError::Unknown => "Unknown",
            }
        )
    }
}

impl std::error::Error for MyError {}

async fn handler(_: serde_json::Value) -> Result<serde_json::Value, MyError> {
    panic!("imagine an unwrap or unimplemented! here");
}
```

It's possible to work around this by introducing a shim:
```rust
[tokio::main]
    lambda::run(lambda::handler_fn(wrapped_handler)).await?;
    Ok(())
}

async fn wrapped_handler(event: serde_json::Value) -> Result<serde_json::Value, MyError> {
    let handle = tokio::spawn(async { handler(event).await }).await;
    handle.map_err(|_| MyError::Unknown.into())?
}
```

medwards:
Fixes #221

*Description of changes:*
Draft attempt to push panic handling up to the runtime. **Does not compile right now**

- [*] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [*] I confirm that I've made a best effort attempt to update all relevant documentation.

### DIFF
--- a/lambda/src/lib.rs
+++ b/lambda/src/lib.rs
@@ -211,15 +211,27 @@ where
         let body = serde_json::from_slice(&body)?;
 
         let request_id = &ctx.request_id.clone();
-        let f = INVOCATION_CTX.scope(ctx, { handler.call(body) });
+        // handler, B need to be Send
+        // let f = tokio::spawn(INVOCATION_CTX.scope(ctx, async move { handler.call(body).await }));
+        let f = tokio::spawn(async { handler.call(body).await });
 
         let req = match f.await {
-            Ok(res) => EventCompletionRequest { request_id, body: res }.into_req()?,
-            Err(e) => EventErrorRequest {
+            Ok(res) => match res {
+                Ok(r) => EventCompletionRequest { request_id, body: r }.into_req()?,
+                Err(e) => EventErrorRequest {
+                    request_id,
+                    diagnostic: Diagnostic {
+                        error_message: format!("{:?}", e),
+                        error_type: type_name_of_val(e).to_owned(),
+                    },
+                }
+                .into_req()?,
+            },
+            Err(panic) => EventErrorRequest {
                 request_id,
                 diagnostic: Diagnostic {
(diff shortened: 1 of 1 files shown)


lambda/src/lib.rs
```diff
@@ -211,15 +211,27 @@ where
...
+                    },
+                }
+                .into_req()?,
+            },
+            Err(panic) => EventErrorRequest {
                 request_id,
                 diagnostic: Diagnostic {
-                    error_message: format!("{:?}", e),
-                    error_type: type_name_of_val(e).to_owned(),
+                    error_message: "Unexpected panic".to_string(),
```
softprops, line 233: Unless I'm reading this wrong, it looks like we are discarding useful debugging information like the type name of the error and the debug display of the error. Is this intended?

medwards: You're right, that's the other part of why this is in draft. I wasn't able to figure out how to get something meaningful out of the panic info yet.

medwards:
> Does this prevent any the ability for users to set their own panic hook bevavior? An motivator for enabling that would be to allow for users to export unhandled errors into an error tracking system of their choice

That's an interesting angle to take too. Users who want to track their own panics might be able to wrap them in yet another `tokio::spawn` though I haven't experimented along those lines yet. This wouldn't be that different from what users have to do today anyways so it wouldn't be a heavy imposition.

I'll keep this in mind though and try to provide some experiments along these lines when I get this to compile.

bahildebrand:
Haven't seen any progress on this draft in a while. Closing for now. Feel free to re-open if you make more progress.

Issue #229, opened 2020-05-21
Handler Error type on Master


aklitzke:
On current master, the handler requires an `anyhow::Error` as the error type for the handler, as opposed to `Box<dyn std::error::Error + Send + Sync + 'static>` in previous versions. However, this new type isn't re-exported from lib, requiring consumers of this crate to also import `anyhow`. Is this going to stick around? Maybe it would make sense to reexport your error type?

davidbarsky:
No, it should be `Box<dyn std::error::Error + Send + Sync + 'static>`. This is a bug, sorry about that. I was lazy in some refactoring and forgot to remove `anyhow::Error`.

softprops:
I got this. Covered in https://github.com/awslabs/aws-lambda-rust-runtime/pull/231

davidbarsky:
Sorry about that, @aklitzke! This should be resolved now.

aklitzke:
Awesome, thanks!

Not sure what your strategy is around testing, but should there maybe be some testing around this? Surprised you'd be able to make a breaking change to the api like that and not have some tests/examples fail to compile or something

softprops:
Few thoughts there. 

1) the state of the current code is always going to be in sync with the tests by virtual of ci. If apis change the tests likely will as well

2) the state of the master branch reflects the current state of development, not a release. Depending on the current state of development comes with the higher risk of changes commit to commit up until a release

3) the next release will not be a patch release! Expect some changes from what's hosts on crates.io. all good and useful changes (we hope)

4) I think it's fair to have some tests that capture a contract for the error type but in the case of current master we've opted not to expose a type alias and instead rely on users to Box up a dyn std Error. Internally we use a type alias. Perhaps on thing that could be improved is type tests to capture it's properties so that when it's type does change we know we're sticking with those well known properies

PR #231, opened 2020-05-26
replace anyhow with std error and an type alias

softprops:
*Issue #229, if available:*

*Description of changes:*

Replaced anyhow with std::error::Error ~type alias~. 

also renamed some lambda-http examples. cargo warns that when you have to crates with examples with the same name, collisions ensue. I gave these unique names


softprops:
I can do `pub (crate)`. The idea behind making it `pub` was do abbreviate the api a bit. I'd be happy to defer to that later if it becomes too onerous to type out in each lambda.

softprops:
Done, hindsight in that last commit is that is a very laborious process to update code references when this type changes. In most ideal cases though it won't change so that's maybe something we don't have to optimize for now.

davidbarsky:
Yeah, it’ll probably be a pain, but not exporting avoids a future compatibility issue.

lambda-http/src/lib.rs
```diff
@@ -98,14 +101,13 @@ mod strmap;
...
-    error::Error,
-    fmt,
     future::Future,
     pin::Pin,
     task::{Context, Poll},
 };
 
-type Err = Box<dyn Error + Send + Sync + 'static>;
+/// Error type that lambdas may result in
+pub(crate) type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
```
mlafeldt, line 110: @softprops I'm curious. Is there a reason for not exposing this error type via `pub`?

At the moment, all crate users have to define the type themselves in each Lambda function:

```rust
```

davidbarsky: Exporting this would introduce a forwards-compatibility hazard if we change any of these bounds. At the moment, you _don't_ need to write this type alias; you only need a type that can be converted into `Box<dyn std::error::Error + Send + Sync + 'static>`. Most errors, barring a (except a few like `std::sync::PoisonError`) _can_ be converted to a boxed trait object.

mlafeldt: Thanks for the clarification. Makes sense. 👍

Issue #216, opened 2020-04-12
State of release


ajpauwels:
Hey all, would it be possible to cut a release at the current master commit level or similar? As long as it's post #111 that should work.

Coming in as a new user I was confused between the differing formats in the examples source and in the docs. It didn't help that the examples source (which is the correct way to do things) doesn't compile with the latest version of the crate. Once the dependency is renamed to `lambda = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", branch = "master" }` everything works for the new syntax.

I can try and provide documentation and examples improvements PRs but if the maintainers could cut a release I think that'd go a long way to usability.

outkine:
Can confirm that using the `lambda_runtime` dependency (as directed by the README) does not compile because of an issue with an old version of the `time` library:
```
error: expected an item keyword
   --> .../time-0.2.8/src/utc_offset.rs:366:13
    |
366 |             let tm = timestamp_to_tm(datetime.timestamp())?;
    |             ^^^
```

softprops:
unfortunately the readme has gone stale. I can help with this. 

You should only need the `lambda` crate for basic use. This crate simplified the previous versions and brought support for async/await into the fold. There is some inflight work to bring the `lambda-http` crate up to speed with the new `lambda` crate interfaces.

At this point you might be best served by using the examples [here](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda/examples) adapting the packaging and deployment notes from the readme.

Let us know if you get stuck.

Luminoth:
Are there plans to cut a new release tho? Or should we plan to keep using the github dependency for a while still?

I'm also curious when the hello-with-ctx example is going to be expanded on. It isn't super clear to me how that's supposed to work with the macro-based implementation and the example is just copy / paste the hello example right now. Everything else seems more or less clear from the examples and a little bit of trial and error (for instance the examples deal with serde Values for input / output but really they just need to derive the serde traits).

I guess one other question I personally have is - can this be used with proxy API gateways?

mickdekkers:
@Luminoth I'm not sure how the context is meant to be accessed, but this seems to work:
```rust
use lambda::{lambda, INVOCATION_CTX};
use serde_json::Value;

type Error = Box<dyn std::error::Error + Send + Sync + 'static>;

#[lambda]
#[tokio::main]
async fn main(event: Value) -> Result<(), Error> {
    INVOCATION_CTX.with(|ctx| {
[...]

PR #233, opened 2020-06-27
Make lambda::Context a first class part of the Handler api

softprops:
*Issue #, if available:*

*Description of changes:*

This makes LambdaCtx a second argument passed to handler types

This makes for a bit more consistency with handler apis in other runtimes like [node](https://docs.aws.amazon.com/lambda/latest/dg/nodejs-handler.html), [python](https://docs.aws.amazon.com/lambda/latest/dg/python-handler.html), and [ruby](https://docs.aws.amazon.com/lambda/latest/dg/ruby-handler.html)

> [go is an outlier](https://docs.aws.amazon.com/lambda/latest/dg/golang-handler.html) because of its native [context pattern idiom](https://blog.golang.org/context)

See conversation [here](https://github.com/awslabs/aws-lambda-rust-runtime/issues/216)

update:

This change also renames `lambda::LambdaCtx` to `lambda::Context`


lambda-attributes/src/lib.rs
```diff
@@ -38,42 +38,48 @@ pub fn lambda(attr: TokenStream, item: TokenStream) -> TokenStream {
...
                 quote_spanned! { input.span() =>
 
-                    use lambda::LambdaCtx;
-
                     #(#attrs)*
                     #asyncness fn main() {
-                        async fn actual(#arg_name: #arg_type) #ret {
-                            #body
-                        }
+                        async fn actual(#event_name: #event_type, #context_name: #context_type) #ret #body
```
softprops, line 81: note: I ditched the branches because I noticed a warning emitted about needless braces. we debugging the result I could see that body actually included braces so the result looked like `async fn actual(foo: XXX, bar: XXX) -> Ret { { ... } }` hence I omitted the extra braces to avoid potential user confusion with rust warnings

lambda-attributes/src/lib.rs
```diff
@@ -38,42 +38,48 @@ pub fn lambda(attr: TokenStream, item: TokenStream) -> TokenStream {
...
+                    };
+                    return TokenStream::from(tokens);
+                }
+            };
+            let context_name = &context.pat;
+            let context_type = &context.ty;
 
             if is_http(&args) {
                 quote_spanned! { input.span() =>
-                    use lambda_http::lambda::LambdaCtx;
```
softprops, line 56: I'm not actually sure why these were here before but they cased duplicate import errors as handlers no import this type themselves in their function signatures

softprops:
Something that made me itch a bit is the name `LambdaCtx` A suggested alternative might be `lambda::Context`. If users have other `Context` types in their code one can always rename with `use lambda::Context as LambdaCtx`

cakekindel:
I've always really enjoyed type names that are very concise, and lean on the module to preserve the context of the type (`lambda::Context`)

softprops:
@cakekindel I'm with you :)

davidbarsky:
I think we should push customers (and in examples) to write `lambda::Context` wherever possible. The current naming is somewhat unfortunate/an oversight.

softprops:
Happy to add that here or after a merge. Which ever comes first

davidbarsky:
Eh, it’s more of a value statement. Happy to merge this thing in first.

softprops:
I ended out doing a quick find and replace

davidbarsky:
I think we should get rid of the `INVOCATION_CTX` wholesale (cc: @insanitybit) but I'll merge this in for now.

softprops:
> I think we should get rid of the INVOCATION_CTX wholesale

me too https://github.com/awslabs/aws-lambda-rust-runtime/pull/237

seanpianka:
FWIW, @softprops it was a breaking change going from #231 (`ed3fd16`) to #233 (`ae3ce1d`), at least for those whose Cargo.toml entry for this crate was tracking with `branch = "master"`. 

Here's my error:

```rust
error[E0593]: function is expected to take 2 arguments, but it takes 1 argument
   --> src/cmd/lambda/main.rs:91:25
    |
53  | async fn entrypoint(req: lambda_http::Request) -> Result<impl IntoResponse, AsyncError> {
    | --------------------------------------------------------------------------------------- takes 1 argument
...
91  |     lambda::run(handler(entrypoint)).await?;
    |                         ^^^^^^^^^^ expected function that takes 2 arguments
    |
   ::: /Users/sean/.cargo/git/checkouts/aws-lambda-rust-runtime-7c865cce90132439/ae3ce1d/lambda-http/src/lib.rs:131:19
    |
131 | pub fn handler<H: Handler>(handler: H) -> Adapter<H> {
    |                   ------- required by this bound in `lambda_http::handler`
    |
    = note: required because of the requirements on the impl of `lambda_http::Handler` for `fn(http::request::Request<lambda_http::body::Body>) -> impl std::future::Future {entrypoint}`
```

softprops:
@seanpianka this is expected. 

Note depending on master holds different guarantees than depending on a release version.

There are no guarantees on on API stability on the master branch as it represents active development. 

We are hoping to pin down what the next release version looks like which, for those depending on the latest release version, should expect a number of breaking changes

seanpianka:
I've since pinned the crate at `ed3fd16` using the following in Cargo.toml:
```toml
lambda_http = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", rev = "ed3fd16" }
```
As it seems the guarantees from depending on master are not suitable for my needs. :)

PR #235, opened 2020-06-27
automate publishing to crates.io on pushes to tags

softprops:
*Issue #, if available:*

*Description of changes:*

strawman: It's possible to automate the crates.io publishing process with GitHub actions which in theory lowers the bar of effort for publishing consistent releases. I've done this with good success in a few other projects. 

The trick here is to generate an crates.io api key [here](https://crates.io/me) and store it in your GitHub repositories secrets under `https://github.com/:owner/:repo/settings/secrets`. These secrets, much like like travis secrets, will not be printed in build logs.

The thing to consider when publishing workspaces projects is that crates.io will not let you publish a crate unless its dependencies are available in crates.io so you have to wait a bit between publishes. 

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,3 +88,33 @@ jobs:
         with:
           github_token: ${{ secrets.GITHUB_TOKEN }}
           publish_dir: ./target/doc
+
+  # automate publishing to crates.io on pushes to tags
+  publish-crates:
+    if: startsWith(github.ref, 'refs/tags/')
+    runs-on: ubuntu-latest
+    needs: [build, fmt]
+    steps:
+      - name: Set up Rust
+        uses: hecrj/setup-rust-action@v1
+      - uses: actions/checkout@v2
+      - name: Publish
+        if: env.CRATES_TOKEN
+        run: |
+          pushd lambda-attributes
+          cargo publish --token ${{ secrets.CRATES_TOKEN }}
+          popd
+          # eventual consistency in crates.io dictates we wait a bit before publishing
+          # so that lambda-attributes is available for resolution
+          sleep 20
+          pushd lambda
+          cargo publish --token ${{ secrets.CRATES_TOKEN }}
+          popd
+          # eventual consistency in crates.io dictates we wait a bit before publishing
+          # so that lambda is available for resolution
+          sleep 20
+          pushd lambda-http
+          cargo publish --token ${{ secrets.CRATES_TOKEN }}
+          popd
+        env:
+          CRATES_TOKEN: ${{ secrets.CRATES_TOKEN }}


cakekindel:
in the spirit of your strawman, you _could_ take the automation a step further!

This is probably a bit larger in scope than the change in this PR, but if the codebase used conventional commits you can configure your tag pushes, version bumps, and changelog generation to all be driven automatically by commit messages.

For example in my crates when I merge a PR, i have a job that bumps the version in my manifests, updates my changelog, and opens a PR. 

Then when I merge that PR, another job pushes my tags and publishes to crates.io - super painless and goof-proof

sorry if im just telling you things you already know, but if theres a good reason not to implement a process like this i'd love to know for my own projects!

softprops:
Thanks for the extra input. 

My experience is that commit message formatting can actually make it harder to contribute project, especially when it's enforced. When it's not enforced, it quickly loses it utility. I've really felt this with other projects that requires it. It made it that much harder to contribute when you're already volunteering free time. When you don't get the format it makes contributors feel like "goofs" when pushing days worth of work only to have ci tell you that it would like you do more work to rewrite your git commits with the format it wants.

That said there are useful benefits to structured commit messages like changleog generators. Commit message formats are a technical solution. This can also be solved with a social solution by requesting them as part of a pull request template.

Tag pushes publishing can be an incremental part of the path to automation made easy with GitHub provided tooling. Like any process, I'd defer on more structure process until we know its needed and I don't know that adding more structure is actually needed yet.

Tldr I agree with you structured commit message can provide value when enforced. It trades off of making a project easy to contribute to so you really want to be sure it would be worth it before investing in it

cakekindel:
Ah I totally get that! There are definitely ways around making it difficult for contributors to just grab an issue and run with it, like using squash and deferring the convention to the merge by the maintainers.

You can also _somewhat_ separate commit format enforcement from "ci driven by commit format" because tools like standard-version assume any commit out of spec is a minor bump (unless "BREAKING" is in the body)

On your point of introducing automation where it's needed: I am totally in agreement with you that automation & process amendments should be weighed against the project's needs before introducing it! 

My intent was just to bring awareness to a potential time saver / lighter lift on the maintainers since this project has gone without a release for a while - thanks for humoring me!

.github/workflows/build.yml
```diff
@@ -67,4 +67,34 @@ jobs:
           components: rustfmt
           override: true
       - name: Run fmt check
-        run: cargo fmt --all -- --check
\ No newline at end of file
+        run: cargo fmt --all -- --check
+
+  # automate publishing to crates.io on pushes to tags
+  publish-crates:
+    if: startsWith(github.ref, 'refs/tags/')
```
davidbarsky, line 74: Am I correct in thinking that we a GitHub release can _also_ create a git tag, which would then kick off this workflow?

davidbarsky:
I'd personally be interested in some sort of mechanism where we can bump versions programmatically, but publishing on a new git tag + release is sufficient for now. Am I correct in thinking that the release process for this would effectively be:

1. Cut a PR bumping the versions of the crates in question.
2. Merge that PR.
3. Create a GitHub release that creates a new git tag, thereby kicking off this publishing job?

.github/workflows/build.yml
softprops: Mm yes but its a little tricky because its the kind of thing that can cause an gh action workflow trigger loop. Let me know if its something you'd like to me to explore. I was aiming a quick mvp to get this started.

softprops:
> Am I correct in thinking that the release process for this would effectively be:
   Cut a PR bumping the versions of the crates in question.
   Merge that PR.
   Create a GitHub release that creates a new git tag, thereby kicking off this publishing job?

The last part is `git tag -a v0.X.X -m "here we go" && git push origin v0.X.X`

Since you typically want all the same verification checks that come with a push build this adds a conditional step at the end specifically for tags. So this validate everything is in ship shape then publish if everything checks out. Note the bit if weirdness for a workspace project. You have to script waiting a bit between publishes because cargo wont let you publish something that has a dependency that it can't resolve in its registry.

There's an open issue in cargo which would let cargo using its DAG knowledge to do the right thing for you with `cargo publish  --all` https://github.com/rust-lang/cargo/issues/1169 In theory this would do what you expect.

softprops:
My suggestion would be to test this with something like a beta release version

softprops:
https://github.com/awslabs/aws-lambda-rust-runtime/issues/216#issuecomment-657658627

Closing for now as aws disables gh secrets in it's repos and we'd need a way to store a crates io token

PR #236, opened 2020-06-28
Changelog what changed

softprops:
*Issue #, if available:*

*Description of changes:*

kick of for a changelog to be iterated on that captures what changed in prep for a release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,14 +26,15 @@
   +use lambda::{handler_fn, Context};
   use serde_json::Value;
   
+  +type Error = Box<dyn std::error::Error + Sync + Send + 'static>;
+
   -fn main() {
   -    lambda!(handler)
-  +type Error = Box<dyn std::error::Error + Sync + Send + 'static>;
   
   +#[tokio::main]
   +async fn main() -> Result<(), Error> {
-      lambda::run(handler_fn(handler)).await?;
-      Ok(())
+  +    lambda::run(handler_fn(handler)).await?;
+  +    Ok(())
   }
   
   -fn handler(
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
   The crate formerly known as `lambda_runtime` is now simply `lambda`, `lambda_http` remains the same.
 
-  Both crates have note worth breaking api changes. The following is a rough approximation is what a transition to the new runtime will look like for your handlers
+  Both crates have note worth breaking api changes. The following is a rough approximation of what a transition to the new runtime will look like for your handlers
 
   → `Cargo.toml`
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,59 @@
 # next (unreleased)
 
+- **New**: Almost everything!
+
+  Heavy restructuring was done in order to simply the runtime and to enable asynchronous handlers by default.
+
+  The crate formerly known as `lambda_runtime` is now simply `lambda`, `lambda_http` remains the same.
+
+  Both crates have note worth breaking api changes. The following is a rough approximation is what a transition to the new runtime will look like for your handlers
+
+  → `Cargo.toml`
+
+  ```diff
+  [dependencies]
+  - lambda_runtime = "0.2"
+  + tokio = { version = "0.2", features = ["macros"] }
+  + lambda = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", branch = "master"}
+    ```
+  
+  > note: the runtime is `std::future::Future` based and does not rely   directly on tokio but tokio serves a very good default runtime for   driving `std::future::Future`s to completion
+  
+  → `main.rs`
+  
+  ```diff
+  -use lambda_runtime::{error::HandlerError, lambda, Context};
+  +use lambda::{handler_fn, Context};
+  use serde_json::Value;
+  
+  -fn main() {
+  -    lambda!(handler)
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  +) -> Result<Value, Error> {
+      Ok(event)
+  }
+  ```
+  
+  Please see the examples directories for more inspiration.
+  
+- **New:** the `lambda_http` crate now supports [API Gateway HTTP API](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api.html) lambda triggers.
+
+# 0.2.1
```
softprops, line 56: technical there's a tag that didn't get pushed here but I don't think the world will end

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
 # next (unreleased)
 
+- **New**: Almost everything!
+
+  Heavy restructuring was done in order to simply the runtime and to enable asynchronous handlers by default.
```
Veetaha, line 5:
```suggestion
  Heavy restructuring was done in order to simplify the runtime and to enable asynchronous handlers by default.
```

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+
+  The crate formerly known as `lambda_runtime` is now simply `lambda`, `lambda_http` remains the same.
+
+  Both crates have note worth breaking api changes. The following is a rough approximation is what a transition to the new runtime will look like for your handlers
```
Veetaha, line 9:
```suggestion
  Both crates have note worth breaking api changes. The following is a rough approximation of what a transition to the new runtime will look like for your handlers
```

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+
+  → `Cargo.toml`
+
+  ```diff
+  [dependencies]
+  - lambda_runtime = "0.2"
+  + tokio = { version = "0.2", features = ["macros"] }
+  + lambda = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", branch = "master"}
```
Veetaha, line 17: What is the reason for a git reference here? Maybe it's better to reference crates.io (well the version the new major release will be published with)?

CHANGELOG.md
```diff
@@ -1,5 +1,60 @@
...
+  > note: the runtime is `std::future::Future` based and does not rely   directly on tokio but tokio serves a very good default runtime for   driving `std::future::Future`s to completion
+  
+  → `main.rs`
+  
+  ```diff
+  -use lambda_runtime::{error::HandlerError, lambda, Context};
+  +use lambda::{handler_fn, Context};
+  use serde_json::Value;
+  
+  +type Error = Box<dyn std::error::Error + Sync + Send + 'static>;
```
Veetaha, line 29: Hmm, just a small comment, this is not related to this PR, but having to repeat this type all over again is quite annoying.
Well, I think this is just a small nit to Rust itself, anyway, users can just do this type alias in their `lib.rs`. Ideally, I'd like to be able to use `anyhow::Result` for lambdas, but ridiculously `anyhow::Error` does not implement std::error::Error?!

CHANGELOG.md
Veetaha: Maybe it makes sense to add `(unreleased)` in parens near that tag not to tell a lie to people who are laggards and don't want to use new async major release, but only the latest `lambda!()` compatible version?

softprops: Good catch. Versioning is a bit odd here since technically there's a [0.2.1 release](https://crates.io/crates/lambda_runtime) also different release numbers for various crates in the same workspace and one git tag semantic.

I future direction I might suggest is to treat these as one versions unit while separate bundled artifacts. A consistent version makes managing changelings easier as easier for user to infer what versions of it's components work well together.

CHANGELOG.md
softprops: It will ultimately depend on what ever version we pick. I just lifted the diff from some real world migrations

CHANGELOG.md
softprops: This came up before. The driving argument was that the error crate marketplace was so flavorful that we wanted that to be a user choice and not a choice places on users. These crates go in an out of fashion in relatively short timespans so it would mean a breaking api change everyone the fashion changes.

I believe you should actually be able to use anyhow and am surprised that anyhow doesn't interop well with std Error

Paging dr @davidbarsky

davidbarsky: > I believe you should actually be able to use anyhow and am surprised that anyhow doesn't interop well with std Error

It doesn't due to trait coherence issues, unfortunately. If I recall correctly, the issue is relates to what types Error is implemented on, and until specialization lands _and is sound_, anyhow won't be able to interact with `std::error::Error`.

The runtime currently sets an error bound of `Into<Box<dyn std::error::Error + Send + Sync + 'static>>`, which is relatively open-ended. You _should_ be able to use `anyhow::Error` just fine due to this implementation: https://docs.rs/anyhow/1.0.31/anyhow/struct.Error.html#impl-From%3CError%3E. I'm personally bummed that we can't do any more fancy error reporting along the lines of https://github.com/yaahc/color-eyre/, which I personally love, but are arguably too opinionated for this library.

softprops:
Bump. I'd like to have something we can iterate on so progress can be made

coltonweaver:
We've had some discussion about this, I think the plan is to move this to Github changelog rather than a `changelog.md` file. Leaving this PR open to reference while we do that.

coltonweaver:
Have drafted this into a new release. Closing this PR in favor of that.

PR #238, opened 2020-07-13
Corrected --target name in README, added links.

rimutaka:
*Issue #, if available:* #216 

*Description of changes:*

The text was referring to `x86_64-linux` as a target, which is not even a valid value. From what I know, the only valid target is `x86_64-unknown-linux-musl`, so we may just as well say so. I lost at least an hour (novice!) trying to figure it out. 


davidbarsky:
D'oh, sorry about that!

Issue #232, opened 2020-05-31
Example has disappeared?


RandomInsano:
The sentence "The code above is the same as the basic example in the lambda-runtime crate." in the readme contains a dead link.

I dug around a little and didn't find anything that obviously matches.

softprops:
Sorry about that.

It's not that it's disappeared is that the readme reflects documentation for the released version while the state of the code base reflects the state of the _next_ version. This should get cleaned up for the next release but until then you can browse the source under the `v0.2.0` git tag. Here's the example you have have been looking for

https://github.com/awslabs/aws-lambda-rust-runtime/blob/v0.2.0/lambda-runtime/examples/basic.rs

rimutaka:
Was v0.2.0 code used for the published crate, which is listed as 0.2.1?
https://docs.rs/lambda_runtime/0.2.1/lambda_runtime/index.html

I'm trying to find the implementation of `lambda!` macro, but it looks like you got rid of it in the latest unpublished version. I couldn't find any traces of its implementation in the older code either. Apologies for this dummy question. I'm new to Rust.

brainstorm:
@rimutaka New to Rust here too. If you want to take a peek at some code that mostly works with `master` (at the time of writing this), take a peek at this branch (WIP) https://github.com/brainstorm/s3-rust-htslib-bam/commit/0b9fb022e7e347fd819990b901265917e2aa23cb... I'm currently struggling a bit locally with the `IntoResponse` type (in the middle of a local code refactor):

```
error[E0277]: the trait bound `http::response::Response<serde_json::value::Value>: lambda_http::response::IntoResponse` is not satisfied
  --> src/main.rs:29:59
   |
29 | async fn bam_header(event: Request, _: Context) -> Result<impl IntoResponse, Error> {
   |                                                           ^^^^^^^^^^^^^^^^^ the trait `lambda_http::response::IntoResponse` is not implemented for `http::response::Response<serde_json::value::Value>`
   |
   = help: the following implementations were found:
             <http::response::Response<B> as lambda_http::response::IntoResponse>
   = note: the return type of a function must have a statically known size
```

But both the example code from that branch as well as the `master` branch should work right out of the box.

[...]

PR #244, opened 2020-07-24
Added error handling for #241, interim, broken

rimutaka:
*Issue #, if available:* #241, relates to PR #242 

*Description of changes:*

Added my changes from #242 on top of David's changes in this branch.

1. The build is broken - lambda-http examples fail to compile.
2. The lambda examples (basic, hello, etc) compile, run, but return an error as a result and DO NOT LOG ANYTHING at all.

For example, hello-without-macro-tracing.rs has these 2 lines: 
```
simple_logger::init().unwrap();
info!("Hello!");
```
so it should at least log Hello. It doesn't.

This line returns an error for a test: `Err(simple_error::SimpleError::new("simple error").into())`.

Output:
```
{
  "errorType": "&alloc::boxed::Box<dyn std::error::Error+core::marker::Sync+core::marker::Send>",
  "errorMessage": "simple error"
}
```

Trace in CloudWatch:
```
START RequestId: 01e6ec13-33c2-40a7-b467-a37b1cbc17f0 Version: $LATEST
 TRACE hyper::proto::h1::conn: Conn::read_head
 DEBUG hyper::proto::h1::io: read 444 bytes
 TRACE hyper::proto::h1::role: Response.parse([Header; 100], [u8; 444])
 TRACE hyper::proto::h1::role: Response.parse Complete(418)
 DEBUG hyper::proto::h1::io: parsed 7 headers
 DEBUG hyper::proto::h1::conn: incoming body is content-length (26 bytes)
 TRACE hyper::proto::h1::decode: decode; state=Length(26)
 DEBUG hyper::proto::h1::conn: incoming body completed
 TRACE hyper::proto::h1::conn: maybe_notify; read_from_io blocked
 TRACE want: signal: Want    
 TRACE hyper::proto::h1::conn: flushed({role=client}): State { reading: Init, writing: Init, keep_alive: Idle }
 TRACE want: signal: Want    
 TRACE hyper::proto::h1::conn: flushed({role=client}): State { reading: Init, writing: Init, keep_alive: Idle }
 TRACE hyper::client::pool: put; add idle connection for ("http", 127.0.0.1:9001)
 DEBUG hyper::client::pool: pooling idle connection for ("http", 127.0.0.1:9001)
 TRACE want: signal: Want    
 TRACE hyper::proto::h1::conn: flushed({role=client}): State { reading: Init, writing: Init, keep_alive: Idle }
 TRACE hyper::client::pool: take? ("http", 127.0.0.1:9001): expiration = Some(90s)
[...]

PR #247, opened 2020-07-28
Bring README up to date with `master` branch

voteblake:
*Issue #, if available:*
#216, #232 

*Description of changes:*

In #216 @davidbarsky mentioned that README updates were the main blocker for the next release. I've taken a pass at making changes to the README to reflect the current state of `master`. This includes the new crate structure, the new options for creating a handler, and references to the new examples.

I have not updated the Docker or Serverless Framework parts of the README as I haven't used either of those options before.

I'm also not sure if the statement about MSRV on line 11 is still accurate.

I don't expect this to be ready to merge as-is, but hopefully gets us closer to a new tagged release!

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,58 +2,32 @@
 
 This package makes it easy to run AWS Lambda Functions written in Rust. This workspace includes multiple crates:
 
-- [![Docs](https://docs.rs/lambda_runtime_client/badge.svg)](https://docs.rs/lambda_runtime_client) **`lambda-runtime-client`** is a client SDK for the Lambda Runtime APIs. You probably don't need to use this crate directly!
-- [![Docs](https://docs.rs/lambda_runtime/badge.svg)](https://docs.rs/lambda_runtime) **`lambda-runtime`** is a library that makes it easy to write Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that makes it easy to write Lambda functions in Rust.
 - [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library that makes it easy to write API Gateway proxy event focused Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda_attributes/badge.svg)](https://docs.rs/lambda_attributes) **`lambda-attributes`** holds an [attribute macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) that runs an `async main` function in the Lamda runtime. You probably don't need to use this crate directly, the macro is exposed by **`lambda`**!
 
 ## Example function
 
-The code below creates a simple function that receives an event with a `greeting` and `name` field and returns a `GreetingResponse` message for the given name and greeting. Notice: to run these examples, we require a minimum Rust version of 1.31.
+The code below creates a simple function that receives an event and echoes it back as a response. Notice: to run these examples, we require a minimum Rust version of 1.31.
 
 ```rust,no_run
-use std::error::Error;
+use lambda::{lambda, Context};
+use serde_json::Value;
 
-use lambda_runtime::{error::HandlerError, lambda, Context};
-use log::{self, error};
-use serde_derive::{Deserialize, Serialize};
-use simple_error::bail;
-use simple_logger;
+type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
 
-#[derive(Deserialize)]
-struct CustomEvent {
-    #[serde(rename = "firstName")]
-    first_name: String,
-}
-
-#[derive(Serialize)]
-struct CustomOutput {
-    message: String,
-}
-
-fn main() -> Result<(), Box<dyn Error>> {
-    simple_logger::init_with_level(log::Level::Debug)?;
-    lambda!(my_handler);
-
-    Ok(())
-}
-
-fn my_handler(e: CustomEvent, c: Context) -> Result<CustomOutput, HandlerError> {
-    if e.first_name == "" {
-        error!("Empty first name in request {}", c.aws_request_id);
-        bail!("Empty first name");
-    }
-
-    Ok(CustomOutput {
-        message: format!("Hello, {}!", e.first_name),
-    })
+#[lambda]
+#[tokio::main]
+async fn main(event: Value, _: Context) -> Result<Value, Error> {
+    Ok(event)
 }
 ```
 
-The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/basic.rs) in the `lambda-runtime` crate.
+The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello.rs) in the `lambda` crate. An [alternative example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello-without-macros.rs) that uses functions instead of an attribute macro to create the handler is available as well.
 
 ### Deployment
 
-There are currently multiple ways of building this package: manually, and with the [Serverless framework](https://serverless.com/framework/).
+There are currently multiple ways of building this package: manually with the AWS CLI, and with the [Serverless framework](https://serverless.com/framework/).
 
 #### AWS CLI
 
@@ -91,9 +65,9 @@ You can now test the function using the AWS CLI or the AWS Lambda console
 
 ```bash
 $ aws lambda invoke --function-name rustTest \
-  --payload '{"firstName": "world"}' \
+  --payload '{"Hello,": "world!"}' \
   output.json
-$ cat output.json  # Prints: {"message":"Hello, world!"}
+$ cat output.json  # Prints: {"Hello,": "world!"}
 ```
 
 **Note:** `--cli-binary-format raw-in-base64-out` is a required
@@ -167,34 +141,13 @@ $ unzip -o \
 # Ctrl-D to yield control back to your function
 ```
 
-## lambda-runtime-client
+## `lambda`
 
-Defines the `RuntimeClient` trait and provides its `HttpRuntimeClient` implementation. The client fetches events and returns output as `Vec<u8>`.
+This library makes it easy to create Rust executables for AWS lambda. The library defines exposes an `#[lambda]` attribute macro. Adding the `#[lambda]` attribute to your `main` function allows you to define your event handler logic in one function as show in the example above.
 
-For error reporting to the runtime APIs the library defines the `RuntimeApiError` trait and the `ErrorResponse` object. Custom errors for the APIs should implement the `to_response() -> ErrorResponse` method of the `RuntimeApiError` trait.
+It also exposes the `Handler` trait. A type that conforms to this trait can be passed to the `lambda::run` function, which launches and runs the Lambda runtime.
 
-## lambda-runtime
-
-This library makes it easy to create Rust executables for AWS lambda. The library defines a `lambda!()` macro. Call the `lambda!()` macro from your main method with an implementation the `Handler` type:
-
-```rust
-pub trait Handler<E, O> {
-    /// Run the handler.
-    fn run(
-        &mut self,
-        event: E,
-        ctx: Context
-    ) -> Result<O, HandlerError>;
-}
-```
-
-`Handler` provides a default implementation that enables you to provide a Rust closure or function pointer to the `lambda!()` macro.
-
-Optionally, you can pass your own instance of Tokio runtime to the `lambda!()` macro: 
-```
-let rt = tokio::runtime::Runtime::new()?;
-lambda!(my_handler, rt);
-```
+The helper function `handler_fn` provides a default implementation that enables you to provide a Rust closure or function pointer to the `lambda::run` function.
 
 ## AWS event objects
 
@@ -206,10 +159,9 @@ To serialize and deserialize events and responses, we suggest using the use the 
 
 ```rust
 extern crate serde;
-extern crate serde_derive;
 extern crate serde_json;
 
-use serde_derive::{Serialize, Deserialize};
+use serde::{Serialize, Deserialize};
 use serde_json::json;
 use std::error::Error;
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 This package makes it easy to run AWS Lambda Functions written in Rust. This workspace includes multiple crates:
 
-- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that makes it easy to write Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that provides a Lambda runtime for applications written in Rust.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -167,34 +141,13 @@ $ unzip -o \
 # Ctrl-D to yield control back to your function
 ```
 
-## lambda-runtime-client
+## `lambda`
 
-Defines the `RuntimeClient` trait and provides its `HttpRuntimeClient` implementation. The client fetches events and returns output as `Vec<u8>`.
+This library makes it easy to create Rust executables for AWS lambda. The library defines exposes an `#[lambda]` attribute macro. Adding the `#[lambda]` attribute to your `main` function allows you to define your event handler logic in one function as show in the example above.
```
rimutaka, line 146: as show -> as shown

README.md
```diff
@@ -2,58 +2,32 @@
 
 This package makes it easy to run AWS Lambda Functions written in Rust. This workspace includes multiple crates:
 
-- [![Docs](https://docs.rs/lambda_runtime_client/badge.svg)](https://docs.rs/lambda_runtime_client) **`lambda-runtime-client`** is a client SDK for the Lambda Runtime APIs. You probably don't need to use this crate directly!
-- [![Docs](https://docs.rs/lambda_runtime/badge.svg)](https://docs.rs/lambda_runtime) **`lambda-runtime`** is a library that makes it easy to write Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that makes it easy to write Lambda functions in Rust.
```
rimutaka, line 5: Should we say somewhere that this is actually the runtime that runs the handler function?

It may be good say that the runtime and the handler function compile into a single executable called bootstrap that is loaded by Lambda and then runs in a loop waiting for an event.

All this may be obvious to those working on the runtime, but an outsider has no idea how it all works inside AWS. Other runtimes like Node or Java only deal with handler functions, so it's likely to be a step up in complexity for anyone landing here for the first time.

README.md
```diff
@@ -2,58 +2,32 @@
...
-    })
+#[lambda]
+#[tokio::main]
+async fn main(event: Value, _: Context) -> Result<Value, Error> {
+    Ok(event)
 }
 ```
 
-The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/basic.rs) in the `lambda-runtime` crate.
+The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello.rs) in the `lambda` crate. An [alternative example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello-without-macros.rs) that uses functions instead of an attribute macro to create the handler is available as well.
```
rimutaka, line 26: The links to examples may change after PR #244 is completed.

README.md
```diff
@@ -206,10 +159,9 @@ To serialize and deserialize events and responses, we suggest using the use the
 
 ```rust
 extern crate serde;
-extern crate serde_derive;
 extern crate serde_json;
```
Veetaha, line 162: I recommend removing `extern crate` entries too

README.md
softprops: Suggestion that might add clarity

```
... provides a Lambda runtime for applications written in Rust.
```

README.md
```diff
@@ -2,58 +2,32 @@
...
 
-- [![Docs](https://docs.rs/lambda_runtime_client/badge.svg)](https://docs.rs/lambda_runtime_client) **`lambda-runtime-client`** is a client SDK for the Lambda Runtime APIs. You probably don't need to use this crate directly!
-- [![Docs](https://docs.rs/lambda_runtime/badge.svg)](https://docs.rs/lambda_runtime) **`lambda-runtime`** is a library that makes it easy to write Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that makes it easy to write Lambda functions in Rust.
 - [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library that makes it easy to write API Gateway proxy event focused Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda_attributes/badge.svg)](https://docs.rs/lambda_attributes) **`lambda-attributes`** holds an [attribute macro](https://doc.rust-lang.org/reference/procedural-macros.html#attribute-macros) that runs an `async main` function in the Lamda runtime. You probably don't need to use this crate directly, the macro is exposed by **`lambda`**!
```
softprops, line 7: Listing this might add more cognitive overhead than is useful for users. I don't think this crate is designed to be depended on directly.

README.md
voteblake: Thanks - will address. It's funny how many times you can read something and miss things like this.

README.md
voteblake: This is a great callout, and also addresses that this line just repeats what is said in the header above. It makes the fact that this is a provided runtime more explicit, which might also help people when they get to deployment. Will update.

README.md
voteblake: I left it for parity with the previous README that listed each member crate of the workspace. Completely open to removing it, I agree it seems like an implementation detail that is best addressed in the docs rather than the README.

README.md
voteblake: Likely need to update the phrasing at line 146 as well

Right now I have:

```suggestion
- [![Docs](https://docs.rs/lambda/badge.svg)](https://docs.rs/lambda) **`lambda`** is a library that provides a Lambda runtime for applications written in Rust.
```
from @softprops and 

> With the `lambda` library you can build your Rust application to implement an [AWS Lambda's custom runtime](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html)

which I was considering at Line 146.

README.md
rimutaka: Well, if there is a crate I'd be asking myself, do I need it or what?
Shall we clearly state that:
```
The macro is exposed by **`lambda`**. You do not need to use this crate directly.
```

README.md
```diff
@@ -2,58 +2,32 @@
...
 
 ## Example function
 
-The code below creates a simple function that receives an event with a `greeting` and `name` field and returns a `GreetingResponse` message for the given name and greeting. Notice: to run these examples, we require a minimum Rust version of 1.31.
+The code below creates a simple function that receives an event and echoes it back as a response. Notice: to run these examples, we require a minimum Rust version of 1.31.
```
davidbarsky, line 11: An MSRV of 1.31 is _not_ correct. Right now, it's _at least_ a 1.42_, but I'd for the time being, I'd like to make the policy "latest stable Rust" until this reaches 1.0.

README.md
```diff
@@ -2,58 +2,32 @@
...
+type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
 
-    Ok(CustomOutput {
-        message: format!("Hello, {}!", e.first_name),
-    })
+#[lambda]
+#[tokio::main]
+async fn main(event: Value, _: Context) -> Result<Value, Error> {
+    Ok(event)
 }
```
davidbarsky, line 23: I've been thinking about this, but I think we should cut the `#[lambda]` macro from the initial release. It's problematic when used with anything that expects to be initialized a single time (e.g., a logger) but a _warm_ invoke occurs. In the case of a warm invoke, this will cause the runtime to panic.

Let's prefer a more full example?

README.md
```diff
@@ -91,9 +65,9 @@ You can now test the function using the AWS CLI or the AWS Lambda console
 
 ```bash
 $ aws lambda invoke --function-name rustTest \
-  --payload '{"firstName": "world"}' \
+  --payload '{"Hello,": "world!"}' \
   output.json
-$ cat output.json  # Prints: {"message":"Hello, world!"}
+$ cat output.json  # Prints: {"Hello,": "world!"}
```
davidbarsky, line 70: I'm not sure I'm in favor of this chance. Can we prefer the original?

README.md
```diff
@@ -2,58 +2,32 @@
...
 }
 ```
 
-The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/basic.rs) in the `lambda-runtime` crate.
+The code above is the same as the [basic example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello.rs) in the `lambda` crate. An [alternative example](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-runtime/examples/hello-without-macros.rs) that uses functions instead of an attribute macro to create the handler is available as well.
 
 ### Deployment
 
-There are currently multiple ways of building this package: manually, and with the [Serverless framework](https://serverless.com/framework/).
+There are currently multiple ways of building this package: manually with the AWS CLI, and with the [Serverless framework](https://serverless.com/framework/).
```
davidbarsky, line 30: Let's maybe include the SAM CLI with [the Makefile approach](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/building-custom-runtimes.html) and (hopefully) https://github.com/awslabs/aws-lambda-builders/pull/174, once that is merged.

README.md
```diff
@@ -167,34 +141,13 @@ $ unzip -o \
...
-```
-
-`Handler` provides a default implementation that enables you to provide a Rust closure or function pointer to the `lambda!()` macro.
-
-Optionally, you can pass your own instance of Tokio runtime to the `lambda!()` macro: 
-```
-let rt = tokio::runtime::Runtime::new()?;
-lambda!(my_handler, rt);
-```
+The helper function `handler_fn` provides a default implementation that enables you to provide a Rust closure or function pointer to the `lambda::run` function.
```
davidbarsky, line 150: Two notes:

- Let's avoid the second person tense.
- Provide an example of using this?

```suggestion
The function `handler_fn` converts a rust function or closure to `Handler`, which can then be run by `lambda::run`.
```

README.md
```diff
@@ -167,34 +141,13 @@ $ unzip -o \
 # Ctrl-D to yield control back to your function
 ```
 
-## lambda-runtime-client
+## `lambda`
 
-Defines the `RuntimeClient` trait and provides its `HttpRuntimeClient` implementation. The client fetches events and returns output as `Vec<u8>`.
+This library makes it easy to create Rust executables for AWS lambda. The library defines an `#[lambda]` attribute macro. Adding the `#[lambda]` attribute to your `main` function allows you to define your event handler logic in one function as shown in the example above.
```
davidbarsky, line 146:
```suggestion
`lambda` is a library for authoring reliable and performant Rust-based AWS Lambda functions. At a high level, it provides a few major components:

- `Handler`, a trait that defines interactions between customer-authored code and this library.
- `lambda::run`, function that runs an `Handler`.
```

README.md
```diff
@@ -167,34 +141,13 @@ $ unzip -o \
...
 
-For error reporting to the runtime APIs the library defines the `RuntimeApiError` trait and the `ErrorResponse` object. Custom errors for the APIs should implement the `to_response() -> ErrorResponse` method of the `RuntimeApiError` trait.
+It also exposes the `Handler` trait. A type that conforms to this trait can be passed to the `lambda::run` function, which launches and runs the Lambda runtime.
```
davidbarsky, line 148: Let's remove this.

davidbarsky (dismissed):

README.md
voteblake: Interesting, I did not know this about the attribute macro. My primary reason for avoiding it has been more do with developer tooling, I found `rust-analyser` and `cargo build` output regarding errors to be much less useful when using it.

That type of error (panic on warm invocation) does sound like a reason to hold it back from the release. I can update the README to reflect that.

davidbarsky: > Interesting, I did not know this about the attribute macro. My primary reason for avoiding it has been more do with developer tooling, I found rust-analyser and cargo build output regarding errors to be much less useful when using it.

That's another compelling reason to not release and I completely forgot about it. I think the `#[lambda]` macro is one of those things that looks nice on a slide/screenshot, but is something most customers will quickly outgrow.

README.md
rimutaka: A lay user perspective: can the Handler part be described in terms consistent with other runtimes?

- `Handler`, a trait that defines interactions between customer-authored code (the handler function described in other runtimes) and this library (the runtime).

- `lambda::run`, function that runs an `Handler` and interacts with the rest of the AWS infrastructure to fetch new event data and return the response.

README.md
softprops: + 1 on stable rust as a contract

softprops: \+ 1 (because github markdown)

README.md
softprops: >  I did not know this about the attribute macro.

this might just be a matter of documentation. I mentioned this in the lambda http docs as a tradeoff to consider. I still think its useful for specific cases but it's also good to be mindful of the tradeoffs.

README.md
softprops: We'll see... :) I'll try to keep a some light pressure on that. Ultimately SAM is the most likely AWS-suggested approach to expect with SAM being the tool AWS pushes. Not having a SAM story is worse than not having some SAM story. I agree with the adding SAM + Makefile example being useful

s551228258327922884_p1702_i14_w300

Veetaha: I'd much more believe in terraform ;d, could add an example with that

rimutaka: > I'd much more believe in terraform ;d, could add an example with that
I prefer TF too, but if AWS prefers SAM then what most people will use anyway.
@Veetaha if you do make an example, can we put it in examples folder with a mention and a link to it from README to keep it concise?

README.md
voteblake: I've added the previous functionality back to the README and the `hello-without-macro` example that it references. It is working on and generating an untyped `json` value. Let me know if you'd like to add custom input and output events back to this headline example as well. Leaving the discussion on custom events to later in the documentation allows a gentler introduction in my opinion.

README.md
voteblake: How does:

> Notice: this crate is tested against latest stable Rust.

sound?

README.md
brainstorm: Perhaps putting a dissonant note on this thread, I'm using CDK for deployment of Rust lambdas these days XD

If people adopt SAM as a Rust-lambda "blueprint project", there I'll go too... My current DX fail with Rust lambdas is that I've been unable to find a lightweight and [properly documented way to debug Rust lambdas locally](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-using-debugging.html)... 

My ideal scenario: not having to deal with docker (i.e: lambci) and my VSCode session having **all the variable values, watchpoints and backtraces I would normally have for "regular local binaries"**, ideally with proper observability all the way down to FFI (bindgen) crates.

Am I dreaming or is this available today and I've missed it? :)

brainstorm: Essentially I would like something like this for Rust+SAM+lambdas: https://pawelgrzybek.com/attach-visual-studio-code-debugger-to-sam-serverless-application-model-local-endpoint/ ... does anybody has such a setup? @rimutaka ?

rimutaka: @brainstorm,  That debug-proxy works for me just fine and I use CLI/TF for deployment.

coltonweaver (approved):
After reviewing this PR, it seems all of the past feedback had been accounted for and things seem accurate to me. With that, I'm approving it.

PR #248, opened 2020-08-04
When constructing a Context from a HeaderMap, fill in the env_config field

khuey:
*Description of changes:* Currently this just produces a useless env_config.

### DIFF
--- a/lambda/src/lib.rs
+++ b/lambda/src/lib.rs
@@ -204,7 +204,8 @@ where
         let event = event?;
         let (parts, body) = event.into_parts();
 
-        let ctx: Context = Context::try_from(parts.headers)?;
+        let mut ctx: Context = Context::try_from(parts.headers)?;
+        ctx.env_config = Config::from_env()?;
         let body = hyper::body::to_bytes(body).await?;
         let body = serde_json::from_slice(&body)?;
 


rimutaka:
Good spotting!

We also missed `pub client_context: Option<ClientContext>` and `pub identity: Option<CognitoIdentity>` in the same `Context ` structure.
They are declared, but are never used. I'll need both of them in a few weeks, so if no one fixes it by then I'll have to :)

khuey:
Doesn't look like `ClientContext` or `CognitoIdentity` have any existing `from_the_relevant_thing` methods though, so I'd prefer just to merge this as is :)

khuey:
Ping?

davidbarsky:
Apologies for the delay, @khuey. This looks good to me, and sorry for missing this!

davidbarsky (approved):

lambda/src/types.rs
```diff
@@ -136,6 +136,7 @@ impl TryFrom<HeaderMap> for Context {
                 .to_str()
                 .expect("Invalid XRayTraceID sent by Lambda; this is a bug")
                 .to_owned(),
+            env_config: Config::from_env()?,
```
davidbarsky, line 139: Actually, I may have spoken too soon. What do you think about populating the `env_config` field _not_ in the `TryFrom<HeaderMap>` impl block but within the execution of the runtime? This would allow this HeaderMap to remain _mostly_ pure (error-handling aside) and make it easier to test and simulate.

davidbarsky (changes requested):

lambda/src/types.rs
khuey: Well the downside would be that every call site would have to do this. Currently there's only one so it's not a big deal. Do you anticipate more being added in the future?

davidbarsky: The only other places where I can imagine this being added is in `lambda_http` and a yet-to-be-implemented mock runtime.

khuey: Comment addressed.

davidbarsky (approved):
Thanks!

khuey:
Anybody going to merge this?

davidbarsky (approved):

Issue #252, opened 2020-08-11
How to write a test for query_string_parameters


johansmitsnl:
I want to write a test for example: https://github.com/awslabs/aws-lambda-rust-runtime/blob/a9de2fcb24030a00e402348aba3c368b717feb6d/lambda-http/examples/hello-http-without-macros.rs#L15-L23

But until now it does not work.

```rust
#[cfg(test)]
mod tests {
    use super::*;
    use lambda_http::http::StatusCode;
    use lambda_http::{http, Body};

    #[tokio::test]
    async fn main_handles() {

        let request = http::Request::builder()
            .uri("https://api.com/?first_name=John")
            .expect("failed to build request");

        let expected = json!({"first_name": "John"}).into_response();
        let response = func(request, Context::default())
            .await
            .expect("expected Ok(_) value")
            .into_response();
        assert_eq!(response.body(), expected.body());
        assert_eq!(response.status(), StatusCode::OK)
    }
}
```

How can I stub the request?

softprops:
You'll want to use [`with_query_string_parameters`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/a9de2fcb24030a00e402348aba3c368b717feb6d/lambda-http/src/ext.rs#L186) should should be accessible in a test config with you import `lambda_http::RequestExt`.

mbrubeck:
`with_query_string_parameters` is not accessible in tests in downstream packages.  It is marked `#[cfg(test)]` so it is included only when building the unit test binary for `lambda_http` itself.  When building tests for downstream crates, `lambda_http` is built as a regular library, without `--cfg test`.

To expose this to downstream crates, it should be compiled unconditionally or behind a Cargo feature.

softprops:
Good to know. If I don't get to it first, we'd be open to a pull request to add that feature flag.

The original intent was to get a small surface area and accomidate use cases that are test specific with a test cfg.

johansmitsnl:
@softprops what would be the code to use it if that flag was enabled?
I run it from github master at the moment.

softprops:
> what would be the code to use it if that flag was enabled?

The target usage is unit tests in consumers crates

johansmitsnl:
I'm relatively new to rust, would it be possible to alter my code above?

johansmitsnl:
This works: #253 in my case. @paulvt helped me finding this change.

If changes are needed please let me know.

miere:
[...]

PR #253, opened 2020-08-11
Allow mock methods to be used in external apps

johansmitsnl:
Make mock api available

*Issue #, if available:* #252

*Description of changes:*
Make stub methods available for testing the API

### DIFF
--- a/lambda-http/src/ext.rs
+++ b/lambda-http/src/ext.rs
@@ -121,7 +121,6 @@ pub trait RequestExt {
     /// Configures instance with query string parameters under #[cfg(test)] configurations
     ///
     /// This is intended for use in mock testing contexts.
-    #[cfg(test)]
     fn with_query_string_parameters<Q>(self, parameters: Q) -> Self
     where
         Q: Into<StrMap>;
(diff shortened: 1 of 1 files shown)


lambda-http/src/ext.rs
```diff
@@ -121,7 +121,6 @@ pub trait RequestExt {
     /// Configures instance with query string parameters under #[cfg(test)] configurations
     ///
     /// This is intended for use in mock testing contexts.
-    #[cfg(test)]
```
softprops, line 124: Are there practical use cases where you'd want to override data coming from lambda triggers outside of a test cases?

johansmitsnl: I guess not. But you can't use these methods if from a crate if they are marked with test. And these parameters you do want to test them in your function.

softprops: Since this functions have no practical purpose out side of test scenarios would it be possible to do with a a feature flag that you could enable in your tests instead?

schell:
:+1: I just ran into this as well. Would be nice to get this through!

davidbarsky (approved):
Thank you!

PR #262, opened 2020-10-15
Allow mocking lambda_http::Request extensions outside of crate

l3ku:
*Description of changes:*

I have a use case, where I need to create a mock `lambda_http::Request` in a test function, and be able to pass a path parameter as part of the request. When the `lambda_http` crate parses a request from `LambdaRequest` to a `lambda_http::Request`, it appends the path parameters using `extension(PathParameters(path_parameters))`. However, the `PathParameters` can not be created outside of the crate, since they are not declared as public.  I also added the public scope to the other extensions, since this use pattern also applies to them.

Is there some better way to do this without having to declare the structs as public?

### DIFF
--- a/lambda-http/src/ext.rs
+++ b/lambda-http/src/ext.rs
@@ -6,18 +6,18 @@ use std::{error::Error, fmt};
 use crate::{request::RequestContext, strmap::StrMap, Body};
 
 /// ALB/API gateway pre-parsed http query string parameters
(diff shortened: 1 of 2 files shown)


l3ku:
Sorry, I hadn't noticed the newest change in master https://github.com/awslabs/aws-lambda-rust-runtime/commit/c36409c5e65f994c7ff48510cd111905b4aa77c9 which makes my use case possible. Closing as unnecessary.

Issue #267, opened 2020-11-23
Lambda-http: setting multiple cookies using multiValueHeaders is not supported in API Gateway version 2.0


l3ku:
From https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html:

> Format 2.0 doesn't have multiValueHeaders or multiValueQueryStringParameters fields. Duplicate headers are combined with commas and included in the headers field. Duplicate query strings are combined with commas and included in the queryStringParameters field.
> 
> Format 2.0 includes a new cookies field. All cookie headers in the request are combined with commas and added to the cookies field. In the response to the client, each cookie becomes a set-cookie header.

I have a use case, where I need to pass several `set-cookie` headers in the response from Lambda to API Gateway 2.0 using the `lambda-http` crate. However, when I set multiple `SET_COOKIE` headers on the response, only one of them is returned from API Gateway, presumably since API Gateway only handles the one header which is present in the `headers` field of `LambdaResponse`.

Is there a way to work around this other than adding the `cookies` field as part of the `LambdaResponse` in some manner, like using a similar enum as `LambdaRequest` for different API Gateway versions?

rimutaka:
@miere, I'm replying to your question here to keep the PR discussion clean. We really need someone to enable the new Discussions feature.

> An honest question: why is this runtime maintaining its own implementation of the AWS Events where one could get them from the aws_lambda_event crate?

If I recall this correctly, there was a conversation of AWS taking over the maintenance of their events crate and then it would be added as a dependency. I cannot find that thread, but I think @softprops was looking into it.

softprops:
I’m not actively working on an open issue atm but just a reminder about the lambda/http crate. 

The idea was to transparently use the rust community http crate as an interface for various flavors of aws http like triggers. The underlying ser/deser impl for for those has many optimizations for that deser fields directly to the http crate types when possible for example. The idea was this would be valuable for those coming from rust and new to lambda having familiar ecosystem types to work with. 

a generic events crate would likely have a more unopinionated set of fields that directly map to aws trigger event types. I believe you could use the events crate and default lambda runtime for this for this today.

The trade off is your lose the flexibility to transparently switch between http triggers without changing you application and lose the leverage of the ecosystem built up around the rust http crate and it’s types.

[...]

PR #269, opened 2020-11-24
Lambda-http: vary type of response based on request origin

l3ku:
*Issue #, if available:*
#267 

*Description of changes:*

ApiGatewayV2, ApiGateway and Alb all expect different types of responses to be returned from the invoked lambda function. Thus, it makes sense to pass the request origin to the creation of the response, so that the correct type of LambdaResponse is returned from the function.

This PR also adds support for the "cookies" attribute which can be used for returning multiple Set-cookie headers from a lambda invoked via ApiGatewayV2, since ApiGatewayV2 no longer seems to recognize the "multiValueHeaders" attribute.

On lines 129-132 of `response.rs` the `SET_COOKIE` headers are removed from the `headers` returned to API Gateway v2, because otherwise API Gateway v2 would return three `set-cookie` headers when just returning two cookies in `cookies`, because one of the cookies would be duplicated in the `headers`. I tried using the `http::header::OccupiedEntry::remove_entry_mult` function for removing the `SET_COOKIE` headers from the `headers`, but I got some panicking from the http crate as described in https://github.com/hyperium/http/issues/446. So I went with the solution where I first fetch the `SET_COOKIE` headers returned with `headers.get_all` and insert them cloned into the `cookies`, and then remove them with `headers.remove(SET_COOKIE)`. However, any ideas and suggestions as to improve this are warmly welcome!


lambda-http/src/lib.rs
```diff
@@ -149,7 +152,7 @@ where
 
 #[doc(hidden)]
 pub struct TransformResponse<R, E> {
-    is_alb: bool,
+    request_origin: RequestOrigin,
```
softprops, line 155: Much cleaner

softprops (approved):
This is excellent.

At first there was only rest apis ... and then albs.. and then http apis.

When we got to supporting three types the current impl didn't feel right but this feels a lot cleaner and provides an way forward when the inevitable 4th http like trigger comes along

softprops (approved):

calavera (approved):
It'd be awesome if this got merged into master @davidbarsky.

miere:
An honest question: why is this runtime maintaining its own implementation of the AWS Events where one could get them from the `aws_lambda_event` crate?

bahildebrand (approved):

Issue #274, opened 2020-12-16
Netlify's Open Lambda Rust Runtime Distribution


calavera:
👋 Hi all,

A few weeks ago, I started to cherry pick open PRs in my own fork. I decided to publish my fork in crates.io with a different name so people can use it without having to maintain their own forks forever. We're at a point where we've fixed some major compatibility bugs, and we're using it internally for some fun new projects.

You can find the fork in https://github.com/netlify/aws-lambda-rust-runtime

These are the PRs that we've cherry picked from this repository (besides fixing some other issues):

https://github.com/awslabs/aws-lambda-rust-runtime/pull/242 by @rimutaka 
https://github.com/awslabs/aws-lambda-rust-runtime/pull/247 by @voteblake
https://github.com/awslabs/aws-lambda-rust-runtime/pull/251 and #263 by @stanislav-tkach 
https://github.com/awslabs/aws-lambda-rust-runtime/pull/255 by @nmoutschen
#265 by @RyanJarv
#266 by @l3ku

These are the crates:

https://crates.io/crates/netlify_lambda_http
https://crates.io/crates/netlify_lambda
https://crates.io/crates/netlify_lambda_attributes

And these are the docs:

https://rs-lambda-runtime.netlify.engineering/netlify_lambda_http/
https://rs-lambda-runtime.netlify.engineering/netlify_lambda/
https://rs-lambda-runtime.netlify.engineering/netlify_lambda_attributes/

Feel free to use it, reach out to me if you have comments, or questions. If there is any issue in this repo that you think it's a big blocker for your use of the Rust Runtime on AWS Lambda, and you think I can help unblock it, feel free to ping me.

rimutaka:
This is actually very sad that we came to this. Netlify should just join as maintainers. @davidbarsky, et all, if this project is no longer a priority for you, please, allow the community to take it forward.

nmoutschen:
Hey @calavera & folks!

Thanks a lot for this and agree with @rimutaka that it's sad we came to this. I'm still a beginner in Rust, but I'm part of the serverless team at AWS and I'd like to see what we could do on that front.

I'll be on vacation starting tomorrow till the 11th of January, but if you could send me a DM on https://twitter.com/NMoutschen and I'll share my email address so we can have a chat about this.

Veetaha:
[...]

PR #276, opened 2021-01-06
Upgrade to tokio 1.0

coolreader18:
*Description of changes:* Change to depend on tokio 1.0 instead of tokio 0.2.


simnalamburt:
For the people looking for Tokio 1.0 support of aws-lambda-rust-runtime: Netlify's fork already has Tokio 1.0 support. FYI

###### References
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/274
- https://github.com/netlify/aws-lambda-rust-runtime/pull/8

johansmitsnl:
@coolreader18 I can confirm that it is working :+1: solved many hours debugging why it was not working with `0.46` of `rusoto`.

Hoping to see this merged soon.

lambda/src/simulated.rs
```diff
@@ -123,12 +122,25 @@ impl BufferState {
...
+            (&mut buf_l[len_l - bytes_from_l..], buf_r)
+        };
+
+        buf_r.reverse();
+        to_buf.put_slice(buf_r);
+        buf_l.reverse();
+        to_buf.put_slice(buf_l);
+
+        let remaining_buffer_size = self.buffer.len() - bytes_to_read;
+        self.buffer.truncate(remaining_buffer_size);
```
coltonweaver, line 143: I'm conflicted on how I feel about this change specifically. It seems like quite a bit of extra code to achieve necessarily the same thing, unless I'm totally mistaken.

Just to confirm my understanding - you split the buffer into two slices left and right, check if `bytes_to_read` is less than or equal to the length of the right buffer, at which point you take what you need from the right buffer to place into the `to_buf`. If `bytes_to_read` is greater than the right, then you take what you need from the left and return both all of the right buffer and what you've taken from the left. And then finally you truncate the buffer before returning.

It just seems like a lot of extra code that makes this somewhat difficult to follow. Is it necessary? If so, is there a way we could simplify this or at least comment a bit on the intention here?

coltonweaver (changes requested):
Overall this is a great contribution and we are very appreciative for it! Looking forward to hearing back.

lambda/src/simulated.rs
coolreader18: Yeah, so `self.buffer` is a VecDeque, i.e. not contiguous. However, that doesn't mean that you can't get contiguous slices from it, and the repeated `pop_back().unwrap()` seemed inefficient, especially when the equivalent for `tokio::io::ReadBuf` would be repeated `.put_slice(&[i])`. I can definitely add comments to the code, but it's essentially taking `bytes_to_read` bytes from the end of the buffer, than feeding them backwards into the read buf. The truncation is to formally "take" the slice we've read out of the buffer. It's more complicated in code than it might sound since there's 2 slices that make up the VecDeque, but that's all it's doing.

coolreader18: Though, I think this can be a lot simpler, since it doesn't need to be reversed at all if we append instead of prepend in `BufferState::write()`

lambda/src/simulated.rs
```diff
@@ -123,14 +120,22 @@ impl BufferState {
...
+            to_buf.put_slice(buf);
+        } else {
+            // otherwise, use up all of buf_l and take the remaining from buf_r
+            to_buf.put_slice(buf_l);
+            to_buf.put_slice(&buf_r[..bytes_to_read - buf_l.len()]);
+        };
 
-        bytes_to_read
+        // cut off what we've read from the start of the buffer
+        self.buffer.drain(..bytes_to_read);
```
TransientError, line 138: Thanks for your PR! We really want to get the new tokio update out as soon as we can. I'm a little skeptical about if this is really necessary though especially since `simulated.rs` is used for testing only. Do you think we can separate this out to a separate issue for more discussion?

coolreader18: Alright, I've removed that change from the PR.

coltonweaver (approved):

coltonweaver:
To me, everything looks good. I played around with this on my side and things seem to be functioning as expected as well. Going to ask the others to take a look just for posterity, but hopefully we can get this merged in soon.

lambda/src/simulated.rs
coltonweaver: Thanks for addressing these. I'll take a look at the PR you split off from this soon.

LucioFranco (approved):

lambda/src/simulated.rs
davidbarsky: Since Tokio now has https://docs.rs/tokio/1.1.0/tokio/io/struct.DuplexStream.html, I think this code can be removed wholesale—I wrote this code initially to mimic this. I think it's largely compatible?

lambda/src/simulated.rs
```diff
@@ -89,7 +88,7 @@ impl AsyncWrite for SimStream {
 
 /// Delegates to the underlying `read` member's methods
 impl AsyncRead for SimStream {
-    fn poll_read(mut self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &mut [u8]) -> Poll<IoResult<usize>> {
+    fn poll_read(mut self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &mut ReadBuf<'_>) -> Poll<IoResult<()>> {
         Pin::new(&mut self.read).poll_read(cx, buf)
     }
 }
```
davidbarsky, line 94: I mentioned this in a comment elsewhere, but I think that SimStream can be replaced with [`tokio::io::Duplex`](https://docs.rs/tokio/1.1.0/tokio/io/struct.DuplexStream.html). This isn't a blocker to merging, however, but it _would_ reduce how much code needs to be in this crate.

(I think some usages in tests would require a `std::sync::Arc`, but I'm not sure.)

davidbarsky (approved):
Approved with a few nits.

coltonweaver:
@coolreader18 Thoughts on the comments left by @davidbarsky? Don't worry about removing the SimStream in this PR, I'm mainly talking about changing the dependency changes slightly. I think once that's done I'll merge this PR in.

lambda/Cargo.toml
```diff
@@ -12,15 +12,14 @@ simulated = []
 derive = ["lambda-attributes"]
 
 [dependencies]
-tokio = { version = "0.2.4", features = ["full"] }
-hyper = "0.13"
+tokio = { version = "1.0", features = ["macros", "io-util", "sync", "rt-multi-thread"] }
+hyper = { version = "0.14", features = ["full"] }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1.0.39"
-tower-service = "0.3"
```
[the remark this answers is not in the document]
coolreader18: I feel like it makes sense to use the re-export from hyper - it is part of the public API, and I think it makes it more clear that hyper uses that trait as well, rather than us integrating with something else from the tower stack. I dunno either :shrug:

coltonweaver:
Alright, we're merged in. Thanks for the PR and addressing the feedback so quickly!

PR #282, opened 2021-01-29
Remove lambda attributes

coltonweaver:
*Issue #, if available:*

N/A

*Description of changes:*

This PR removes the `#[lambda]` macro from this library. @davidbarsky provided the following valid reasons for why this should be done:

* Compiler error reporting is not handled correctly using this macro, so the spans where an error occurs are undefined and point to incorrect locations.
* Using this macro conflates warm and cold starts for Lambda functions, as this removes the ability to initialize global resources that might panic if initialized twice, such as a global logger.

Given that the alternative to using this macro is a single line addition to a `main` function, there is no real need to continue supporting this macro.

- [*] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [*] I confirm that I've made a best effort attempt to update all relevant documentation.


bahildebrand (approved):
LGTM

davidbarsky (approved):

PR #285, opened 2021-02-05
point to basic.rs instead of hello

richardhboyd:
using the parameter `hello` results in the following error.

```text
error: no example target named `hello`
```

changing it to basic *appears* to work in that it doesn't throw an error, but this is only my second ever time running `cargo build ...`

*Issue #, if available:*

*Description of changes:*


bahildebrand (approved):
Nice catch!

PR #288, opened 2021-02-13
Replace SimStream with DuplexStream

TransientError:
*Issue #, if available:* 281

*Description of changes:*
Using the tokio DuplexStream instead of SimStream.

Because DuplexStream doesn't implement Connection, I created a wrapper for it that does, but I'm open to some alternatives.

- [ x] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [ x] I confirm that I've made a best effort attempt to update all relevant documentation.

### DIFF
--- a/lambda/src/simulated.rs
+++ b/lambda/src/simulated.rs
@@ -1,229 +1,100 @@
 use http::Uri;
 use hyper::client::connect::Connection;
 use std::{
-    cmp::min,
-    collections::VecDeque,
+    collections::HashMap,
     future::Future,
     io::Result as IoResult,
     pin::Pin,
     sync::{Arc, Mutex},
-    task::{Context, Poll, Waker},
+    task::{Context, Poll},
 };
-use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+use tokio::io::{AsyncRead, AsyncWrite, DuplexStream, ReadBuf};
 
-/// Creates a pair of `AsyncRead`/`AsyncWrite` data streams, where the write end of each member of the pair
-/// is the read end of the other member of the pair.  This allows us to emulate the behavior of a TcpStream
-/// but in-memory, deterministically, and with full control over failure injection.
-pub(crate) fn chan() -> (SimStream, SimStream) {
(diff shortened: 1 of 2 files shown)


lambda/src/simulated.rs
```diff
@@ -1,229 +1,94 @@
...
-/// Delegates to the underlying `write` member's methods
-impl AsyncWrite for SimStream {
-    fn poll_write(mut self: Pin<&mut Self>, cx: &mut Context<'_>, buf: &[u8]) -> Poll<IoResult<usize>> {
-        Pin::new(&mut self.write).poll_write(cx, buf)
-    }
-
-    fn poll_flush(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<IoResult<()>> {
-        Pin::new(&mut self.write).poll_flush(cx)
-    }
+pub struct DuplexStreamWrapper(pub DuplexStream);
```
davidbarsky, line 20: Let's not make this field public as public fields are potential forwards compatibility hazards.

```suggestion
pub struct DuplexStreamWrapper(DuplexStream);
```

lambda/src/simulated.rs
```diff
@@ -1,229 +1,94 @@
...
-
-    fn poll_ready(&mut self, _: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
-        Poll::Ready(Ok(()))
-    }
-
-    fn call(&mut self, _: Uri) -> Self::Future {
-        let inner = self.inner.clone();
-        Box::pin(async move { Ok(inner) })
-    }
+    pub inner: Arc<Mutex<HashMap<Uri, DuplexStreamWrapper>>>,
```
davidbarsky, line 17: Same thing—I'd rather make this field private, I worry that that this is a forwards compatibility hazard.
```suggestion
    inner: Arc<Mutex<HashMap<Uri, DuplexStreamWrapper>>>,
```

davidbarsky (changes requested):
This looks like a good change overall! Thanks for doing it. I just had two comments on field visibility.

davidbarsky (approved):
Thanks so much!

Issue #297, opened 2021-03-11
Improve error handling usability


miere:
As discussed at https://github.com/awslabs/aws-lambda-rust-runtime/issues/294, there's room for usability/design improvement on how internal errors are handled.

I believe it would be beneficial if we remove the requirement to rewrite our own [Error alias](https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/examples/error-handling.rs#L10) for each lambda endpoint. By simply making the [existing one](https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/src/lib.rs#L34) public would solve this problem. It would also allow the maintainers to create their own backward compatible Error type, with little to (possibly) zero impact on users of this library.

rimutaka:
@miere , I agree, but are you sure it will work? There was a discussion re this some time ago, if I recall it correctly. @softprops may remember the reason why it was done that way.

softprops:
I was a fan of the type alias for the record :)

The pros outweigh the cons for me.

Pros

* concise while still flexible type api
* easier migration path for users if the type changes. Most won't need to change their code

Cons

* widens surface area (imports needed from this crate be ate lib)


There have been a few discussions in the past about using an external crate specific type. I try to stay clear for this because the rust ecosystem tends to change its taste for error crates every few years (error-chain, failure, anyhow, etc) which tends to become a maintenance burden for crates exporting those types. This end up breaking your api contracts by migrating your crate's error to the current trend. I do not enjoy this kind of work. I like the std library because it's by definition standard! 

A type alias over std lib types seems like a happy medium for me

coltonweaver:
I tried my hand at the suggestion in #298. Let me know if you have any feedback!

coltonweaver:
I've merged that change in now. Gonna close this issue.

PR #298, opened 2021-03-12
Making Error type public so that consumers don't have to create their own type alias

coltonweaver:
*Issue #, if available:* #297

*Description of changes:* The `Error` type alias was formerly public only to the lambda_runtime crate. Made this public and replace customer type aliases elsewhere.

### DIFF
--- a/lambda-runtime/examples/basic.rs
+++ b/lambda-runtime/examples/basic.rs
@@ -35,7 +35,7 @@ async fn main() -> Result<(), lambda_runtime::Error> {
     Ok(())
 }
 
-pub(crate) async fn my_handler(event: Request, ctx: Context) -> Result<Response, Error> {
+pub(crate) async fn my_handler(event: Request, ctx: Context) -> Result<Response, lambda_runtime::Error> {
     // extract some useful info from the request
     let command = event.command;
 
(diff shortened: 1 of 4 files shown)


bahildebrand:
You beat me to the punch by minutes!

coltonweaver:
@bahildebrand My bad I should have marked this one as assigned!

lambda-runtime/examples/basic.rs
```diff
@@ -29,7 +25,7 @@ struct Response {
 }
 
 #[tokio::main]
-async fn main() -> Result<(), Error> {
+async fn main() -> Result<(), lambda_runtime::Error> {
```
bahildebrand, line 28: Can we include the error type instead? Same for the other examples as well.

bahildebrand (commented):
LGTM other than the small change.

bahildebrand:
> 
> 
> @bahildebrand My bad I should have marked this one as assigned!

All good haha! It wasn't a large change, so not much time was spent on it.

bahildebrand (approved):
Sorry I meant to approve on the last one. My comment is a nit. Up to you if you want to address it.

rimutaka (approved):
Looks good and timely for v.0.3

lambda-runtime/examples/basic.rs
coltonweaver: Ah yeah I totally agree, I think it would look better as an include. Let me address that and push the change up.

coltonweaver:
> 
> 
> Sorry I meant to approve on the last one. My comment is a nit. Up to you if you want to address it.

I agree with your comment! Just pushed a commit up with that change.

Issue #287, opened 2021-02-12
Unable to use third party async libraries due to Sync requirements


vultix:
Pull Request #284  Introduced the `Sync` requirement for all http lambdas.  

This broke any http lambdas using any of these libraries:
- Rusoto
- Mongodb
- Sqlx

### Reproduction
```rust
use lambda_http::{
    handler,
    lambda::{self, Context},
    IntoResponse, Request, RequestExt, Response,
};

use serde_json::{json, Value};
use rusoto_core::Region;
use rusoto_dynamodb::{DynamoDb, DynamoDbClient, ListTablesInput};

type Error = Box<dyn std::error::Error + Send + Sync + 'static>;

#[tokio::main]
async fn main() -> Result<(), Error> {
    lambda::run(handler(func)).await?;
    Ok(())
}

async fn func(_: Request, _: Context) -> Result<Value, Error> {
    let client = DynamoDbClient::new(Region::UsEast1);

    let output = client.list_tables(Default::default()).await?;

    dbg!{output};

    Ok(json!({"success": true}))
}
```

```toml
[dependencies]
serde_json = "1.0"
serde = "1.0"
tokio = {version = "1.0", features = ["macros"]}

lambda_http = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", branch = "master"}
lambda = { git = "https://github.com/awslabs/aws-lambda-rust-runtime/", branch = "master"}
rusoto_dynamodb = "0.46.0"
rusoto_core = "0.46.0"
```
This is the error I get:
```
error: future cannot be shared between threads safely
   --> src/main.rs:16:17
    |
16  |     lambda::run(handler(func)).await?;
    |                 ^^^^^^^ future returned by `func` is not `Sync`
    | 
   ::: ~/.cargo/git/checkouts/aws-lambda-rust-runtime-7c865cce90132439/9bd45c1/lambda-http/src/lib.rs:109:19
    |
109 | pub fn handler<H: Handler>(handler: H) -> Adapter<H> {
    |                   ------- required by this bound in `handler`
    |
    = help: the trait `Sync` is not implemented for `dyn Future<Output = std::result::Result<ListTablesOutput, RusotoError<ListTablesError>>> + Send`
note: future is not `Sync` as it awaits another future which is not `Sync`
   --> src/main.rs:23:18
    |
23  |     let output = client.list_tables(Default::default()).await?;
[...]

PR #309, opened 2021-03-26
RFC: Support borrowed deserializaton

duarten:
*Description of changes:*

This PR allows the request passed to a handler to be deserialized with a lifetime `'de`, thus avoiding copies when deserializing. Due to the lack of higher kinded types in Rust, the implementation relies on GATs. This makes things ugly on the Lambda side. For example, calling `run()` for `DeserializeOwned` types now becomes:

```rust
lambda_runtime::run::<Request, _, _>(func).await?;
```

For borrowed deserialized types, this becomes:

```rust
struct BorrowedRequest {}
impl<'de> Deserializable<'de> for BorrowedRequest {
    type Deserialize = Request<'de>;
}
lambda_runtime::run::<BorrowedRequest, _, _>(func).await?;
```

This ugliness can be fixed by macros, like:

```rust
handler!(func).await?;
```

If the PR looks good, I can prepare a follow-up with the macro.

### DIFF
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -17,13 +17,13 @@
 //! your function's execution path.
 //!
 //! ```rust,no_run
-//! use lambda_http::{handler, lambda_runtime::{self, Error}};
+//! use lambda_http::{handler, lambda_runtime::Error};
 //!
 //! #[tokio::main]
 //! async fn main() -> Result<(), Error> {
 //!     // initialize dependencies once here for the lifetime of your
 //!     // lambda task
-//!     lambda_runtime::run(handler(|request, context| async { Ok("👋 world!") })).await?;
+//!     lambda_http::run(handler(|request, context| async { Ok("👋 world!") })).await?;
 //!     Ok(())
 //! }
 //! ```
@@ -34,11 +34,11 @@
 //! with the [`RequestExt`](trait.RequestExt.html) trait.
 //!
 //! ```rust,no_run
-//! use lambda_http::{handler, lambda_runtime::{self, Context, Error}, IntoResponse, Request, RequestExt};
+//! use lambda_http::{handler, lambda_runtime::{Context, Error}, IntoResponse, Request, RequestExt};
 //!
 //! #[tokio::main]
 //! async fn main() -> Result<(), Error> {
-//!     lambda_runtime::run(handler(hello)).await?;
+//!     lambda_http::run(handler(hello)).await?;
 //!     Ok(())
 //! }
 //!
@@ -183,7 +183,7 @@ where
 ///
 /// # Example
 /// ```no_run
-/// use lambda_http::{run, Context};
+/// use lambda_http::{handler, Context, IntoResponse, Request, RequestExt};
 /// use serde_json::Value;
 ///
 /// type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -62,7 +62,7 @@
 extern crate maplit;
 
 pub use http::{self, Response};
-pub use lambda_runtime::{self, Context};
+pub use lambda_runtime::{self, Context, Deserializable};
 use lambda_runtime::{Error, Handler as LambdaHandler};
 
 mod body;
@@ -76,6 +76,7 @@ use crate::{
     response::LambdaResponse,
 };
 use std::{
+    fmt,
     future::Future,
     pin::Pin,
     task::{Context as TaskContext, Poll},
@@ -89,9 +90,9 @@ pub type Request = http::Request<Body>;
 /// This can be viewed as a `lambda_runtime::Handler` constrained to `http` crate `Request` and `Response` types
 pub trait Handler: Sized {
     /// The type of Error that this Handler will return
-    type Error;
+    type Error: fmt::Display + Send + Sync;
     /// The type of Response this Handler will return
-    type Response: IntoResponse;
+    type Response: IntoResponse + Send + Sync;
     /// The type of Future this Handler will return
     type Fut: Future<Output = Result<Self::Response, Self::Error>> + Send + 'static;
     /// Function used to execute handler behavior
@@ -107,7 +108,7 @@ pub fn handler<H: Handler>(handler: H) -> Adapter<H> {
 impl<F, R, Fut> Handler for F
 where
     F: Fn(Request, Context) -> Fut,
-    R: IntoResponse,
+    R: IntoResponse + Send + Sync,
     Fut: Future<Output = Result<R, Error>> + Send + 'static,
 {
     type Response = R;
@@ -126,7 +127,8 @@ pub struct TransformResponse<R, E> {
 
 impl<R, E> Future for TransformResponse<R, E>
 where
-    R: IntoResponse,
+    R: IntoResponse + Send + Sync,
+    E: Send + Sync,
 {
     type Output = Result<LambdaResponse, E>;
     fn poll(mut self: Pin<&mut Self>, cx: &mut TaskContext) -> Poll<Self::Output> {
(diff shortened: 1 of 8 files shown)


duarten:
Tested with rustc 1.51.0 (2fd73fabe 2021-03-23).

lambda-runtime/src/lib.rs
```diff
@@ -152,39 +162,54 @@ where
...
-            let req = match task.await {
-                Ok(response) => match response.await {
-                    Ok(response) => {
-                        trace!("Ok response from handler (run loop)");
-                        EventCompletionRequest {
-                            request_id,
-                            body: response,
+            let request_id = ctx.request_id.clone();
+            let request_id_for_panic = ctx.request_id.clone();
+            let task = tokio::spawn(async move {
```
rimutaka, line 169: I found this part a bit confusing. It was a bit easier to follow after I renamed a few variables and added comments.

Here is my edit for lines  169 - 222:
```rust
            let handler_call = tokio::spawn(async move {
                let event_body = serde_json::from_slice(&body);
                // call the handler function if `event_body` was deserialized successfully or log an error
                // in either case return `Result<Request<Body>, Error>`
                match event_body {
                    Ok(event_body) => match handler.call(event_body, ctx).await {
                        Ok(response) => {
                            // handler call success
                            trace!("Ok response from handler (run loop)");
                            EventCompletionRequest {
                                request_id: &request_id,
                                body: response,
                            }
                            .into_req()
                        }
                        Err(err) => {
                            // handler call error
                            error!("{}", err); // logs the error in CloudWatch
                            EventErrorRequest {
                                request_id: &request_id,
                                diagnostic: Diagnostic {
                                    error_type: type_name_of_val(&err).to_owned(),
                                    error_message: format!("{}", err), // returns the error to the caller via Lambda API
                                },
                            }
                            .into_req()
                        }
                    },
                    Err(err) => {
                        // `event_body` deserialization error
                        error!("{}", err); // logs the error in CloudWatch
                        EventErrorRequest {
                            request_id: &request_id,
                            diagnostic: Diagnostic {
                                error_type: type_name_of_val(&err).to_owned(),
                                error_message: format!("{}", err), // returns the error to the caller via Lambda API
                            },
                        }
                        .into_req()
                    }
                }
            });
            // execute handler call from the task above
            // the error handling is done inside the `handler_call` task,
            // so the only way it may fail is if it panics 
            let handler_response = match handler_call.await {
                Ok(req) => req,
                Err(err) if err.is_panic() => {
                    error!("{:?}", err); // inconsistent with other log record formats - to be reviewed
                    EventErrorRequest {
                        request_id: &request_id_for_panic,
                        diagnostic: Diagnostic {
                            error_type: type_name_of_val(&err).to_owned(),
                            error_message: format!("Lambda panicked: {}", err),
                        },
                    }
                    .into_req()
                }
                Err(_) => unreachable!("tokio::task should not be canceled"),
            }?;
            // send the handler function response back to AWS API
            client
                .call(handler_response)
                .await
                .expect("Unable to send response to Runtime APIs");
        }
```

lambda-runtime/examples/basic.rs
```diff
@@ -31,7 +31,7 @@ async fn main() -> Result<(), Error> {
     SimpleLogger::new().with_level(LevelFilter::Info).init().unwrap();
 
     let func = handler_fn(my_handler);
-    lambda_runtime::run(func).await?;
+    lambda_runtime::run::<Request, _, _>(func).await?;
```
rimutaka, line 34: A comment about `_,_>` here would be good.

To be honest, I didn't understand how this change works and why we need `lambda_runtime::run::<Value, _, _>(func).await?;`, but that's just me not know how GATs and SERDE well enough. What I think would be good is to explain what the `_,_>` part means, why need it and if it should be copied as-is or there can be something in place of `_`. Quite often `_` is used in examples for brevity, which raises the question, what do I put in there in my case? Do I go with `_` or do I have to provide some value?

A macro will definitely help resolve the confusion for dummies like me.

lambda-runtime/src/client.rs
```diff
@@ -272,11 +281,7 @@ mod endpoint_tests {
         async fn func(event: serde_json::Value, _: crate::Context) -> Result<serde_json::Value, Error> {
             Ok(event)
         }
-        let f = crate::handler_fn(func);
-
-        let client = &runtime.client;
-        let incoming = incoming(client).take(1);
-        runtime.run(incoming, f).await?;
+        do_run(runtime, crate::handler_fn(func)).await?;
```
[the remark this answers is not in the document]
duarten: That's because of https://github.com/rust-lang/rust/issues/44721, we currently can't mix `impl Trait` and explicit generic arguments.

lambda-runtime/src/lib.rs
duarten: I'll incorporate those changes asap.

lambda-runtime/examples/basic.rs
duarten: I think a final version of this PR will definitely include the macro - there's just a lot of incidental complexity without it. The `_`'s are currently used for brevity, because we are required to provide the first generic parameter. Because that first one is different from the actual handler argument for borrowed deserialize, there's no way for the compiler to infer it. That's not true for the handler itself and the return type, so we can just say `_`.

rimutaka: @duarten, thank you for explaining it to me. I think I get it now :)
I paraphrased your explanation as a comment:
```rust
// The 1st param (Request) has to be specified every time because the compiler cannot infer it.
// The 2nd and 3rd parameters can use their default types and be omitted (_).
```

davidbarsky (commented):
Thanks for opening this. While most of my comments were focused on "bit of this are a breaking change" and "this might need to be sealed", I also want to ask—perhaps unfairly—whether this is change is necessary _in the first place_. Given the architecture of the Lambda service itself, I'm not sure that zero-copy deserialization will help here _for a while_—I think there's significantly lower-hanging, performance-related, fruit available (such as using different allocators like jemalloc or mimalloc). I'm worried that this change does not carry its weight in complexity in return for the performance gains. Have you benchmarked these changes?

lambda-http/src/lib.rs
```diff
@@ -89,9 +90,9 @@ pub type Request = http::Request<Body>;
 /// This can be viewed as a `lambda_runtime::Handler` constrained to `http` crate `Request` and `Response` types
 pub trait Handler: Sized {
     /// The type of Error that this Handler will return
-    type Error;
+    type Error: fmt::Display + Send + Sync;
     /// The type of Response this Handler will return
-    type Response: IntoResponse;
+    type Response: IntoResponse + Send + Sync;
```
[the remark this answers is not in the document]
duarten: These bounds are lifted from `run()`, so in that sense they're not really breaking (any `Handler` would already need to satisfy those bounds for it to be useful).

duarten:
Regarding Sync: note this PR doesn't (intentionally) change anything, it merely lifts bounds to the trait. The error and response type (`B`) are both sync, which is required AFAIK, since the handler is called in a tokio task. The future is not required to be sync, which is what I think #287 refers to.

duarten:
> Thanks for opening this. While most of my comments were focused on "bit of this are a breaking change" and "this might need to be sealed", I also want to ask—perhaps unfairly—whether this is change is necessary _in the first place_. Given the architecture of the Lambda service itself, I'm not sure that zero-copy deserialization will help here _for a while_—I think there's significantly lower-hanging, performance-related, fruit available (such as using different allocators like jemalloc or mimalloc). I'm worried that this change does not carry its weight in complexity in return for the performance gains. Have you benchmarked these changes?

I'm sure there are lots of other performance gains to have, and I agree that without higher-kinded types the implementation is complex (maybe it can be simplified by someone with more rust experience). I haven't benchmarked these specific changes (although over time I've benchmarked my fair share of zero-copy algorithms), but I can write a quick benchmark if that will influence the merge/don't merge decision. For now we can assume there will always be gains, proportional to the payload size.

This being said, not making this change means that event definitions must be `DeserializeOwned`. When the deserialization gains begin to matter (e.g., due to fixing the other performance issues, or billing becoming per micro-second), then all consumers wanting to take advantage of the gains will incur the cost of making the change.

As a consumer of this create I _want_ to write the most performant code I can right now, and reap cost/latency benefits over time without having to do anything. But I totally understand wanting to wait until the impact is more significant or the language makes the change trivial :)

davidbarsky:
> This being said, not making this change means that event definitions must be DeserializeOwned. When the deserialization gains begin to matter (e.g., due to fixing the other performance issues, or billing becoming per micro-second), then all consumers wanting to take advantage of the gains will incur the cost of making the change.
> 
> As a consumer of this create I want to write the most performant code I can right now, and reap cost/latency benefits over time without having to do anything. But I totally understand wanting to wait until the impact is more significant or the language makes the change trivial :)

(Disclosure: I no longer work at Amazon, so my vote/opinion will matter less—this project is more in @coltonweaver et. al.'s hands now, but I figured it made sense for me to respond rather than to ghost you.)

Broadly speaking, I think I agree with this sentiment of this statement. However, Rust's present limitations, such as the absence of GATs, implies that pervasive, cheap cloning in the style of [bytes](docs.rs/bytes) or [tower](https://docs.rs/tower/0.4.6/tower/buffer/index.html) is often necessary. If you're interested in making performance improvements here, I'd suggest looking into Arc'ing/Bytes'ing the payload. However, I feel obligated to warn you that you might not see _very little_ benefit in this runtime due to factors outside of this library's control.

Since I don't have the permissions to do this, my suggestion to the maintainers of this project would be close this PR and revisit this approach if/when Lambda supports microsecond-level billing _and_ Rust has stabilized support for GATs.

bahildebrand:
Weighing the possible gains here against the complexity of the solution, and the need to prepare another breaking change so soon after 0.3 I'm inclined to agree with @davidbarsky here. 

@duarten thank you for all your work on this, but I'm going to go ahead and close this PR. Perhaps this is something we can revisit in the future, but I'm not sure it's the right direction considering the current environment.

rimutaka:
@duarten , keep us updated if you make any progress on this. Closing this PR is probably the right thing to do, but I think this problem is worth solving one way or the other.

PR #312, opened 2021-03-28
Change run loop to execute all invocations concurrently

### LINKED ISSUE #311: Execute lambda invocations concurrently
In #310  it was pointed out that we are currently not executing lambda invocations concurrently. This code section has us spawn a task that is immediately waited, so another invocations is not pulled in until the current is finished.

https://github.com/awslabs/aws-lambda-rust-runtime/blob/d3ff4352c4b0e64fb46af2ca425db39211237e26/lambda-runtime/src/lib.rs#L160-L162

This loop needs to be reworked so that incoming invocations can be processed concurrently.

bahildebrand:
closes #311 

This is my first past at running all invocations concurrently. This code as it exists has a pretty major flaw in that all calls to [next](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html) need to be resolved before we would start executing any of the handlers. We could conceive of a situation where we have a large number of invocations queued, and we would not be able start executing the requests until all calls to next were complete. This would likely reduce the overall response time for the set of invocations, but invocations early in the queue could would be eating the next calls of those later in the queue.

### DIFF
--- a/lambda-runtime/src/client.rs
+++ b/lambda-runtime/src/client.rs
@@ -3,7 +3,7 @@ use http::{uri::Scheme, Request, Response, Uri};
 use hyper::{client::HttpConnector, Body};
 use std::fmt::Debug;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub(crate) struct Client<C = HttpConnector> {
     pub(crate) base: Uri,
     pub(crate) client: hyper::Client<C>,
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -114,7 +114,7 @@ enum BuilderError {
 }
 
 struct Runtime<C: Service<http::Uri> = HttpConnector> {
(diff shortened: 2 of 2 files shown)


bahildebrand:
After thinking about this I believe I can fix these problems by creating separate tasks for retrieving the next invocation, and processing each invocation. I'll put out another revision later tonight.

rimutaka:
This is actually a major change in behavior. The current version, with all its flaws gives me the guarantee of only one invocation per instance. Then I can manage the resources per lambda and scale out by running as many lambdas as I need. If we have multiple invocations per instance how can I control the resources and avoid race conditions? Maybe I misunderstood the nature of the change.

bahildebrand:
Yeah, I put this out as a draft because I was pretty unsure on it. I think the runtimes for lambda are supposed to be serialized, and the more I read the documentation I think the previous configuration was functionally correct.

Issue #310, opened 2021-03-28
avoid static lifetime requirement for handler functions


rasviitanen:
I would like to remove the requirement of having `'static` bounds on the handler generics if possible. This would make it easier to pass in shared resources that have been setup/created before creating the handler and avoid a bunch of cloning.

It seems like a `tokio::spawn` is used to produce a future here: 
https://github.com/awslabs/aws-lambda-rust-runtime/blob/d3ff4352c4b0e64fb46af2ca425db39211237e26/lambda-runtime/src/lib.rs#L160-L162

By taking a guess, I would assume that the reason is to catch a panic when generating the future (the JoinHandle is immediately awaited and nothing else seems to be going on). 
I might very well be wrong here, so if the reason for using a `tokio::spawn` has to do with something else, please correct me.

Since `tokio::spawn` requires both the future and the output to be `'static`, all the handler generics needs to be bounded by `'static`. This makes it quite cumbersome to use shared resources. Consider:

```rust
let client = SharedClient::try_from("Shared Client 1 (perhaps a database)")?;
let client_ref = &client;
lambda_runtime::run(handler_fn(|event: Request, ctx: Context| async move {
    let command = event.command;
    Ok::<Response, Error>(client_ref.response(ctx.request_id, command))
}))
.await?;
```

This would not compile, as `client_ref` is not `'static`. To me it would make sense if you were able to do some initial setup (e.g. database clients/initializing env-configs etc) before calling `run`.

I was thinking it would be possible to remove the static requirements by removing `tokio::spawn`?
Perhaps it can be replaced with a `catch_unwind`?

I have been experimenting a bit here (I added an example as well):
https://github.com/rasviitanen/aws-lambda-rust-runtime/commit/9f9b6746f89eda39f95b149c8fca699c4bcd1203
Would this work or would it result in other problems I have not considered?

PS great work on the 0.3 release! :)

bahildebrand:
We use `tokio::spawn` in this case to allow us to handle multiple requests at once. I'd recommend reading through the whole tutorial, but [this](https://tokio.rs/tokio/tutorial/spawning) section in particular goes over why would want to use `tokio::spawn`.  Your changes would allow us to only respond to a single invocation of the lambda at a time.

[...]

PR #319, opened 2021-04-20
switch panic handling method

rasviitanen:
*Issue:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/310

*Description of changes:*
Uses `std::panic::catch_unwind` instead of spawning a new tokio task to catch any panics.
By using `std::panic::AssertUnwindSafe` we don't have to add any new bounds on the handler, so this should be backwards compatible.

Using `std::panic::AssertUnwindSafe` should be OK. This is essentially what tokio did under the hood with the previous implementation, so the behavior should be equivalent. See:
https://github.com/tokio-rs/tokio/blob/b42f21ec3e212ace25331d0c13889a45769e6006/tokio/src/runtime/task/harness.rs#L409

What this attempts to solve:
Get rid of the `Send + Sync + 'static` bounds.

### DIFF
--- a/lambda-runtime/examples/shared_resource.rs
+++ b/lambda-runtime/examples/shared_resource.rs
@@ -1,0 +1,63 @@
+// This example demonstrates use of shared resources such as DB connections
+// or local caches that can be initialized at the start of the runtime and
+// reused by subsequent lambda handler calls.
+// Run it with the following input:
+// { "command": "do something" }
+
+use lambda_runtime::{handler_fn, Context, Error};
+use log::LevelFilter;
+use serde::{Deserialize, Serialize};
+use simple_logger::SimpleLogger;
+
+/// This is also a made-up example. Requests come into the runtime as unicode
+/// strings in json format, which can map to any structure that implements `serde::Deserialize`
+/// The runtime pays no attention to the contents of the request payload.
+#[derive(Deserialize)]
+struct Request {
+    command: String,
+}
+
+/// This is a made-up example of what a response structure may look like.
+/// There is no restriction on what it can be. The runtime requires responses
+/// to be serialized into json. The runtime pays no attention
+/// to the contents of the response payload.
+#[derive(Serialize)]
+struct Response {
+    req_id: String,
+    msg: String,
+}
+
+struct SharedClient {
+    name: &'static str,
+}
+
+impl SharedClient {
+    fn new(name: &'static str) -> Self {
+        Self {
+            name
+        }
+    }
+
+    fn response(&self, req_id: String, command: String) -> Response {
+        Response {
+            req_id,
+            msg: format!("Command {} executed by {}.", command, self.name),
+        }
+    }
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    // required to enable CloudWatch error logging by the runtime
+    // can be replaced with any other method of initializing `log`
+    SimpleLogger::new().with_level(LevelFilter::Info).init().unwrap();
+
+    let client = SharedClient::new("Shared Client 1 (perhaps a database)");
+    let client_ref = &client;
+    lambda_runtime::run(handler_fn(move |event: Request, ctx: Context| async move {
+        let command = event.command;
+        Ok::<Response, Error>(client_ref.response(ctx.request_id, command))
+    }))
+    .await?;
+    Ok(())
+}
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -13,7 +13,7 @@ use std::{
     convert::{TryFrom, TryInto},
     env, fmt,
     future::Future,
-    sync::Arc,
+    panic,
 };
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_stream::{Stream, StreamExt};
@@ -93,7 +93,7 @@ pub struct HandlerFn<F> {
 impl<F, A, B, Error, Fut> Handler<A, B> for HandlerFn<F>
 where
     F: Fn(A, Context) -> Fut,
-    Fut: Future<Output = Result<B, Error>> + Send,
+    Fut: Future<Output = Result<B, Error>>,
     Error: Into<Box<dyn std::error::Error + Send + Sync + 'static>> + fmt::Display,
 {
     type Error = Error;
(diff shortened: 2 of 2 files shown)


lambda-runtime/examples/shared_resource.rs
```diff
@@ -0,0 +1,63 @@
+// This example requires the following input to succeed:
```
rimutaka, line 1: This example demonstrates use of shared resources such as DB connections or local caches that can be initialized at the start of the runtime and reused by subsequent lambda handler calls. Run it with the following input:

rasviitanen: Very good comment, thank you! Fixed in the latest push

coltonweaver (approved):
Great contribution, thank you! LGTM

coltonweaver:
Added @bahildebrand since he's been more involved with this discussion too.

rimutaka:
Can we not rush this PR? Even if it is backward compatible it is a major change in the expected behavior with some unexpected results in edge cases.

1. catch_unwind does not guarantee all panics will be caught
2. AssertUnwindSafe is not bullet-proof either

Even if the compiler guarantees memory safety of panics it does not statically guarantee that all panics will be safely unwound. Correct me if I'm wrong here.

I fully support this change, but would like to give it more time to understand all the implications.

coltonweaver:
> Can we not rush this PR? Even if it is backward compatible it is a major change in the expected behavior with some unexpected results in edge cases.
> 
>     1. catch_unwind does not guarantee all panics will be caught
> 
>     2. AssertUnwindSafe is not bullet-proof either
> 
> 
> Even if the compiler guarantees memory safety of panics it does not statically guarantee that all panics will be safely unwound. Correct me if I'm wrong here.
> 
> I fully support this change, but would like to give it more time to understand all the implications.

Intention definitely wasn't to rush this PR, which is why I asked Blake to also take a look :). Completely agree with your points, my desire is to get more eyes on it in case something has been missed.

bahildebrand:
> 
> 
> Can we not rush this PR? Even if it is backward compatible it is a major change in the expected behavior with some unexpected results in edge cases.
> 
>     1. catch_unwind does not guarantee all panics will be caught
> 
>     2. AssertUnwindSafe is not bullet-proof either
> 
> 
> Even if the compiler guarantees memory safety of panics it does not statically guarantee that all panics will be safely unwound. Correct me if I'm wrong here.
> 
> I fully support this change, but would like to give it more time to understand all the implications.

My reading of [catch_unwind(https://doc.rust-lang.org/std/panic/fn.catch_unwind.html) is that we would be guaranteed to catch any panics that were inherent to the types composing the handler since it requires an [UnwindSafe](https://doc.rust-lang.org/std/panic/trait.UnwindSafe.html) implementation. I believe the only things that are `!UnwindSafe` are mutable references, which is not something that would make a ton of sense to use in this context.  

One danger would be if someone were enable [abort](https://doc.rust-lang.org/edition-guide/rust-2018/error-handling-and-panics/aborting-on-panic.html) panics on their binary. I'm not sure if the compiler will get angry since we have a call to `catch_unwind`, or if it will allow someone to call an abort panic from within their handler. That being said I believe the behavior would be equivalent to `spawn`. I think this is kind of a foot gun that has existed from before this change.

Overall I don't see any problems with this change. It might be worthwhile to add a note that users should not compile this library panic abort.

rasviitanen:
> I believe the only things that are !UnwindSafe are mutable references, which is not something that would make a ton of sense to use in this context.

I think there are a could of other things that would make sense in this context that are `UnwindSafe`. For example [`std::sync::Once`](https://doc.rust-lang.org/std/sync/struct.Once.html) or [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html). That said, [a regular Rust thread isn't really Unwindsafe either](https://doc.rust-lang.org/src/std/thread/mod.rs.html#473-475), I don't see why we would need to it here.

About `panic = abort`: The compiler will not complain. It will allow someone to abort on panic if they want to ( `catch_unwind` *only* catches unwinding panics).

lambda-runtime/examples/shared_resource.rs
```diff
@@ -0,0 +1,66 @@
...
+struct Response {
+    req_id: String,
+    msg: String,
+}
+
+struct SharedClient {
+    name: &'static str,
+}
+
+impl TryFrom<&'static str> for SharedClient {
```
LucioFranco, line 36: Why does this need to be tryfrom if its infallible?

lambda-runtime/src/lib.rs
```diff
@@ -154,12 +153,10 @@ where
             trace!("{}", std::str::from_utf8(&body)?); // this may be very verbose
             let body = serde_json::from_slice(&body)?;
 
-            let handler = Arc::clone(&handler);
             let request_id = &ctx.request_id.clone();
-            #[allow(clippy::async_yields_async)]
-            let task = tokio::spawn(async move { handler.call(body, ctx) });
+            let task = panic::catch_unwind(panic::AssertUnwindSafe(|| handler.call(body, ctx)));
```
LucioFranco, line 157: Why this over `set_hook`?

davidbarsky: The runtime needs to report any panics to the invocation error endpoint: https://docs.aws.amazon.com/lambda/latest/dg/runtimes-api.html.

(I'm assuming that the `set_hook` you're referring to would be set by the customer?)

lambda-runtime/examples/shared_resource.rs
rasviitanen: The initial though was to avoid consumers of this library (or future contributors) to not run into the idea where: "Hey why is this a `From`, you can just do ` let client = &SharedClient("name");` directly". The problem with this is that `client` in this case would be static, so it wouldn't be as clear that you can use some non-static reference. Whereas `TryFrom` would make people think one extra step and realize that you can do some more advanced setup.

But I am glad you pointed it out; the example in its current form is indeed confusing.
The better solution here is to simply implement a `new` function for the `SharedClient` and force it to be bound like that instead. Then I think the barrier to understand the example will be much lower.

I'll work on a fix, thanks!

bahildebrand (approved):
LGTM. I'm going to leave this open for another day in case anyone else has comments they'd like to get in. Otherwise great job @rasviitanen. This is an awesome contribution!

coltonweaver (approved):

seanpianka:
Hi @bahildebrand, what is the timeline for merging this PR?

bahildebrand:
Merging this now. Rust fmt is broken, but I'll push up another PR to fix it.

PR #322, opened 2021-05-04
fix: don't panic on missing headers under SAM

Fishrock123:
*Description of changes:*

SAM is missing these headers, making local testing impossible without this change.

### DIFF
--- a/lambda-runtime/src/types.rs
+++ b/lambda-runtime/src/types.rs
@@ -128,14 +128,18 @@ impl TryFrom<HeaderMap> for Context {
                 .to_str()?
                 .parse()
                 .expect("Missing deadline"),
-            invoked_function_arn: headers["lambda-runtime-invoked-function-arn"]
-                .to_str()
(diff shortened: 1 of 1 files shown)


coltonweaver (commented):
This is not the first time we've received this request, and I understand the issue you're facing. However, I don't believe that this runtime is the right place to inject these default values, and could lead to weird behavior for developers relying on this in production Lambda.

https://github.com/awslabs/aws-lambda-rust-runtime/pull/314

I would recommend looking at the comment made by @davidbarsky on this previous PR and reach out to the upstream repo to try and fix this regression. https://github.com/aws/aws-sam-cli/issues/2715

coltonweaver:
I'm going to close this PR as I don't believe this runtime to be the right place to try and remediate this issue. If you strongly disagree, please feel free to reply and we can discuss.

Fishrock123:
I feel like it should be vocal about these missing somehow but I don't know that it is worth crashing? Not sure.

PR #323, opened 2021-05-05
feat: expose types useful for other runtimes

Fishrock123:
Hi Folks - we'd like to run Rust on AWS Lambda, however our Rust services stack is async-std based. (We very keenly picked [Tide](https://docs.rs/tide/0.16.0/tide/index.html) as our server framework of choice due to general approachability for new rust developers!)

Things do mostly work running this on-top of Tokio, but we'd like to have a full async-std version of the lambda http handler, ideally written as a Tide [Listener](https://docs.rs/tide/0.16.0/tide/listener/trait.Listener.html). Tide's http stack includes a feature to translate between hyperium's http requests and http-rs's http-types requests, using this in combination with exposing more input/output structures here, we can reduce a lot of code we'd otherwise have to replicate (particularly incoming request deserialization and http request restructuring code). It would be greatly appreciated if some form of this could make it's way into these crates.

*Description of changes:*

Other runtimes would benefit from the types for deserializing incoming lambda requests (and a few others), so, in an effort to not have people replicate everything, expose these extra parts but hide them from the docs, since this essentially unstable code spelunking at this point.

### DIFF
--- a/lambda-runtime/src/types.rs
+++ b/lambda-runtime/src/types.rs
@@ -3,11 +3,12 @@ use http::HeaderMap;
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, convert::TryFrom};
 
+#[doc(hidden)]
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct Diagnostic {
-    pub(crate) error_type: String,
-    pub(crate) error_message: String,
+pub struct Diagnostic {
+    pub error_type: String,
+    pub error_message: String,
 }
 
 #[test]
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -68,7 +68,8 @@ use lambda_runtime::{Error, Handler as LambdaHandler};
 mod body;
 pub mod ext;
 pub mod request;
-mod response;
+#[doc(hidden)]
+pub mod response;
 mod strmap;
 pub use crate::{body::Body, ext::RequestExt, response::IntoResponse, strmap::StrMap};
 use crate::{
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -107,7 +107,7 @@ where
 
 /// tranformation from http type to internal type
 impl LambdaResponse {
-    pub(crate) fn from_response<T>(request_origin: &RequestOrigin, value: Response<T>) -> Self
+    pub fn from_response<T>(request_origin: &RequestOrigin, value: Response<T>) -> Self
     where
         T: Into<Body>,
     {
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -20,15 +20,17 @@ use tokio_stream::{Stream, StreamExt};
 use tower_service::Service;
 use tracing::{error, trace};
 
+#[doc(hidden)]
+pub mod requests;
+
 mod client;
-mod requests;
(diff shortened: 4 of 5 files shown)


rimutaka:
I think it would be of even more value to the community if we documented the newly exposed parts instead of hiding them from the docs.

lambda-runtime/src/types.rs
```diff
@@ -3,11 +3,12 @@ use http::HeaderMap;
...
 
+#[doc(hidden)]
 #[derive(Debug, PartialEq, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
-pub(crate) struct Diagnostic {
-    pub(crate) error_type: String,
-    pub(crate) error_message: String,
+pub struct Diagnostic {
+    pub error_type: String,
+    pub error_message: String,
```
TransientError, line 11: While I think it's fine to expose the struct itself, I think we probably don't want to expose the fields for forward compatibility reasons e.g. if we add a field then clients would have to make code changes. I would prefer we expose constructors / methods instead.

bahildebrand:
While I understand the motivation behind this PR, I'm not sure I'm onboard with how we're going about it. I'm on the same page as @TransientError in that we don't want to expose these internals for fear of breaking forwards compatibility. I think even adding constructors/methods to fetch these fields still exposes a bit too much of the innards of the runtime. 

This would be quite a bit more work, but I wonder if it would be worthwhile to have async-std/tokio as seperate features that can be enabled. We recently had a request for a non-async version, and I'm wondering now if we should start considering these alternate configurations more seriously.

lambda-runtime/src/types.rs
Fishrock123: Perhaps, but I think this is kinda just the cost of being an alternate runtime here.

If that's the desired route though, then these should probably also get `#[non_exhaustive]`.

Fishrock123:
> I think even adding constructors/methods to fetch these fields still exposes a bit too much of the innards of the runtime.

Hence the undocumented-ness. I'd also be ok instead by documenting but putting this behind an extra feature flag.

> This would be quite a bit more work, but I wonder if it would be worthwhile to have async-std/tokio as seperate features that can be enabled.

That would probably be most ideal, yeah. That is really quite substantial though, and I do need something sooner, even if that 'thing' is working off a patched branch of this repo. I'll publish a repo of what I've been working on hopefully later today.

Ideally for me the end result is something that could be wrapped and used as the implementation for a Tide listener. Aside from tokio/hyperium-http I don't _think_ there's anything else block that but it's also hard to test.

> ... I'm wondering now if we should start considering these alternate configurations more seriously.

I mean, both my employer and myself hope so. :smile:

Fishrock123:
I was hoping to publish the current state of my Tide listener today, but I wasn't able to get to it, and I'm OO from now through next week. I'll publish that and update anything once I am back.

Fishrock123:
This is the listener backend for Tide that I have working: https://github.com/Fishrock123/tide-lambda-listener

(It depends on this branch, feel free to take a look at the ... jumping through a couple hoops and back ... of code)

Fishrock123:
It would really be much easier for us if we could just get some version of this merged first without needing to write extensive docs, would an unstable cargo feature do?

bahildebrand:
Sorry for the bit of radio silence. I was also out on vacation for a bit.

I'm going to close this out, as I'm not very comfortable exposing these internals. I've opened #331, and I will make it the main feature of a 0.4.0 release. Sorry that were not able to support this on a shorter time frame.

Issue #271, opened 2020-12-06
Incompatible with lambda environment simulation by SAM


ichyo:
Using `lambda_http` from the master branch, Invoking `sam local invoke` or `sam local start-api` returns the following error.

```
$ sam local generate-event apigateway aws-proxy | sam local invoke HelloWorldFunction
Invoking not.used (provided)
Skip pulling image and use local one: amazon/aws-sam-cli-emulation-image-provided:rapid-1.13.2.

Mounting /home/***/dev/src/github.com/ichyo/aws-sam-rust-hello-world/.aws-sam/build/HelloWorldFunction as /var/task:ro,delegated inside runtime container
START RequestId: 14785d12-91f4-4630-a9ba-1d5bc88a0d86 Version: $LATEST
thread 'main' panicked at 'no entry found for key "lambda-runtime-invoked-function-arn"', /home/***/.cargo/git/checkouts/aws-lambda-rust-runtime-7c865cce90132439/13aa8f0/lambda/src/types.rs:131:35
stack backtrace:
   0: rust_begin_unwind
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:483
   1: std::panicking::begin_panic_fmt
             at /rustc/7eac88abb2e57e752f3302f02be5f3ce3d7adfb4/library/std/src/panicking.rs:437
   2: <lambda::types::Context as core::convert::TryFrom<http::header::map::HeaderMap>>::try_from
   3: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   4: std::thread::local::LocalKey<T>::with
   5: tokio::runtime::enter::Enter::block_on
   6: tokio::runtime::thread_pool::ThreadPool::block_on
   7: tokio::runtime::context::enter
   8: hello::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
END RequestId: 14785d12-91f4-4630-a9ba-1d5bc88a0d86
REPORT RequestId: 14785d12-91f4-4630-a9ba-1d5bc88a0d86  Init Duration: 0.16 ms  Duration: 37.97 ms      Billed Duration: 100 ms Memory Size: 128 MB     Max Memory Used: 128 MB
```

[...]

Issue #38, opened 2018-12-09
Can't get this working with "sam local"


radix:
When trying to run my lambda function with `sam local start-api`, I get the following error about `_LAMBDA_SERVER_PORT` not being defined:

```
(sam) PS C:\Users\radix\Projects\pandt> sam local start-api --template .\SAM.yaml
2018-12-09 15:48:26 Mounting pandt at http://127.0.0.1:3000/ [GET]
2018-12-09 15:48:26 You can now browse to the above endpoints to invoke your functions. You do not need to restart/reload SAM CLI while working on your functions changes will be reflected instantly/automatically. You only need to restart SAM CLI if you update your AWS SAM template
2018-12-09 15:48:27  * Running on http://127.0.0.1:3000/ (Press CTRL+C to quit)
2018-12-09 15:48:35 Invoking lollerskates (provided)

Fetching lambci/lambda:provided Docker container image......
2018-12-09 15:48:36 Mounting C:\Users\radix\Projects\pandt\artifacts as /var/task:ro inside runtime container
thread 'main' panicked at 'failed to start runtime: environment error: the _LAMBDA_SERVER_PORT variable must specify a valid port to listen on', /home/rust/.cargo/git/checkouts/rust-aws-lambda-2a7b20b15eabd9cc/1435a4d/aws_lambda/src/lib.rs:56:31
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:211
   3: std::panicking::default_hook
             at libstd/panicking.rs:227
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:476
   5: std::panicking::continue_panic_fmt
             at libstd/panicking.rs:390
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:345
   7: aws_lambda::start::{{closure}}
   8: aws_lambda::start
   9: pandt_lambda::main
  10: std::rt::lang_start::{{closure}}
  11: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:310
  12: __rust_maybe_catch_panic
[...]

PR #332, opened 2021-06-19
Fix sam local support for missing headers (#38)

kenshih:
Hello, I hope this finds you well!

I was trying to run `sam local` in order to make using rust more viable at work & got it working locally, so I thought I'd throw up a pull request, in case it solves an issue for other alpha runtime users.

-----

**Fix sam local support for missing headers (#38)**

*Issue [#38](https://github.com/awslabs/aws-lambda-rust-runtime/issues/38)*
also related to: 
* [aws/aws-lambda-runtime-interface-emulator #11](https://github.com/aws/aws-lambda-runtime-interface-emulator/issues/11)
* [#271](https://github.com/awslabs/aws-lambda-rust-runtime/issues/271)

*Example of problem:*
```
# when running "sam invoke" against an existing "sam start-lambda", something like this happened for me
...
START RequestId: 38dd307a-a56a-4def-9adf-136a5a43ac70 Version: $LATEST
thread 'main' panicked at 'no entry found for key "lambda-runtime-invoked-function-arn"', /Users/ken/wk/experiments/rust/aws-lambda-rust-runtime/lambda-runtime/src/types.rs:131:35
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: std::panicking::begin_panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:435:5
   2: <lambda_runtime::types::Context as core::convert::TryFrom<http::header::map::HeaderMap>>::try_from
   3: ken_test_rust_lambda::main::{{closure}}
   4: std::thread::local::LocalKey<T>::with
   5: tokio::runtime::enter::Enter::block_on
   6: tokio::runtime::Runtime::block_on
   7: ken_test_rust_lambda::main
...

```

After the following changes, the above stack trace was no more & the lambda invoke was able to run successfully to completion with expected response payload.

*Description of changes:*

Fix sam local support for missing headers (#38)
* Added unit tests for TryFrom<HeaderMap> for Context

The issue was that when the lambda-runtime is invoked by the sam local simulated environment, the headers `lambda-runtime-invoked-function-arn` and `lambda-runtime-trace-id` are not provided (as they are in the production cloud environment). 

I have confirmed this issue still exists in `v0.3.0` and on `master`

I'm not aware if these headers are supposed to be invariants or not, but, it seems that other language runtimes are more forgiving with these missing headers with `sam local invoke`, so it seems appropriate for the rust runtime to be resilient to these as well?

Instead of panic-ing on a missing header key, this code now provides a default value for the two headers in question. The default string indicates that they were not found in the environment as expected.

Let me know if you the style is not to your liking, happy to change as you suggest and/or accept incoming changes. I am rather new to rust. Apologies if I have misinterpreted any conventions.

Thanks for considering this PR.

- Ken
---

### DIFF
--- a/lambda-runtime/src/types.rs
+++ b/lambda-runtime/src/types.rs
@@ -1,5 +1,5 @@
 use crate::{Config, Error};
-use http::HeaderMap;
+use http::{HeaderMap, HeaderValue};
 use serde::{Deserialize, Serialize};
 use std::{collections::HashMap, convert::TryFrom};
 
@@ -120,25 +120,104 @@ impl TryFrom<HeaderMap> for Context {
     type Error = Error;
     fn try_from(headers: HeaderMap) -> Result<Self, Self::Error> {
         let ctx = Context {
-            request_id: headers["lambda-runtime-aws-request-id"]
-                .to_str()
-                .expect("Missing Request ID")
-                .to_owned(),
-            deadline: headers["lambda-runtime-deadline-ms"]
+            request_id: headers
+                .get("lambda-runtime-aws-request-id")
+                .expect("missing lambda-runtime-aws-request-id header")
                 .to_str()?
-                .parse()
-                .expect("Missing deadline"),
-            invoked_function_arn: headers["lambda-runtime-invoked-function-arn"]
-                .to_str()
-                .expect("Missing arn; this is a bug")
                 .to_owned(),
-            xray_trace_id: headers["lambda-runtime-trace-id"]
-                .to_str()
-                .expect("Invalid XRayTraceID sent by Lambda; this is a bug")
+            deadline: headers
+                .get("lambda-runtime-deadline-ms")
+                .expect("missing lambda-runtime-deadline-ms header")
+                .to_str()?
+                .parse::<u64>()?,
+            invoked_function_arn: headers
+                .get("lambda-runtime-invoked-function-arn")
+                .unwrap_or(&HeaderValue::from_static(
+                    "No header lambda-runtime-invoked-function-arn found.",
+                ))
+                .to_str()?
+                .to_owned(),
+            xray_trace_id: headers
+                .get("lambda-runtime-trace-id")
+                .unwrap_or(&HeaderValue::from_static("No header lambda-runtime-trace-id found."))
+                .to_str()?
                 .to_owned(),
             ..Default::default()
         };
         Ok(ctx)
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn context_with_expected_values_and_types_resolves() {
+        let mut headers = HeaderMap::new();
+        headers.insert("lambda-runtime-aws-request-id", HeaderValue::from_static("my-id"));
+        headers.insert("lambda-runtime-deadline-ms", HeaderValue::from_static("123"));
+        headers.insert(
+            "lambda-runtime-invoked-function-arn",
+            HeaderValue::from_static("arn::myarn"),
+        );
+        headers.insert("lambda-runtime-trace-id", HeaderValue::from_static("arn::myarn"));
+        let tried = Context::try_from(headers);
+        assert!(tried.is_ok());
+    }
+
+    #[test]
+    fn context_with_certain_missing_headers_still_resolves() {
+        let mut headers = HeaderMap::new();
+        headers.insert("lambda-runtime-aws-request-id", HeaderValue::from_static("my-id"));
+        headers.insert("lambda-runtime-deadline-ms", HeaderValue::from_static("123"));
+        let tried = Context::try_from(headers);
+        assert!(tried.is_ok());
+    }
+
+    #[test]
+    fn context_with_bad_deadline_type_is_err() {
+        let mut headers = HeaderMap::new();
+        headers.insert("lambda-runtime-aws-request-id", HeaderValue::from_static("my-id"));
+        headers.insert(
+            "lambda-runtime-deadline-ms",
+            HeaderValue::from_static("BAD-Type,not <u64>"),
+        );
+        headers.insert(
+            "lambda-runtime-invoked-function-arn",
+            HeaderValue::from_static("arn::myarn"),
+        );
+        headers.insert("lambda-runtime-trace-id", HeaderValue::from_static("arn::myarn"));
+        let tried = Context::try_from(headers);
+        assert!(tried.is_err());
+    }
+
+    #[test]
+    #[should_panic]
+    #[allow(unused_must_use)]
+    fn context_with_missing_request_id_should_panic() {
+        let mut headers = HeaderMap::new();
+        headers.insert("lambda-runtime-aws-request-id", HeaderValue::from_static("my-id"));
+        headers.insert(
+            "lambda-runtime-invoked-function-arn",
+            HeaderValue::from_static("arn::myarn"),
+        );
+        headers.insert("lambda-runtime-trace-id", HeaderValue::from_static("arn::myarn"));
+        Context::try_from(headers);
+    }
+
+    #[test]
+    #[should_panic]
+    #[allow(unused_must_use)]
+    fn context_with_missing_deadline_should_panic() {
+        let mut headers = HeaderMap::new();
+        headers.insert("lambda-runtime-deadline-ms", HeaderValue::from_static("123"));
+        headers.insert(
+            "lambda-runtime-invoked-function-arn",
+            HeaderValue::from_static("arn::myarn"),
+        );
+        headers.insert("lambda-runtime-trace-id", HeaderValue::from_static("arn::myarn"));
+        Context::try_from(headers);
     }
 }
 


lambda-runtime/src/types.rs
```diff
@@ -120,28 +120,107 @@ impl TryFrom<HeaderMap> for Context {
     type Error = Error;
     fn try_from(headers: HeaderMap) -> Result<Self, Self::Error> {
         let ctx = Context {
-            request_id: headers["lambda-runtime-aws-request-id"]
-                .to_str()
-                .expect("Missing Request ID")
+            request_id: headers
+                .get("lambda-runtime-aws-request-id")
```
kenshih, line 124: instead of implicitly failing (by panic-ing on a missing key), i retain the original panic behavior, but now we panic at the `expect` as expected :)

I chose not to change behavior here, since sam local does fulfill this requirement, so no need to be prematurely lax.

lambda-runtime/src/types.rs
```diff
@@ -120,28 +120,107 @@ impl TryFrom<HeaderMap> for Context {
...
+                .expect("missing lambda-runtime-aws-request-id header")
+                .to_str()?
                 .to_owned(),
-            deadline: headers["lambda-runtime-deadline-ms"]
+            deadline: headers
+                .get("lambda-runtime-deadline-ms")
```
kenshih, line 129: took same approach as "lambda-runtime-aws-request-id" here.

lambda-runtime/src/types.rs
```diff
@@ -120,28 +120,107 @@ impl TryFrom<HeaderMap> for Context {
...
+                .expect("missing lambda-runtime-deadline-ms header")
+                .to_str()?
+                .parse::<u64>()?,
+            invoked_function_arn: headers
+                .get("lambda-runtime-invoked-function-arn")
+                .unwrap_or(&HeaderValue::from_static(
+                    "No header lambda-runtime-invoked-function-arn found.",
```
kenshih, line 136: Instead of honoring ARN format, I only honored the "String" type of ARN. In production, of course, it should be a real ARN. But in development (such as sam local), it should be clear that something is wrong with the environment, if the developer somehow needs to inspect this context.

lambda-runtime/src/types.rs
```diff
@@ -120,28 +120,107 @@ impl TryFrom<HeaderMap> for Context {
...
-            invoked_function_arn: headers["lambda-runtime-invoked-function-arn"]
-                .to_str()
-                .expect("Missing arn; this is a bug")
                 .to_owned(),
-            xray_trace_id: headers["lambda-runtime-trace-id"]
-                .to_str()
-                .expect("Invalid XRayTraceID sent by Lambda; this is a bug")
+            xray_trace_id: headers
+                .get("lambda-runtime-trace-id")
+                .unwrap_or(&HeaderValue::from_static("No header lambda-runtime-trace-id found."))
```
kenshih, line 142: As reported in other issues, as well as confirming this myself, sam local invoke also leaves off this header.

lambda-runtime/src/types.rs
```diff
@@ -120,28 +120,107 @@ impl TryFrom<HeaderMap> for Context {
...
+                .to_str()?
                 .to_owned(),
             ..Default::default()
         };
         Ok(ctx)
     }
 }
 
+#[cfg(test)]
```
kenshih, line 151: I wasn't sure exactly where to put these tests, so I just placed them right after the thing they are testing. 

I noticed the other test in this file does not use #[cfg(test)]... I wasn't sure if I should have left it off here or not. As I understand, without adding #[cfg(test)], it would artificially bloat the release build as well. As I said, I'm still new enough to rust though, so please let me know if I misunderstand!

kenshih:
hum. so, i just noticed this:
https://github.com/awslabs/aws-lambda-rust-runtime/pull/322

I'm curious if the position has changed at all.

Given that other runtimes seem fine with these missing headers in the context of sam local, I'm wondering why the rust runtime would be different. I realize this might be something to revisit, if that context header becomes an invariant for later runtime machinery; but doesn't this bug make it even less likely that this runtime will rely an that header given the slower feedback cycle for development on it? 

For the context of the the runtime consumer, that is, the lambda application developer, the devx of lambda without sam local is a big gap. I realized that while I was writing this template for myself: https://github.com/kenshih/rust-aws-lambda-template. I was reluctant to share it with anyone on my team without a good lightweight solution for local development.

kenshih:
Ok. So feel free to close this, as you like.

I'm seeing the arguments more clearly: https://github.com/awslabs/aws-lambda-rust-runtime/pull/314

I'll leave this open until you get a chance to see it, just as another +1 to the situation this runtime is left in.

Thanks for your work.

kenshih:
Revisiting what i did here, I now think neither approach (this PR, as well, not doing this PR) is good.

Following the [Robustness Principle](https://en.wikipedia.org/wiki/Robustness_principle)...
it seems like the runtime *should* allow the bad header (being "liberal with what it accepts" from others) which would make it less fragile; 
but, setting an arbitrary string for the values is not a good solution (not :"being conservative about what it passes along" to the consumer of the lambda environment).

I wonder if the runtime instead should: 
1. emit a non-blocking WARN about misconfiguration
2. the `Context` data structure should have Optional values.

The cost, of course, is a more complicated interface with the consumer (Option vs. guaranteed data). 

But, at the same time, good side effects are: it would encourage promulgation of the robustness principle down to the consumer & reflects the reality of the often imperfect situation inherent in layer & cross-subsystem dependencies. Contracts break. So do we live with months/if-ever delays in the meantime, or robust our way out of it, knowing we're exposed to some interface risk? I honestly don't know the right answer here. But curious to hear the balance the maintainers of this runtime hope to achieve.

Thanks for listening.

asg0451:
FWIW as a user of this project, I don't care at all about the lambda context, and I do wish we could run with sam local. Option values in Context isn't the worst idea in the world.

brainstorm:
Yes please, un-draft this PR and get it shipped, it'll help a ton to have a proper `sam local` run & debug loop.

kenshih:
So I'll be happy to un-Draft this, @brainstorm, thank you. But I need a Maintainer to allow me to merge, or at least comment about how I can move this into a solution that would be acceptable.

Would moving Context values to Option values and emitting WARN be an acceptable compromise, @Maintainers? In the interface code, I would link a comment to the doc of the contract interface and explain the robustness approach we're taking. Meaning, we can make the framework (in this case, the lambda runtime) more resilient than the contract it accepts data from via such defensive coding. While this may seem like a technique we don't need as much of in Rust, certainly the world outside of Rust does not abide by Rust assumptions, so at the edges, we still use resilient/robust techniques instead of falling back to fragility.

I think of it this way, if I'm an engineer/architect that wants to push my team to the latest platform, but I see it has a bug in headers it passes, I'm blocked on AWS fixing it before I can take advantage of latest bug fixes/improvements.

I realize assuming Context contract as golden means the runtime itself can make assumptions and more easily maintain future instrumentation, etc... but in that too, we should have graceful failure modes. So it might be a better assumption to data clean just inside the interface anyway.

Curious to hear your thoughts, @maintainers, and if you buy what I'm saying here.

coltonweaver:
Apologies for the radio silence, all. I've been on vacation a bit and also been very busy with another deliverable at work so I didn't have time to check everything out in detail. I've come back to looking at this now.

In the past I took a hard stance against setting defaults like this for the reasons you mentioned above, but at this point it appears this won't ever be resolved unless the runtime is more graceful in handling these missing headers for local environments. At this point I'm inclined to think the way you've done this PR right now is fine. I like that you provided default strings that make it abundantly clear that something funky is going on if someone were to run into this in production. I think I'm inclined to keep it this way instead of making values in the Lambda context optional, as that isn't really reflective of reality when run outside of SAM.

I'm curious to know what @bahildebrand thinks about this, but I think this PR is acceptable, personally.

lambda-runtime/src/types.rs
coltonweaver: Good catch, thanks for fixing this.

lambda-runtime/src/types.rs
coltonweaver: Good call. One of my original worries was if someone was using the ARN in some way it would be annoying to debug if the default looked like a real value.

coltonweaver (approved):

coltonweaver:
I'm going to go ahead and merge this, if there are any issues with this approach we can discuss.

coltonweaver:
Thanks much for the contribution! @kenshih

brainstorm:
This works neatly! Could you produce a release? I'm pinning to a commit right now, but I reckon that a ton of users would benefit from this.

coltonweaver:
@brainstorm Sure thing, I will work on getting 3.1 published today or tomorrow.

coltonweaver:
@brainstorm Okay, all pushed. Apologies on the delay of a couple days!

asg0451:
It works beautifully, thanks!

PR #333, opened 2021-06-30
Removed 'static bound requirements for the lambda_http crate and create a shared resources example for the crate

coltonweaver:
Following discussion in https://github.com/awslabs/aws-lambda-rust-runtime/issues/310, it makes sense to include these changes in the lambda_http crate as well. I've given this work a pass, and am looking for feedback now.

*Issue #, if available:* #310

*Description of changes:*

* Added explicit lifetime bounds on each of the handler and adaptor types in the lambda_http crate, so we don't have to rely on the futures for these being 'static any longer.
* Added an example of how one might use this to the examples folder. Have tested that these changes work in the Lambda console with an APIGW event:

```
Response
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "f9b73b61-82cf-4da5-b084-19f01ee76346: Client (random_client_name_1) invoked by Colton.",
  "isBase64Encoded": false
}
```

The request was a regular APIGW proxy event with the query string value "first_name" set to my name.

- [ X] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [ X] I confirm that I've made a best effort attempt to update all relevant documentation.

### DIFF
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -75,11 +75,7 @@ use crate::{
     request::{LambdaRequest, RequestOrigin},
     response::LambdaResponse,
 };
-use std::{
-    future::Future,
-    pin::Pin,
-    task::{Context as TaskContext, Poll},
-};
+use std::{future::Future, marker::PhantomData, pin::Pin, task::{Context as TaskContext, Poll}};
 
 /// Type alias for `http::Request`s with a fixed [`Body`](enum.Body.html) type
 pub type Request = http::Request<Body>;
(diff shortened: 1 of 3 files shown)


lambda-http/src/lib.rs
```diff
@@ -87,28 +88,31 @@ pub type Request = http::Request<Body>;
...
     fn call(&self, event: Request, context: Context) -> Self::Fut;
 }
 
 /// Adapts a [`Handler`](trait.Handler.html) to the `lambda_runtime::run` interface
-pub fn handler<H: Handler>(handler: H) -> Adapter<H> {
-    Adapter { handler }
+pub fn handler<'a, H: Handler<'a>>(handler: H) -> Adapter<'a, H> {
+    Adapter {
+        handler,
+        _pd: PhantomData,
```
bahildebrand, line 106: Can you give more context on what this is for? I'm not sure I understand.

coltonweaver: Sure, if you look down below it's because I added the explicit lifetime to the `Adapter` struct, because it holds a `Handler` that requires a lifetime parameter. Because the `'a` lifetime is unused in `Adapter`, otherwise, the compiler complains about the unused lifetime parameter we supplied to it. The PhantomData is just a zero-sized compiler dummy value that holds the `'a` lifetime and "uses" the lifetime to make the compiler happy.

bahildebrand: Ah, okay. That makes sense now thank you.

bahildebrand (approved):

Issue #338, opened 2021-08-06
Last version lambda_http 0.3.1 is not SemVer compatible.


bburnichon:
Hello,

I use renovate bot to keep my dependencies up-to-date.
When going from release `0.3.0` to `0.3.1` of crate `lambda_http`, I need to update code in order to make it compile.
Sadly, I can not share actual code with you.

Here is the compiler line complaining about added lifetime parameter:
```
error[E0726]: implicit elided lifetime not allowed here
  --> src/mslambda.rs:21:6
   |
21 | impl lambda_http::Handler for Router {
   |      ^^^^^^^^^^^^^^^^^^^^- help: indicate the anonymous lifetime: `<'_>`
```

To remain semver, version `0.3.1` should be yanked and renamed `0.4.0`

Apart from this, fixing the compile error is just a matter of adding the anonymous lifetime: 
```rust
impl lambda_http::Handler<'_> for Router {
```

coltonweaver:
Ah, yep. That appears to be the case. Apologies, I had neglected to consider this use case when publishing. Working on yanking and moving to 0.4.0 now. Thanks for bringing this to our attention.

coltonweaver:
Both of the runtime and http crates at 0.3.1 were yanked, and I republished them as 0.4.0. Again, thanks for the heads up. Going to close this, but feel free to reopen it.

bburnichon:
Thank you for your prompt support. I was on vacation and did not saw the new version was published so quickly! Great job for the async code, my code is faster than ever.

PR #339, opened 2021-08-06
Incrementing to 0.4.0

coltonweaver:
*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/338

*Description of changes:*
0.3.1 is not semantically versioned correctly as it added a lifetime to the Handler, which was not a backwards compatible change with users of 0.3.0 that implemented their own handlers. Republishing as 0.4.0 with this.


bahildebrand (approved):
LGTM!

Issue #334, opened 2021-08-01
Missing _X_AMZN_TRACE_ID environment variable


bnusunny:
Rust runtime does not assign x-ray trace id header value to environment variable '_X_AMZN_TRACE_ID'. Without this environment variable, x-ray won't be able to trace the request end to end. Please consider to assign this environment variable with correct value. 

https://docs.aws.amazon.com/lambda/latest/dg/configuration-envvars.html#configuration-envvars-runtime

PR #341, opened 2021-08-18
set _X_AMZN_TRACE_ID environment variable for x-ray

bnusunny:
*Issue #, if available:*

#334 

*Description of changes:*
set _X_AMZN_TRACE_ID environment variable for x-ray

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -152,6 +152,12 @@ where
             let event = event?;
             let (parts, body) = event.into_parts();
 
+            if let Some(xray_trace_id) = parts.headers.get("lambda-runtime-trace-id") {
+                env::set_var("_X_AMZN_TRACE_ID", xray_trace_id.to_str()?);
+            } else {
+                env::remove_var("_X_AMZN_TRACE_ID");
+            }
+
             let ctx: Context = Context::try_from(parts.headers)?;
             let ctx: Context = ctx.with_config(config);
             let body = hyper::body::to_bytes(body).await?;


lambda-runtime/src/lib.rs
```diff
@@ -152,6 +152,12 @@ where
             let event = event?;
             let (parts, body) = event.into_parts();
 
+            if let Some(xray_trace_id) = parts.headers.get("lambda-runtime-trace-id") {
+                env::set_var("_X_AMZN_TRACE_ID", xray_trace_id.to_str()?);
```
rimutaka, line 156: What if `xray_trace_id` is an empty string?

bnusunny: Inside Lambda execution environment, that header should have valid xray trace id or not present. In sam local, the header does not exist. 

Lambda golang runtime use similar approach to handle this. Code is [here](https://github.com/aws/aws-lambda-go/blob/e5da494987442caa3749213d5fc56ea6e440bf77/lambda/function.go#L62).

coltonweaver: As far as I'm aware, there is no guarantee that this header will *always* be set.

If you dig into the golang implementation a little more: https://github.com/aws/aws-lambda-go/blob/e5da494987442caa3749213d5fc56ea6e440bf77/lambda/invoke_loop.go#L78

We unpack the invoke from the runtime API here, which is calling `Get` on an `http.Header` value. The default behavior for that method is that if the header isn't defined, it returns an empty string `""`. We should be consistent with golang in this behavior. You could just say `let xray_trace_id = parts.headers.get("lambda-runtime-trace-id").unwrap_or_default();` and always set the env variable to the result.

coltonweaver: Actually, I totally forgot we already handle this. Take a look: https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/src/types.rs#L140-L144

We set the value to a default here. I would recommend changing that to `""` instead of the current default. This value is stored in the Context that we have in this lib here: https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/src/lib.rs#L155

You can just set the environment variable to whatever the current value of `ctx.xray_trace_id` is.

coltonweaver (changes requested):

Issue #342, opened 2021-09-07
Does lambda_http::Handler::Future requires Send ?


fmonniot:
Hello there,

The current implementation of `lambda_http`'s [`Handler::Fut`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-http/src/lib.rs#L97) requires a `Send` bound, but the underlying `lambda_runtime`'s [`Handler::Fut`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/src/lib.rs#L76) does not. Is there a reason why this bound is present? Future-proofing the API or remnant of the past perhaps? The crate still compiles (and the tests pass) after removing the bound.

Would the team be interesting in a PR with [this commit](https://github.com/fmonniot/aws-lambda-rust-runtime/commit/5737b61b818617bb1fba0636ad749687c9c528b7).

This open up some interesting avenue too. For example, I found this issue while trying to integrate actix on top of the `lambda_http` crate (swaping out `HttpServer` with `lambda_runtime::run(lambda_http::handler` basically).

Thanks !

coltonweaver:
Hi @fmonniot, thanks for pointing this out! This is indeed unintentional. I missed this change whenever I made the changes to remove the `'static` bounds from `lambda_http` to match the runtime. If you want to go ahead and publish your linked commit as a PR, I would happily review it.

fmonniot:
No worry and thanks for the prompt response. I opened #344 with the change.

coltonweaver:
Just approved and merged your code in. Thanks for making this change! Feel free to reopen this issue if needed.

singularsyntax:
Has there been a regression on this? Or a decision to reverse course and again require `Send`?

Using

    lambda_http = "0.16.0"

and this code

```
use std::{cell::RefCell, sync::RwLock};
use lambda_http::{service_fn, Body, Error, Request, Response};
use log::debug;

#[derive(Debug)]
struct ServerContext {
    _something: RefCell<()>    
}

async fn handler(server_ctx: &ServerContext, request: Request) -> Result<Response<Body>, Error> {
    debug!("server_ctx: {:?}, request: {:?}", server_ctx, request);

    let response = Response::builder()
        .status(200)
        .header("Content-Type", "text/plain")
        .body("Hello World!".into())
        .map_err(Box::new)?;

    Ok(response)
}

#[tokio::main]
async fn main() -> Result<(), Error> {
[...]

PR #344, opened 2021-09-08
Remove Send bound in http's Handler

fmonniot:
Follow up to #342. This PR remove the `Send` bound on `lambda_http`'s `Handler::Fut` type parameter.


coltonweaver (approved):
LGTM, thanks for making these changes!

PR #356, opened 2021-10-27
make handler reference mutable

Mdrbhatti:
*Issue #, if available:*
N/A

*Description of changes:*
this change will allow calling a handler that implements tower/hyper `Service` trait. 

https://docs.rs/hyper/0.14.14/hyper/service/trait.Service.html
https://docs.rs/tower-service/0.3.1/tower_service/trait.Service.html


coltonweaver:
I agree with David in this, as it seems to make sense to try and make this a bit more ergonomic.

@rimutaka What are your thoughts on the matter?

coltonweaver (approved):
Approving. We can revisit if this turns out to be a contentious decision

coltonweaver (approved):
Approved. Can revisit if this is a contentious decision.

PR #363, opened 2021-11-11
feat: inject user agent in Lambda runtime API calls

nmoutschen:
*Issue #, if available:*

*Description of changes:*

This injects a User-Agent header into the Lambda runtime API calls to track usage of Rust for provided/provided.al2 Lambda runtimes.

This uses `aws-lambda-rust/%s`, with __%s__ being the crate version, to mimick the behaviour of other runtimes. E.g. for [Go](https://github.com/aws/aws-lambda-go/blob/main/lambda/runtime_api_client.go#L40), [Java RIC](https://github.com/aws/aws-lambda-java-libs/pull/277). This is slightly different as we cannot get the rustc version directly from the environment variables set by Cargo.

### DIFF
--- a/lambda-runtime/src/requests.rs
+++ b/lambda-runtime/src/requests.rs
@@ -20,7 +20,7 @@ impl IntoRequest for NextEventRequest {
     fn into_req(self) -> Result<Request<Body>, Error> {
         let req = Request::builder()
             .method(Method::GET)
-            .header("User-Agent", format!("aws-lambda-rust/{}", env!("CARGO_PKG_VERSION")))
+            .header("User-Agent", ["aws-lambda-rust/", env!("CARGO_PKG_VERSION")].concat())
             .uri(Uri::from_static("/2018-06-01/runtime/invocation/next"))
             .body(Body::empty())?;
         Ok(req)
@@ -59,7 +59,7 @@ fn test_next_event_request() {
     assert_eq!(req.method(), Method::GET);
     assert_eq!(req.uri(), &Uri::from_static("/2018-06-01/runtime/invocation/next"));
     assert!(match req.headers().get("User-Agent") {
-        Some(header) => header.to_str().unwrap() == format!("aws-lambda-rust/{}", env!("CARGO_PKG_VERSION")),
+        Some(header) => header.to_str().unwrap() == ["aws-lambda-rust/", env!("CARGO_PKG_VERSION")].concat(),
         None => false,
     });
 }
@@ -81,7 +81,7 @@ where
         let body = Body::from(body);
 
         let req = Request::builder()
-            .header("User-Agent", format!("aws-lambda-rust/{}", env!("CARGO_PKG_VERSION")))
+            .header("User-Agent", ["aws-lambda-rust/", env!("CARGO_PKG_VERSION")].concat())
             .method(Method::POST)
             .uri(uri)
             .body(body)?;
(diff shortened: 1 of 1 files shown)


lambda-runtime/src/requests.rs
```diff
@@ -20,6 +20,7 @@ impl IntoRequest for NextEventRequest {
     fn into_req(self) -> Result<Request<Body>, Error> {
         let req = Request::builder()
             .method(Method::GET)
+            .header("User-Agent", format!("aws-lambda-rust/{}", env!("CARGO_PKG_VERSION")))
```
rimutaka, line 23: `format!()` is 3 times slower than `[].concat()`. See https://github.com/hoodie/concatenation_benchmarks-rs. I've just re-run the benches on the latest nightly and it gave me 23ns vs 79ns on my laptop.

nmoutschen: Thanks! I've changed it to `["aws-lambda-rust/", env!("CARGO_PKG_VERSION")].concat()`.

Question from me: since both of these values are known at compile time, wouldn't the compiler already optimize them to prevent concatenation at run time?

SebastianEdwards:
@nmoutschen anything those of us using custom Rust runtimes (not this crate) can do to signal our usage and support of Rust to AWS?

nmoutschen:
@SebastianEdwards If you use a similar pattern (<runtime name>/<version>) for the User-Agent, and you let me know what pattern you'll use, we can track it. 😁 I'll just ask that you don't prefix your runtime name with "aws-lambda".

E.g. if you made a runtime called "sebs-awesome-runtime", you can pass a "sebs-awesome-runtime/0.1.0" User-Agent to the Lambda runtime API.

SebastianEdwards:
> E.g. if you made a runtime called "sebs-awesome-runtime", you can pass a "sebs-awesome-runtime/0.1.0" User-Agent to the Lambda runtime API.

Haha, alright, I guess the user-agent is just going to have to be "sebs-awesome-runtime" then. Will add the header next week

simonvandel (changes requested):
You can avoid allocations for `User-Agent` by 
defining it once as a const:
```rust
const USER_AGENT: &'static str = concat!("aws-lambda-rust/", env!("CARGO_PKG_VERSION"));
```

See https://godbolt.org/z/8fPdxWKrn for a comparison on the generated ASM.

lambda-runtime/src/requests.rs
coltonweaver: Hey! I actually ran into this recently in another project. I think it comes down to `concat` being able to call through to an implementation that is specifically optimized for combining the two `&str` inputs. `format!` doesn't have this due to the need for being extremely flexible.

coltonweaver:
> You can avoid allocations for `User-Agent` by defining it once as a const:
> 
> ```rust
> const USER_AGENT: &'static str = concat!("aws-lambda-rust/", env!("CARGO_PKG_VERSION"));
> ```
> 
> See https://godbolt.org/z/8fPdxWKrn for a comparison on the generated ASM.

Since we're planning on setting this by default on all requests we should 

> You can avoid allocations for `User-Agent` by defining it once as a const:
> 
> ```rust
> const USER_AGENT: &'static str = concat!("aws-lambda-rust/", env!("CARGO_PKG_VERSION"));
> ```
> 
> See https://godbolt.org/z/8fPdxWKrn for a comparison on the generated ASM.

Change overall looks good to me. I agree with @simonvandel that we could probably get away with making this a const using the above. This way we avoid all of the additional allocations on request and response. Thoughts @nmoutschen?

nmoutschen:
Sounds good! I'll make the changes tomorrow morning.

coltonweaver (approved):
This looks good. Thanks for addressing the feedback!

Issue #365, opened 2021-11-16
Data did not match any variant of untagged enum LambdaRequest with SAM local


nmoutschen:
When running a Lambda function in Rust using the `lambda-http` crate with `sam local start-api` and an HttpApi event, the Lambda runtime is not able to deserialize the payload sent by SAM into a LambdaRequest.

I've created a minimum reproducible sample that just takes an event received by a Lambda function running locally in another language, and try to deserialize it into a `lambda_http::request::LambdaRequest`:

```rust
const PAYLOAD: &str = r#"{
    "version": "2.0",
    "routeKey": "GET /hello",
    "rawPath": "/hello",
    "rawQueryString": "",
    "cookies": [],
    "headers": {
      "Host": "127.0.0.1:3000",
      "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:94.0) Gecko/20100101 Firefox/94.0",
      "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8",
      "Accept-Language": "en-US,en;q=0.5",
      "Accept-Encoding": "gzip, deflate",
      "Connection": "keep-alive",
      "Upgrade-Insecure-Requests": "1",
      "Sec-Fetch-Dest": "document",
      "Sec-Fetch-Mode": "navigate",
      "Sec-Fetch-Site": "none",
      "Sec-Fetch-User": "?1",
      "Cache-Control": "max-age=0",
      "X-Forwarded-Proto": "http",
      "X-Forwarded-Port": "3000"
    },
    "queryStringParameters": {},
    "requestContext": {
      "accountId": "123456789012",
      "apiId": "1234567890",
      "http": {
        "method": "GET",
        "path": "/hello",
        "protocol": "HTTP/1.1",
        "sourceIp": "127.0.0.1",
        "userAgent": "Custom User Agent String"
      },
      "requestId": "1ac06eee-f687-44ec-9036-dfd49d0be0a3",
      "routeKey": "GET /hello",
      "stage": "$default",
      "time": "16/Nov/2021:11:54:33 +0000",
      "timeEpoch": 1637063673,
      "domainName": "localhost",
      "domainPrefix": "localhost"
    },
    "body": "",
    "pathParameters": {},
    "stageVariables": null,
    "isBase64Encoded": false
  }"#;

fn main() {
[...]

PR #366, opened 2021-11-16
fix: null deserialization of stageVariables for sam local

nmoutschen:
*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/365

*Description of changes:*

Add support for deserializing API Gateway v2 payloads when `stageVariables` is set to null.

AWS SAM local sends a payload that contains a null value for `stageVariables`, which causes the deserialization into a `lambda_http::request::LambdaRequest` to fail. This makes this field deserializable using the `nullable_default` function to handle that edge case.

__Open question__: should we also make this change to all fields that deserializes into a `StrMap`?

### DIFF
--- a/lambda-http/src/request.rs
+++ b/lambda-http/src/request.rs
@@ -33,9 +33,9 @@ pub enum LambdaRequest<'a> {
         cookies: Option<Vec<Cow<'a, str>>>,
         #[serde(deserialize_with = "deserialize_headers")]
         headers: http::HeaderMap,
-        #[serde(default)]
+        #[serde(default, deserialize_with = "nullable_default")]
         query_string_parameters: StrMap,
-        #[serde(default)]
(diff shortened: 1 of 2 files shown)


coltonweaver:
> Open question: should we also make this change to all fields that deserializes into a StrMap?

In favor of your robustness comment in the issue related to this, I think it makes sense to do this. Overall the change looks good. Do you want to include this in this PR? I'll approve but feel free to push up the change to do that.

coltonweaver (approved):

nmoutschen:
> In favor of your robustness comment in the issue related to this, I think it makes sense to do this. Overall the change looks good. Do you want to include this in this PR? I'll approve but feel free to push up the change to do that.

Will do - better to do it in one go. I'll let you review before merging to get another pair of eyes. :)

coltonweaver:
Looks good, thanks!

Issue #368, opened 2021-11-19
failed to build request: http::Error(InvalidUri(InvalidFormat)) when testing from API Gateway in the AWS Console


ionionascu:
When testing from the API Gateway's test method without explicitely providing the `Host` http header, I get the error `failed to build request: http::Error(InvalidUri(InvalidFormat))`, which I didn't use to get (maybe the test request has been changed on AWS side).

This test reproduces the error on the current master branch:

```rust
#[test]
fn test() {
    const PAYLOAD: &str = r#"{
        "resource": "/test",
        "path": "/test",
        "httpMethod": "POST",
        "headers": {
            "test": "test"
        },
        "multiValueHeaders": {
            "test": [
                "test"
            ]
        },
        "queryStringParameters": {
            "test": "test"
        },
        "multiValueQueryStringParameters": {
            "test": [
                "test"
            ]
        },
        "pathParameters": null,
        "stageVariables": null,
        "requestContext": {
            "resourceId": "resourceId",
            "resourcePath": "/test",
            "operationName": "lambda",
            "httpMethod": "POST",
            "requestTime": "19/Nov/2021:12:19:02 +0000",
            "path": "/test",
            "accountId": "123",
            "protocol": "HTTP/1.1",
            "stage": "test-invoke-stage",
            "domainPrefix": "testPrefix",
            "requestTimeEpoch": 1637324342768,
            "requestId": "123",
            "identity": {
                "cognitoIdentityPoolId": null,
                "cognitoIdentityId": null,
                "apiKey": "test-invoke-api-key",
                "principalOrgId": null,
                "cognitoAuthenticationType": null,
                "userArn": "arn",
                "apiKeyId": "test-invoke-api-key-id",
                "userAgent": "aws-internal/3 aws-sdk-java/1.12.76 Linux/5.4.147-83.259.amzn2int.x86_64 OpenJDK_64-Bit_Server_VM/25.302-b08 java/1.8.0_302 vendor/Oracle_Corporation cfg/retry-mode/standard",
                "accountId": "123",
                "caller": "caller",
                "sourceIp": "test-invoke-source-ip",
                "accessKey": "accessKey",
                "cognitoAuthenticationProvider": null,
                "user": "user"
            },
            "domainName": "testPrefix.testDomainName",
            "apiId": "apiId"
        },
        "body": null,
        "isBase64Encoded": false
    }"#;
    
[...]

PR #369, opened 2021-11-19
fix: create a valid URL for REST APIs and ALBs when the host header i…

nmoutschen:
…s missing

*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/368

*Description of changes:*

When the __Host__ header is missing from an API Gateway REST API or ALB request, the current implementations defaults to an empty value. This means that the `lambda_http` crate tries to transform a value of `https:///my/path` into an Uri, which results in an `http::Error(InvalidUri(InvalidFormat))` error.

This PR changes the behavior by checking first if there is a valid host header present. It then uses the `http(s)://my-host/my/path` pattern if it is present, or just `/my/path` otherwise.

No changes are needed for API Gateway v2 since it adds a `requestContext.domainName` value, which means we can always derive a valid domain name in that situation.

### DIFF
--- a/lambda-http/src/request.rs
+++ b/lambda-http/src/request.rs
@@ -440,18 +440,23 @@ impl<'a> From<LambdaRequest<'a>> for http::Request<Body> {
                 let builder = http::Request::builder()
                     .method(http_method)
                     .uri({
-                        format!(
-                            "{}://{}{}",
-                            headers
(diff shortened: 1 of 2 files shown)


coltonweaver (approved):
LGTM, thanks for fixing!

PR #370, opened 2021-11-21
Removed references to type Error aliases in docs

rimutaka:
We removed the need to define an alias for our internal Error type in this commit https://github.com/awslabs/aws-lambda-rust-runtime/commit/96fc5bce8543b42339c22f5c3ed9f83bfabd3991 a while back. I've just noticed that some internal docs didn't get updated.

### Description of changes

* removed alias definition
* added Error to `use`

I didn't run the examples, but this is a 1:1 copy from my working code.


coltonweaver (approved):
Good catch, thanks for the update!

Issue #367, opened 2021-11-19
Data did not match any variant of untagged enum LambdaRequest when testing from API Gateway in the AWS Console


ionionascu:
I have a Rust lambda that is accessed through an API Gateway with proxy integration. When I test the API Gateway's method directly in the AWS Console without passing any headers (or no query parameters), I get an error from the Lambda: `Error: Error("data did not match any variant of untagged enum LambdaRequest", line: 0, column: 0)`

This seems to be caused by the fact that in the API Gateway JSON request, the `headers`, `multiValueHeaders`, `queryStringParameters` and `multiValueQueryStringParameters` elements are passed as `null` (I do not remember seeing this error before so it may be that this used to be passed as an empty object, `{}`).

Using a modified sample request JSON from a different issue, I can reproduce the error with this test:

```rust
#[test]
fn test() {
    const PAYLOAD: &str = r#"{
        "version": "2.0",
        "routeKey": "GET /hello",
        "rawPath": "/hello",
        "rawQueryString": "",
        "cookies": [],

        "headers": null,
        "multiValueHeaders": null,
        "queryStringParameters": null,
        "multiValueQueryStringParameters": null,

        "requestContext": {
            "accountId": "123456789012",
            "apiId": "1234567890",
            "http": {
                "method": "GET",
                "path": "/hello",
                "protocol": "HTTP/1.1",
                "sourceIp": "127.0.0.1",
                "userAgent": "Custom User Agent String"
            },
            "requestId": "1ac06eee-f687-44ec-9036-dfd49d0be0a3",
            "routeKey": "GET /hello",
            "stage": "$default",
            "time": "16/Nov/2021:11:54:33 +0000",
            "timeEpoch": 1637063673,
            "domainName": "localhost",
            "domainPrefix": "localhost"
        },
        "body": "",
        "pathParameters": {},
        "stageVariables": null,
        "isBase64Encoded": false
    }"#;
    
    let val: lambda_http::request::LambdaRequest = serde_json::from_str(PAYLOAD).unwrap();
    println!("{:?}", val);
}
```

I tested with the current master branch (commit `576f6013dd188de08928dd503170b525bc17c26e`), which contains the changes from another issue similar to this one (#365).

nmoutschen:
Hey @ionionascu!

[...]

PR #371, opened 2021-11-21
fix: add support for null/invalid type for headers

nmoutschen:
*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/367

*Description of changes:*

This adds support for invalid/null value for headers, defaulting to an empty HeaderMap if need be, increasing resilience if the Lambda function would get an invalid value for headers.

I've also added additional checks in the request deserialization tests to ensure we deserialize from the right type of API, as making deserialization more permissive could potentially break this.

❓  __Open question before merging this__: this change is overly permissive. Should I scope this down to only accepting null and map values?

### DIFF
--- a/lambda-http/src/request.rs
+++ b/lambda-http/src/request.rs
@@ -352,7 +352,7 @@ where
         }
     }
 
-    deserializer.deserialize_map(HeaderVisitor)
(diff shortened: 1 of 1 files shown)


coltonweaver (approved):
LGTM! Thanks for doing this.

WRT your overly permissive question - I think it's fine to do things like this in lambda_http.

Issue #374, opened 2021-11-26
Replace lambda_runtime::Handler with tower::Service?


nmoutschen:
Hey folks!

After looking at [how to wrap `lambda_runtime::Handler`](https://github.com/awslabs/aws-lambda-rust-runtime/issues/373) compared to tower's [layers](https://docs.rs/tower/0.4.11/tower/trait.Layer.html) (thanks to @davidbarsky), and @Mdrbhatti's recent [PR on making self mutable for `Handler`](https://github.com/awslabs/aws-lambda-rust-runtime/pull/356), I'm wondering if we wouldn't benefit from using `tower::Service` instead.

The signatures today are pretty close already, with the two differences between the addition of a `Response` type and `poll_ready` method.

## Why should we do this?

We would then benefit from being able to reuse existing (and well-tested) libraries on that side. For me, I'd like to implement a few tools to help with Rust functions on Lambda similar to what we've done for [Python](https://github.com/awslabs/aws-lambda-powertools-python) and [Java](https://github.com/awslabs/aws-lambda-powertools-java), such as creating structured logs, handling tracing, etc. Being able to use layers to wrap the function itself would be quite helpful there.

It could make some of the existing tools such as [warp_lambda](https://github.com/Mdrbhatti/warp_lambda) or [poem-lambda](https://github.com/poem-web/poem/tree/master/poem-lambda) simpler too.

## Why shouldn't we do this?

I could write my own implementation of layers for my use case. However, before committing to a specific implementation, I want to double-check we're not missing on something by __not__ doing this.

This is also a pretty significant change for the libraries that depends on [lambda-runtime](https://crates.io/crates/lambda-runtime/reverse_dependencies) or [lambda-http](https://crates.io/crates/lambda-http/reverse_dependencies). However, We could minimize the impact for most users of this runtime since they usually interact only with `lambda_runtime::run`, `lambda_runtime::handler_fn`, or `lambda_http::handler`. We could keep the same function names but change their signatures.

## What would the experience look like?

The least impactful change would be to implement `tower::Service` for `HandlerFn`, and make `run` accept a service.

We'll probably need to create a new type that would handle both the event payload and context at the same time, as `Handler.call` take two arguments while `tower::Service.call` only takes one.

If people are interested in this, I'm happy to create a proof of concept.

davidbarsky:
[...]

Issue #372, opened 2021-11-21
Caching data between invocations


rimutaka:
It is possible to initialize a resource in `main` and pass it onto a closure inside `handler_fn` as a shared reference. Example: https://github.com/awslabs/aws-lambda-rust-runtime/blob/master/lambda-runtime/examples/shared_resource.rs

The downside of this method is that the initialization must be done upfront and the resource cannot be updated.

I've been trying to implement a cache that can be updated from inside the handler, but no matter what I tried it's a breaking change.

**Does anyone know a way of updating a shared resource from inside the handler fn without breaking the current handler_fn interface?**

softprops:
[hander_fn](https://github.com/awslabs/aws-lambda-rust-runtime/blob/154cf312a0b4d7ffec2fa035695151106a44f286/lambda-runtime/src/lib.rs#L84-L86) is a factory method for creating an instance of a type which [implements](https://github.com/awslabs/aws-lambda-rust-runtime/blob/154cf312a0b4d7ffec2fa035695151106a44f286/lambda-runtime/src/lib.rs#L96-L107) a generic [Handler trait](https://github.com/awslabs/aws-lambda-rust-runtime/blob/154cf312a0b4d7ffec2fa035695151106a44f286/lambda-runtime/src/lib.rs#L72-L79). 

The  `run` function accepts any type which implements `Handler` so I believe you can just have a type with state that lives across invocations of a lambda task that implements `Handler`

example pseudo code

```rust
struct Counter {
  data: usize
} 

impl Default for Counter {
  fn default() -> Self {
     Counter { data: 0 }
  }
} 

impl Handler<A, B> for Counter {
 type Error = YourErrType;
 type Fut = YourFutType;
 fn call(&mut self, req: A, ctx: Context) -> Self::Fut {
    // update your state
    self.data += 1;
    // do the thing that returns Self::Fut
 }   
}

fn main() {
  lambda_runtime::run(Counter::default()).await?;
  Ok(())
}
```

PR #375, opened 2021-11-27
[WIP] feat!: Replace Handler with tower::Service

nmoutschen:
*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/374

*Description of changes:*

__This PR is meant to compare the current implementation with `Handler` with using `tower::Service` and should not be merged as-is. See the discussion in https://github.com/awslabs/aws-lambda-rust-runtime/issues/374 for more information.__

List of changes:

* Removes `Handler` and `HandlerFn`.
* Makes `handler_fn`  return a `ServiceFn` instead.
* Adds a new `LambdaRequest` type to wrap request and context in a single struct.
* Changes `run` signature to take a `Service<LambdaRequest<A>>` instead.

__BREAKING CHANGE__: this removes the `Handler` trait, which could be used by existing libraries. However, developers that are using `lambda_runtime::run` and `lambda_runtime::handler_fn` should not be impacted. The closure taken by `HandlerFn` remains the same, so developers can keep using function with separate event and context parameters.

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -17,7 +17,7 @@ use std::{
 };
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_stream::{Stream, StreamExt};
-use tower_service::Service;
+pub use tower_service::Service;
 use tracing::{error, trace};
 
 mod client;
@@ -29,6 +29,8 @@ mod types;
 
 use requests::{EventCompletionRequest, EventErrorRequest, IntoRequest, NextEventRequest};
 use types::Diagnostic;
+
+pub use types::LambdaRequest;
 
 /// Error type that lambdas may result in
 pub type Error = Box<dyn std::error::Error + Send + Sync + 'static>;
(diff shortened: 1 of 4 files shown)


lambda-runtime/src/lib.rs
```diff
@@ -68,42 +72,17 @@ impl Config {
...
-
-/// Returns a new [`HandlerFn`] with the given closure.
-///
-/// [`HandlerFn`]: struct.HandlerFn.html
-pub fn handler_fn<F>(f: F) -> HandlerFn<F> {
-    HandlerFn { f }
+/// Wraps a function that takes 2 arguments into one that only takes a [`LambdaRequest`].
+fn handler_wrapper<A, Fut>(f: impl Fn(A, Context) -> Fut) -> impl Fn(LambdaRequest<A>) -> Fut {
+    move |req| f(req.event, req.context)
 }
```
davidbarsky, line 78: Is this intentionally not public?

lambda-runtime/src/types.rs
```diff
@@ -148,6 +148,15 @@ impl TryFrom<HeaderMap> for Context {
     }
 }
 
+/// Incoming Lambda request containing the event payload and context.
+#[derive(Clone, Debug)]
+pub struct LambdaRequest<T> {
+    /// Event payload.
+    pub event: T,
+    /// Invocation context.
+    pub context: Context,
```
davidbarsky, line 157: Nit: I'd make these fields private and only expose them via getters to not make adding fields a semver-breaking change.

lambda-runtime/src/lib.rs
nmoutschen: Yes! My thinking here was to only use this function as part of `handler_fn` to help wrap a function that takes two arguments.

lambda-runtime/src/types.rs
nmoutschen: Good point!

For reference if anyone else reads this, I wanted to dig into why getters would be a good idea here and found [this message on the Rust forum](https://users.rust-lang.org/t/public-getter-method-vs-pub-field/20147/2).

That said, given that the only benefit of getters here would be preventing breaking changes, isn't it better to use [`#[non_exhaustive]`](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) instead?

lambda-runtime/src/lib.rs
davidbarsky: gotcha. for the public interface, i think i'd personally try to use (and-reexport) `tower::service_fn`.

lambda-runtime/src/types.rs
davidbarsky: > That said, given that the only benefit of getters here would be preventing breaking changes, isn't it better to use #[non_exhaustive] instead?

`#[non_exhaustive]` also makes it impossible for the `Context` struct to be constructed in e.g., test code. i'd consider creating a builder for this, but i need to remember the standard behavior.

rimutaka: I may be confusing what is being discussed. Feel free to ignore ...
I've been trying to add a cache (#372) that can be managed by the handler and preserved by the runtime between invocations. I wonder how it may work with this change.

My understanding is that &mut is not possible due to `move`, so we need to pass the ownership of the cache to the handler, get it back with the response, preserve until the next invocation and then pass onto the handler again. The type of the cache is unknown. It will be up to the handler_fn implementers to define it.

Would it be a good idea to have something like this?
```rust
pub struct LambdaRequest<T,Q> {
    /// Event payload.
    pub event: T,
    /// Invocation context.
    pub context: Context,
    /// User-defined cache
    pub cache: Q,
}
```
`Q` would probably need `Send` bound, but otherwise keep it as unconstrained as possible.

nmoutschen: Hey @rimutaka ! Thanks for taking the time to look at this. 😄

With this change, the example from @softprops would not directly, but you could make your own `tower::Service` or `tower::Layer`. Basically, there would be 2 interfaces for creating Lambda functions:

* A high-level interface that works in the same way as other languages on Lambda (with `handler_fn` and functions that take 2 arguments (event and context);
* A low-level interface that takes a `Service<LambdaRequest<A>>`, where you can adopt the same strategy as the comment you linked, but substitute `Handler<A, B>` with `Service<LambdaRequest<A>>`.

That said, for your specific use-case, you could also use the high-level interface if [you're OK with using a `RefCell`](https://gist.github.com/nmoutschen/521d8cee05b0a5c4ea019b57672efb73).

nmoutschen: Actually, taking a step back, I think the `LambdaRequest` could work slightly differently. I'm not too keen on the context key, so we could have an [extensions pattern similar to `http::Request`](https://docs.rs/http/0.2.5/http/request/struct.Request.html#method.extensions).

Then from there, you could provide `tower::Layer`s that would inject values into the request that you could retrieve later. We'd inject the `Context` by default, but you could use that to inject a cache for example, and retrieve it with `request::extensions_mut::<MyCache>()`.

rimutaka: Extensions would be a nice addition to the current option of implementing Handler trait. I am concerned we may overcomplicate the user interface, though. Just a thought.

nmoutschen:
Closing this one since this was just for experimentation purposes. I'll create a new one for real this time.

PR #376, opened 2021-11-29
Lambda Extensions

### LINKED ISSUE #261: Rust library for the Lambda Extensions API?
AWS Lambda now support [extensions](https://aws.amazon.com/blogs/compute/introducing-aws-lambda-extensions-in-preview/). To improve the developer experience, it'd be nice to have a library similar to this but that handle the [Extensions API specific calls](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html).

It'd be nice to have something that works in the same way as the `lambda-runtime` but for the extension API, so that people could use Rust for creating Lambda extensions.

calavera:
*Issue #, if available:*

Fixes #261 

*Description of changes:*

This PR introduces two new creates into this project to be able to implement Lambda Extensions in Rust. The work is not completed, and I wanted to share it with you all now that I have functional examples to get people's feedback on the direction I'm taking.

The first crate is `lambda-runtime-client`. I've extracted this crate from the current runtime. The idea with this is to be able to reuse clients for different APIs. This brings some guarantees about client behaviors, like having a User-Agent header by default.

The second crate is `lambda-extension`. This crate implements the [Lambda Extensions API,](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-lifecycle-invoke) which is the main point of this PR. I've included two examples in this new crate so I can get feedback about the API exposed to developers that want to create extensions in Rust.

[Basic Example](https://github.com/awslabs/aws-lambda-rust-runtime/pull/376/files#diff-73785481e24e1f7e116854203cd21288b4f0d14ef8d958f0796f00518215583d)
[Example that subscribes to specific events](https://github.com/awslabs/aws-lambda-rust-runtime/pull/376/files#diff-7109631d88180a061dcc1085576cbd38987753568c6d5657936873ec5d535935)

There are a few things missing to get this PR in a mergeable state, mostly documentation, but I'll appreciate your input sooner rather than later so I can make these changes better.


nmoutschen:
Thanks a lot for taking the time to create this PR @calavera ! I'll take a look at it tomorrow. 😄

lambda-extension/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+authors = ["David Calavera <david.calavera@gmail.com>"]
+description = "AWS Lambda Extension API"
+license = "Apache-2.0"
+repository = "https://github.com/awslabs/aws-lambda-rust-runtime"
+categories = ["web-programming::http-server"]
+keywords = ["AWS", "Lambda", "API"]
+readme = "../README.md"
+
+[dependencies]
+anyhow = "1.0.48"
```
rimutaka, line 14: We dropped `anyhow` a while ago.

calavera: oh yeah, it's not even used. I forgot to remove it. Thanks for pointing that out.

lambda-extension/examples/error_handling.rs
```diff
@@ -0,0 +1,55 @@
...
+
+struct ErrorHandlingExtension {
+    client: Client
+}
+
+#[async_trait::async_trait]
+impl Extension for ErrorHandlingExtension {
+    async fn on_invoke(&self, extension_id: &str, _event: InvokeEvent) -> Result<(), Error> {
+        let err = ErrorExample::OnInvokeError;
+        let req = init_error(extension_id, &format!("{}", err), None)?;
```
nmoutschen, line 31: Unsure if this is the right way to handle it. Ideally, we should probably abstract errors away from the developers, through the `Err`, and map it manually within the `run` function.

If we get an error after register, we should automatically send a call to `/extension/init/error`, or `/extension/exit/error` if it's after a shutdown call.

lambda-extension/src/lib.rs
```diff
@@ -0,0 +1,129 @@
...
+            let event = event?;
+            let (_parts, body) = event.into_parts();
+
+            let body = hyper::body::to_bytes(body).await?;
+            trace!("{}", std::str::from_utf8(&body)?); // this may be very verbose
+            let event: NextEvent = serde_json::from_slice(&body)?;
+
+            match event {
+                NextEvent::Invoke(event) => {
+                    extension.on_invoke(extension_id, event).await?;
```
nmoutschen, line 91: Continuing on the question of error handling, we should probably send a call back to the Extension API at this stage, and then exit.

Reference from the [Lambda documentation](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-init-error):

> Call it when the extension fails to initialize after it has registered. After Lambda receives the error, no further API calls succeed. The extension should exit after it receives the response from Lambda.

nmoutschen (changes requested):
Hey @calavera!

Thanks a lot for the time and efforts in creating this PR! Overall it's pretty solid and I'm for the idea of splitting the API client into its separate crate.

In term of potential improvements before merging this, there are three main points for me:

* I'm not too keen on adding a dependency on `async_trait`, as it would diverge from how the `Handler` trait works at the moment.
* I think we should push error handling under the hood as much as possible by catching the `Err`s and sending the right API call on behalf of the developer.
* Right now, `lambda-extension` listens to both INVOKE and SHUTDOWN events. While doing an Ok(()) is not a big deal, I think it would be nice to be able to configure what events to listen to, and then get an enum as payload.

Cargo.toml
```diff
@@ -1,5 +1,7 @@
 [workspace]
 members = [
     "lambda-http",
-    "lambda-runtime"
+    "lambda-runtime-client",
```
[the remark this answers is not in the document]
calavera: I forgot about that. Do you have a name suggestion?

lambda-extension/src/requests.rs
```diff
@@ -0,0 +1,78 @@
...
+        .method(Method::GET)
+        .header(EXTENSION_ID_HEADER, extension_id)
+        .uri("/2020-01-01/extension/event/next")
+        .body(Body::empty())?;
+    Ok(req)
+}
+
+pub(crate) fn register_request(extension_name: &str) -> Result<Request<Body>, Error> {
+    let events = serde_json::json!({
+        "events": ["INVOKE", "SHUTDOWN"]
```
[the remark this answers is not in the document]
calavera: let me think how that could look like

Cargo.toml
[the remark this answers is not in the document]
davidbarsky: i would probably resurrect that crate name and use it for this. happy to transfer it over, need be.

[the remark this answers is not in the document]
nmoutschen: Yep, I think we could just keep that one there.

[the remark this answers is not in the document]
calavera: I've renamed the crate `lambda-runtime-api-client` to avoid confusion. I've also updated the version to keep it in sync with the `lambda-runtime` crate.

calavera:
@nmoutschen I've updated this PR to address your comments. Can you take a look at the examples now and let me know what you think?

lambda-extension/src/lib.rs
```diff
@@ -44,19 +45,61 @@ pub enum NextEvent {
...
+
 /// A trait describing an asynchronous extension.
-#[async_trait]
 pub trait Extension {
-    async fn on_invoke(&self, extension_id: &str, event: InvokeEvent) -> Result<(), Error>;
-    async fn on_shutdown(&self, extension_id: &str, event: ShutdownEvent) -> Result<(), Error>;
+    /// Response of this Extension.
+    type Fut: Future<Output = Result<(), Error>>;
+    /// Handle the incoming event.
+    fn call(&self, extension_id: ExtensionId, event: NextEvent) -> Self::Fut;
```
nmoutschen, line 62: 1. I think it would be nice to have `&mut self` here to match more closely the current implementation in the repository (but not yet published). See [this PR](https://github.com/awslabs/aws-lambda-rust-runtime/pull/356) for reference.
2. __NITPICK__: Another remark, but more of a nitpick from me: is there a good use-case to expose the extension ID, or could we keep that internal to the crate?
3. __NITPICK__: Would it be also better to swap the order of arguments here since `event` would be the most used one?

lambda-extension/examples/basic.rs
```diff
@@ -21,6 +20,6 @@ async fn main() -> Result<(), Error> {
     // can be replaced with any other method of initializing `log`
     SimpleLogger::new().with_level(LevelFilter::Info).init().unwrap();
 
-    run(BasicExtension {}).await?;
-    Ok(())
+    let func = extension_fn(my_extension);
```
nmoutschen, line 23: Love the high-level experience here. 😄

lambda-extension/examples/custom_events.rs
```diff
@@ -0,0 +1,30 @@
...
+            // ignore any other event
+            // because we've registered the extension
+            // only to receive SHUTDOWN events
+        }
+    }
+    Ok(())
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
```
nmoutschen, line 20: __OPEN QUESTION__: should we also have an example with a struct that implements the `Extension` trait? E.g. caching/storing data.

nmoutschen (changes requested):
Thanks a lot for the update @calavera!

I really like the changes you made! The main change I think would help here before we merge this is on the `Extension` trait signature, by making self mutable to be in line with recent changes in this repository (but not published yet).

I also have some other small remarks/nitpicks, but up to you to decide if you agree with me or not.

lambda-extension/examples/custom_events.rs
calavera: good idea, I'll add one

lambda-extension/src/lib.rs
calavera: you might be right. The extension id is there because the error notifications where send from the user code before. I'll remove it, I rather keep the api simple, and expand it later if necessary.

lambda-extension/examples/custom_trait_implementation.rs
```diff
@@ -0,0 +1,34 @@
+use lambda_extension::{run, Error, NextEvent, Extension};
+use log::LevelFilter;
+use simple_logger::SimpleLogger;
+use std::future::{Future, ready};
+use std::pin::Pin;
+
+struct MyExtension {}
```
nmoutschen, line 7: Should we showcase a trait implementation with some data in the struct?

Could be something as simple as storing all the events coming in and flushing them on shutdown for example.

nmoutschen (approved):
LGTM from an implementation POV!

calavera:
@nmoutschen I've added documentation where it was necessary and opened the PR for review. Thanks a lot for your feedback.

calavera:
I still have to update the README with information about this new crate.

calavera:
@nmoutschen I was not sure where to add information to the main readme, so I opted for creating two specific readmes inside the new crate directories. Let me know what you think.

lambda-extension/README.md
```diff
@@ -0,0 +1,56 @@
...
+    SimpleLogger::new().with_level(LevelFilter::Info).init().unwrap();
+
+    let func = extension_fn(log_extension);
+    lambda_extension::run(func).await
+}
+```
+
+## Deployment
+
+Lambda extensions can be added to your functions either using [Lambda layers](https://docs.aws.amazon.com/lambda/latest/dg/using-extensions.html#using-extensions-config), or adding them to [containers images](https://docs.aws.amazon.com/lambda/latest/dg/using-extensions.html#invocation-extensions-images). Regardless of how you deploy them, the extensions MUST be compiled against the same architecture that your lambda functions runs on. The only two architectures that AWS Lambda supports are `aarch64-unknown-linux-gnu` for ARM functions, and `x86_64-unknown-linux-gnu` for x86 functions.
```
nmoutschen, line 40: __Nit__: the version of glibc on Amazon Linux 1 runtimes is too old. For AL1, we need to use musl instead of glibc. If someone wants to develop an extension that will run irregardless of OS version, they could use musl.

lambda-runtime-api-client/README.md
```diff
@@ -0,0 +1,35 @@
+# AWS Lambda Runtime API Client
+
+[![Docs](https://docs.rs/lambda_runtime_api_client/badge.svg)](https://docs.rs/lambda_runtime_api_client)
+
+**`lambda-runtime-api-client`** is a library to interact with the AWS Lambda Runtime API.
```
nmoutschen, line 5: __Nit__: I'd add something about how most people probably don't need to use this library directly, and should look at `lambda_runtime` or `lambda_extension` instead.

nmoutschen (approved):
LGTM! Just 2 small details on libc for various OS versions, and on how users probably won't use the runtime client directly.

calavera:
@nmoutschen I dedided to remove the paragraph about targets explicitly. I think it was confusing and hard to explain. I'd expect that people that want to write extensions are already familiar with functions, and how the runtime works. It's still worth mentioning that the target must match the architecture that functions run on.

calavera:
@nmoutschen what are the next steps to get this PR merged?

nmoutschen:
Hey @calavera !

Not much on your side. I wanted to run a complete test by building an extension and running on Lambda before merging this just to make sure. I'm on sick leave at the moment, but I should be able to do it next week.

calavera:
> Not much on your side. I wanted to run a complete test by building an extension and running on Lambda before merging this just to make sure. I'm on sick leave at the moment, but I should be able to do it next week.

Thanks for letting me know, I wasn't sure if there was anything missing. Take care

lambda-extension/src/lib.rs
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+
+    /// Initialize and register the extension in the Extensions API
+    pub async fn register(&self) -> Result<Runtime, Error> {
+        let name = match self.extension_name {
+            Some(name) => name.into(),
+            None => {
+                let args: Vec<String> = std::env::args().collect();
+                args[0].clone()
```
nmoutschen, line 223: After doing some tests on Lambda, this causes an issue with the Lambda Extensions API. Because this value contains slashes (e.g. `/opt/extension/my-extension`), the Lambda Extension API will fail to process it correctly, and assumes that this is an internal extension.

Because of this, it will fail to register the extension if it contains/starts with (not sure which one) a slash.

However, taking the last part of the path works. E.g. instead of `/opt/extension/my-extension`, `my-extension` works.

nmoutschen:
Hey @calavera ! I just found a bug when running a sample Lambda function with a layer, but otherwise everything else works when I changed the full extension path for its basename.

lambda-extension/src/lib.rs
calavera: good catch! I've updated this code to cleanup the executable name.

calavera:
@nmoutschen I pushed a new commit that should fixed that problem 🙌

nmoutschen:
All good now! Merged the PR! Thank again for your contribution here! 🙌

PR #382, opened 2021-12-16
fix: add client context and cognito identity to runtime context

diceride:
*Issue #, if available:*

*Description of changes:*

Fixes an issue where the client context `lambda-runtime-client-context` and the cognito identity `lambda-runtime-cognito-identity` were not deserialized from the incoming request headers.

### DIFF
--- a/lambda-runtime/src/types.rs
+++ b/lambda-runtime/src/types.rs
@@ -57,15 +57,18 @@ struct MobileClientIdentity(String);
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct ClientContext {
     /// Information about the mobile application invoking the function.
+    #[serde(default)]
     pub client: ClientApplication,
     /// Custom properties attached to the mobile event context.
+    #[serde(default)]
     pub custom: HashMap<String, String>,
     /// Environment settings from the mobile client.
+    #[serde(default)]
     pub environment: HashMap<String, String>,
 }
 
 /// AWS Mobile SDK client fields.
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
+#[derive(Serialize, Deserialize, Default, Clone, Debug, PartialEq)]
 #[serde(rename_all = "camelCase")]
 pub struct ClientApplication {
     /// The mobile app installation id
@@ -119,6 +122,18 @@ pub struct Context {
 impl TryFrom<HeaderMap> for Context {
     type Error = Error;
     fn try_from(headers: HeaderMap) -> Result<Self, Self::Error> {
+        let client_context: Option<ClientContext> = if let Some(value) = headers.get("lambda-runtime-client-context") {
+            serde_json::from_str(value.to_str()?)?
+        } else {
+            None
+        };
+
+        let identity: Option<CognitoIdentity> = if let Some(value) = headers.get("lambda-runtime-cognito-identity") {
+            serde_json::from_str(value.to_str()?)?
+        } else {
+            None
+        };
+
         let ctx = Context {
             request_id: headers
                 .get("lambda-runtime-aws-request-id")
@@ -142,8 +157,11 @@ impl TryFrom<HeaderMap> for Context {
                 .unwrap_or(&HeaderValue::from_static(""))
                 .to_str()?
                 .to_owned(),
+            client_context,
+            identity,
             ..Default::default()
         };
+
         Ok(ctx)
     }
 }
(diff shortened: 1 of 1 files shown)


diceride:
The CI clippy warning is unrelated to the changes

diceride:
Rebased from master a7e17ee

diceride:
The CI clippy warning is unrelated to the changes. 

@nmoutschen I guess https://github.com/awslabs/aws-lambda-rust-runtime/pull/383 will resolve the warnings?

lambda-runtime/src/types.rs
```diff
@@ -119,6 +119,18 @@ pub struct Context {
 impl TryFrom<HeaderMap> for Context {
     type Error = Error;
     fn try_from(headers: HeaderMap) -> Result<Self, Self::Error> {
+        let client_context: Option<ClientContext> = if let Some(value) = headers.get("lambda-runtime-client-context") {
```
bahildebrand, line 122: I think we can replace both of these ifs with [Option::and](https://doc.rust-lang.org/std/option/enum.Option.html#method.and).

bahildebrand (changes requested):
Just to make sure I understand the issue. This was previously panicking if the headers were not present correct? So the addition of these `Option`s allows the deserialization to succeed if they are not present?

Also can you fix any clippy issues?

lambda-runtime/src/types.rs
bahildebrand: Actually should probably be [Option::and_then](https://doc.rust-lang.org/std/option/enum.Option.html#method.and_then)

lambda-runtime/src/types.rs
```diff
@@ -57,11 +57,11 @@ struct MobileClientIdentity(String);
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct ClientContext {
     /// Information about the mobile application invoking the function.
-    pub client: ClientApplication,
+    pub client: Option<ClientApplication>,
     /// Custom properties attached to the mobile event context.
-    pub custom: HashMap<String, String>,
+    pub custom: Option<HashMap<String, String>>,
```
calavera, line 62: I would recommend to deserialize `custom` and `environment` with their default representation, instead of wrapping them with an option. This is a personal preference but it makes the code easier to manage if you can assume that they always exist. You can do that by adding `#[serde(default)]` above those two fields.

bahildebrand: I think I'm in agreement that this is the better route here. Using `Option`s here will just result in a lot of extra code for upstream users, and if we can avoid that it would be preferable.

lambda-runtime/src/types.rs
bahildebrand: Not necessary if we go with @calavera suggestion. See his comment below.

davidbarsky:
This isn't really my project to maintain anymore, but as the person who made this decision a few years ago, I think it'd be helpful if I provided a bit of context. At the time—and I believe this is still the case—the client context and Cognito Identity were only filled by the AWS Mobile SDKs that _directly_ invoked Lambda functions. This pattern was discouraged in favor of delegating authentication/authorization to API Gateway, and the thinking was that anyone who'd be using this new runtime will _probably_ not be using those older SDKs.

However, I don't know if this guidance has changed and someone a bit closer to those decisions should weigh in.

lambda-runtime/src/types.rs
calavera: this is also a breaking change without much reasoning behind it.

diceride: Oh yes, I like it, it's a very elegant solution.

lambda-runtime/src/types.rs
diceride: Initially, I took the Node.js implementation as a reference, where `custom` and `environment` can be of type `undefined`.

diceride: I took the Node.js implementation of the AWS Lambda runtime as a reference, where `custom` and `environment` can be of type `undefined`.

Example

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L114

and

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L159

diceride:
@davidbarsky Alright, I understand now. Thank you. We use the `custom` property of the client context to transfer metadata to the client. We don't use the Mobile SDKs. The client context seemed to be universal in nature for us, so we were like .. "Hey! There is something that we can use to solve our problem". I noticed that it works in the Node.js client and the .NET client, but not in the Rust client.

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L114-L121

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/RuntimeApiHeaders.cs#L33-L34

diceride:
@davidbarsky Alright, I understand now. Thank you. We use the `custom` property of the client context to transfer metadata to the client. We don't use the Mobile SDKs. The client context seemed to be universal in nature for us, so we were like .. "Hey, there is something that we can use to solve our problem". 

@bahildebrand 
> This was previously panicking if the headers were not present correct? So the addition of these Options allows the deserialization to succeed if they are not present?
Yes, I noticed that it works in the Node.js client and the .NET client, but not in the Rust client.

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L114-L121

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/RuntimeApiHeaders.cs#L33-L34

diceride:
@davidbarsky Alright, I understand now. Thank you. We use the `custom` property of the client context to transfer metadata to the client. We don't use the Mobile SDKs. The client context seemed to be universal in nature for us, so we were like .. "Hey, there is something that we can use to solve our problem". 

@bahildebrand 
> This was previously panicking if the headers were not present correct? So the addition of these Options allows the deserialization to succeed if they are not present?

Yes, I noticed that it works in the Node.js client and the .NET client, but not in the Rust client.

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L114-L121

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/RuntimeApiHeaders.cs#L33-L34

diceride:
@davidbarsky We use the `custom` property of the client context to transfer metadata to the client. We don't use the Mobile SDKs. The client context seemed to be universal in nature for us, so we were like .. "Hey, there is something that we can use to solve our problem". 

@bahildebrand 
> This was previously panicking if the headers were not present correct? So the addition of these Options allows the deserialization to succeed if they are not present?

Yes, I noticed that it works in the Node.js client and the .NET client, but not in the Rust client.

https://github.com/aws/aws-lambda-nodejs-runtime-interface-client/blob/e5d62346eb08f1723348197df70d4f55f9c2ebcf/src/Runtime/InvokeContext.ts#L114-L121

https://github.com/aws/aws-lambda-dotnet/blob/master/Libraries/src/Amazon.Lambda.RuntimeSupport/Client/RuntimeApiHeaders.cs#L33-L34

lambda-runtime/src/types.rs
```diff
@@ -57,10 +57,12 @@ struct MobileClientIdentity(String);
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq)]
 pub struct ClientContext {
     /// Information about the mobile application invoking the function.
-    pub client: ClientApplication,
+    pub client: Option<ClientApplication>,
```
calavera, line 60: I wonder if this should also implement default, instead of wrapping it with an option. My main concern is that this is another breaking change for people that might currently be using this `ClientContext`. The `ClientApplication` struct looks easy enough to derive `Defaut`. What do you think @diceride ?

diceride: Well, using `Optional` seemed to be more sound to me. `ClientContext` never worked before (it was never deserialized). So I don't see the _breaking_ (breaking like in "it breaks 50% of user code") change. If you do not use the Mobile SDK (which is not recommended by AWS as @davidbarsky mentioned), the client `context lambda-runtime-client-context` and the cognito `identity lambda-runtime-cognito-identity` headers will never be set. So we can save some bytes by using `None` instead of `Default`. I mean, we are talking about peanuts. Originally, I looked at the Node.js client, where it can be `undefined` and in the .NET client it can be `null`. I'm not sure about other client implementations? I have no strong feelings here, let me know what you think. I can change it either way. Just wanted to point out some arguments. I already wrote too much, this comment is way too long.

diceride: I made it `Default` instead of `Optional` as suggested.

calavera (approved):
thanks for indulging me with my very own personal opinions @diceride. I really like how clean this looks like now 🙌

bahildebrand (approved):
LGTM. Waiting on all checks to pass and then I'll merge.

PR #384, opened 2021-12-24
lambda-extension: remove logging

mindriot101:
*Issue #, if available:*

*Description of changes:*

When running on AWS lambda I encountered the following error:

> thread 'main' panicked at 'Could not determine the UTC offset on this
> system. Possible causes are that the time crate does not implement
> "local_offset_at" on your system, or that you are running in a
> multi-threaded environment and the time crate is returning "None" from
> "local_offset_at" to avoid unsafe behaviour. See the time crate's
> documentation for more information.
> (https://time-rs.github.io/internal-api/time/index.html#feature-flags):
> IndeterminateOffset',
> /home/simon/.cargo/registry/src/github.com-1ecc6299db9ec823/simple_logger-1.16.0/src/lib.rs:409:85

This means that the simple logger setup could not find the current time.
Instead, we can print to stdout which is also collected by CloudWatch.
We can leave more sophisticated control of the output to the user.

### DIFF
--- a/lambda-extension/Cargo.toml
+++ b/lambda-extension/Cargo.toml
@@ -24,6 +24,4 @@ tokio-stream = "0.1.2"
 lambda_runtime_api_client = { version = "0.4", path = "../lambda-runtime-api-client" }
 
 [dev-dependencies]
-simple_logger = "1.6.0"
-log = "^0.4"
 simple-error = "0.2"
--- a/lambda-extension/README.md
+++ b/lambda-extension/README.md
@@ -10,17 +10,14 @@ The code below creates a simple extension that's registered to every `INVOKE` an
 
 ```rust,no_run
 use lambda_extension::{extension_fn, Error, NextEvent};
-use log::LevelFilter;
-use simple_logger::SimpleLogger;
-use tracing::info;
 
 async fn log_extension(event: NextEvent) -> Result<(), Error> {
     match event {
         NextEvent::Shutdown(event) => {
-            info!("{}", event);
+            println!("Shutdown {:?}", event);
         }
         NextEvent::Invoke(event) => {
-            info!("{}", event);
+            println!("Invoke {:?}", event);
         }
     }
     Ok(())
@@ -28,8 +25,6 @@ async fn log_extension(event: NextEvent) -> Result<(), Error> {
 
 #[tokio::main]
 async fn main() -> Result<(), Error> {
-    SimpleLogger::new().with_level(LevelFilter::Info).init().unwrap();
-
     let func = extension_fn(log_extension);
     lambda_extension::run(func).await
 }
--- a/lambda-extension/examples/basic.rs
+++ b/lambda-extension/examples/basic.rs
@@ -1,6 +1,4 @@
 use lambda_extension::{extension_fn, Error, NextEvent};
-use log::LevelFilter;
-use simple_logger::SimpleLogger;
 
 async fn my_extension(event: NextEvent) -> Result<(), Error> {
     match event {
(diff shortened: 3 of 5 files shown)


mindriot101:
Happy to discuss further how to handle this issue properly if required. Perhaps it's user error on my part.

bahildebrand:
I'm a little confused what the issue is here. Can you give a little more detail on your setup? Are you running theses examples directly, or basing your code off of them?

calavera (commented):
That code serves as examples, you don't really need it to run extensions. Wouldn't it be better to fix the underlying problem than leaving it for everyone to figure out how to setup logging?

I believe the examples should work again if you add this to the dev-dependencies:

```
time = { version = "0.3.5", features = ["local-offset"] }
```

See https://time-rs.github.io/internal-api/time/sys/local_offset_at/index.html

bahildebrand:
Ah this makes more sense now. Nice catch @calavera

davidbarsky:
I mentioned this to @nmoutschen over Twitter DMs, but I _think_ the right solution here is to _not_ use local time in when logging and instead default to UTC and rely on the log ingestion system (CloudWatch or otherwise) to determine the local ingestion time and convert it to a some localized time instead. If a change is to be made, I'd probably get rid of `simple_logger` and replace it with `env_logger` or `tracing_subscriber`.

(An important disclosure: I help maintain some tracing libraries, including `tracing_subscriber`. However, I think defaulting to `tracing_subscriber` is a good choice because [awslabs/aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust) also uses `tracing` and I strongly suspect that the _other_ big library people use with this crate is the AWS SDK.)

calavera:
> (An important disclosure: I help maintain some tracing libraries, including `tracing_subscriber`. However, I think defaulting to `tracing_subscriber` is a good choice because [awslabs/aws-sdk-rust](https://github.com/awslabs/aws-sdk-rust) also uses `tracing` and I strongly suspect that the _other_ big library people use with this crate is the AWS SDK.)

yeah, that's a good point. The only reason why these examples use simple logger is because the other examples also use it, and I wanted to keep them similar. I normally use `tracing` for all my logs.

mindriot101:
Sorry for the brief description of the issue, I realise that wasn't enough information at the time of posting. Thank you all for understanding my issue and resolving it.

I've tried using `tracing_subscriber` for setting up the logger, and this works well. I think since the examples already use `tracing::info!`, we could update the log setup to use `tracing_subscriber`. In my case, I have replaced the `SimpleLogger` with:

```rust
use tracing_subscriber::util::SubscriberInitExt;

#[tokio::main]
    tracing_subscriber::fmt().json().finish().init();
    // ...
}
```

I agree with logging in UTC - I believe this should be the default. I'm happy to update this PR to use `tracing_subscriber` if we want?

coltonweaver:
> Sorry for the brief description of the issue, I realise that wasn't enough information at the time of posting. Thank you all for understanding my issue and resolving it.
> 
> I've tried using `tracing_subscriber` for setting up the logger, and this works well. I think since the examples already use `tracing::info!`, we could update the log setup to use `tracing_subscriber`. In my case, I have replaced the `SimpleLogger` with:
> 
> ```rust
> use tracing_subscriber::util::SubscriberInitExt;
> 
> #[tokio::main]
> async fn main() -> Result<(), Error> {
>     tracing_subscriber::fmt().json().finish().init();
>     // ...
> }
> ```
> 
> I agree with logging in UTC - I believe this should be the default. I'm happy to update this PR to use `tracing_subscriber` if we want?

If you are willing to do it in this PR we're more than happy to review it! Otherwise I agree with @davidbarsky, and think we should at the very least open an issue for this separately and start working on it.

calavera:
hey @mindriot101, thanks a lot for bringing this up. We've decided to replace all the logging examples with tracing, so this change is not necessary anymore, see #389. I'm going to close this PR 🙌

mindriot101:
That's great, I'm glad this discussion happened at least 👍

PR #387, opened 2021-12-31
fix: add support for null/invalid type for multi-value-headers

mbergkvist:
*Issue #, if available: 367*

*Description of changes:*

This adds support for invalid/null value for multi-value headers, defaulting to an empty `HeaderMap` if need be, increasing resilience if the Lambda function would get an invalid value for multi-value headers.

This is a continuation of the work done by @nmoutschen in #371.

I have also successfully run the test in https://github.com/awslabs/aws-lambda-rust-runtime/issues/367#issuecomment-998753570 locally.

### DIFF
--- a/lambda-http/src/request.rs
+++ b/lambda-http/src/request.rs
@@ -317,7 +317,7 @@ where
         }
     }
 
-    deserializer.deserialize_map(HeaderVisitor)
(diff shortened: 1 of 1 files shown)


calavera (approved):

bahildebrand (approved):
LGTM

PR #388, opened 2021-12-31
fix: move clippy::cargo lint to warn level

bahildebrand:
fix: Allow multiple versions of crates

Downstream dependencies may have different versions of crates, and
this could change at any time. Disable this lint so that clippy
builds are not at the mercy of donwstream dependencies.


coltonweaver (approved):
Nice, thanks for fixing this!

PR #393, opened 2022-01-07
Allow base paths in the runtime api uri

### LINKED ISSUE #390: Allow any path parts in `AWS_LAMBDA_RUNTIME_API` to remain intact
Hey folks, this project is great! I have been at work working on implementing full Lambda custom runtime API support in OpenJS Architect, and wanted to raise some behavior that appears to be specific to the Rust Lambda runtime library.

A goal of Architect's local execution model is to be as fast (or faster) than real Lambda; as such, we do not rely on Docker. To accomplish this, we run handler functions in short-lived child processes on the local machine and do our own emulation (instead of relying on a lambCI or SAM Docker images. Since each invocation shares the same host ports, this means we cannot run the local RIE server, as there would be port conflict.

So what I've been working on is building our own simple take on the Lambda runtime API, but shared across multiple Lambdas – but there’s a bit of a problem. The current runtime API design assumes it is not necessary for the Next invocation request to need any identification, as the each Lambda in question is contacting `localhost:9001` – so knowledge of which Lambda is getting the invocation data is implied. In our case, again, the Lambda runtime API will be shared across multiple Lambda invocations (and we'll be responsible for maintaining state), so I need a way to identify which Lambda is requesting the invocation.

I had hoped to set `AWS_LAMBDA_RUNTIME_API` to `localhost:{port}/{lambdaID}/`, resulting in the Lambda runtime API client hitting `/{lambdaID}/2018-06-01/runtime/invocation/next`. I believe this [should work in the Go Lambda client](https://github.com/aws/aws-lambda-go/blob/main/lambda/runtime_api_client.go#L39), for example. Unfortunately, the implementation in the Rust client as it stands today strips out everything except the host name and port, ensuring that every request is only ever sent to `/2018-06-01/runtime/invocation/next`. And without headers or query params, I have no way of identifying which Lambda is requesting its next invocation.

It would be great if you could follow the Lambda Go runtime approach, leaving the contents of the environment variable intact.

## Repro steps
- Build the [basic example function](https://github.com/awslabs/aws-lambda-rust-runtime#example-function)
- Set up a simple HTTP server on port `9001` to simply print request metadata
- Execute a Lambda with `AWS_LAMBDA_RUNTIME_API` set to `localhost:9001/foo/`
- `GET /2018-06-01/runtime/invocation/next` will be fired, and not `GET /foo/2018-06-01/runtime/invocation/next`

calavera:

*Issue #, if available:*

Fixes #390

*Description of changes:*

This change allows to set the runtime api location with a base path.
This helps tools outside production to set their own full locations
that might not be in the root of the uri.

### DIFF
--- a/lambda-runtime-api-client/src/lib.rs
+++ b/lambda-runtime-api-client/src/lib.rs
@@ -4,7 +4,7 @@
 
 //! This crate includes a base HTTP client to interact with
 //! the AWS Lambda Runtime API.
-use http::{uri::Scheme, Request, Response, Uri};
+use http::{uri::PathAndQuery, uri::Scheme, Request, Response, Uri};
 use hyper::{
     client::{connect::Connection, HttpConnector},
     Body,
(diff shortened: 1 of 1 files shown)


nmoutschen (approved):
LGTM!

PR #394, opened 2022-01-08
Wrap incoming extension events in a struct

### LINKED ISSUE #392: Lambda Logs API?
I've successfully used this project's [lambda-extension](https://github.com/awslabs/aws-lambda-rust-runtime/tree/master/lambda-extension) library to respond to Invoke and Shutdown events (as the examples demonstrate), but I'm struggling to figure out the best path to use the [Lambda Logs API](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html#runtimes-logs-api-msg-function).

My understanding of the `lambda-extension` library code (as a Rust novice) is that the `extension_id` is private.  In order to use the Lambda Logs API you must subscribe with the extension ID, which you get from registering the extension with the Extensions API.  But there's no way that I can see to get the `extension_id` after registering the extension.

So my question is: If I've used the `lambda-extension` library as the base for my extension, how can I use the Lambda Logs API if the extension ID is unavailable externally?  I could imagine a new `lambda-logs` library or additional functions in the `lambda-extension` library, but neither of those options exist currently.  What can I do in the meantime without essentially reimplementing the `lambda-extension` library myself?

calavera:

*Issue #, if available:*

Fixes #392 

*Description of changes:*

This new structure can be used to pass additional information to the extension in each invocation, like the extension id that the Runtime assigns to the extension when it's initialized.

I'm not super keen in cloning the id in every event, but if I start messing with lifetimes, the extension trait becomes much more complicated to implement. I think this is good enough to fix the current issue, and we can iterate on it if it really becomes a problem before making the crate more stable.

### DIFF
--- a/lambda-extension/examples/basic.rs
+++ b/lambda-extension/examples/basic.rs
@@ -1,7 +1,7 @@
-use lambda_extension::{extension_fn, Error, NextEvent};
+use lambda_extension::{extension_fn, Error, LambdaEvent, NextEvent};
 
-async fn my_extension(event: NextEvent) -> Result<(), Error> {
-    match event {
+async fn my_extension(event: LambdaEvent) -> Result<(), Error> {
+    match event.next {
         NextEvent::Shutdown(_e) => {
             // do something with the shutdown event
         }
--- a/lambda-extension/examples/custom_events.rs
+++ b/lambda-extension/examples/custom_events.rs
@@ -1,7 +1,7 @@
-use lambda_extension::{extension_fn, Error, NextEvent, Runtime};
+use lambda_extension::{extension_fn, Error, LambdaEvent, NextEvent, Runtime};
 
-async fn my_extension(event: NextEvent) -> Result<(), Error> {
-    match event {
+async fn my_extension(event: LambdaEvent) -> Result<(), Error> {
+    match event.next {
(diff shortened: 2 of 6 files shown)


bahildebrand (approved):
LGTM!

tannerntannern:
@calavera Correct me if I'm wrong, but I'm not sure this actually resolves the linked issue.  According to the [Lambda Logs API docs](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html#runtimes-logs-api-subscribing) (emphasis mine):
> To subscribe to receive logs, you need the extension identifier (Lambda-Extension-Identifier). First register the extension to receive the extension identifier. **Then subscribe to the Logs API during initialization. After the initialization phase completes, Lambda does not process subscription requests.**

As I understand it, accessing the extension id during an invoke event and subscribing to the Logs API at that point would be too late (not sure if that was the intended path).  Subscribing to the Logs API must happen between extension registration and the first call to `/event/next`, which still doesn't appear to be possible.

To clarify, I don't think this is a bad change.  The Lambda Logs API also sends back the extension id in the response headers for the `/event/next` endpoint, so if anything these changes bring us more in line with the actual AWS APIs.  I don't think we can say #392 is resolved however.

Issue #109, opened 2019-06-06
Query String not present in Request::uri()


bernardobelchior:
Hey! I was trying to obtain values from the query string from a API GW request and I think I found a bug.

I was trying get the query string values by using [`Url::parse`](https://docs.rs/url/1.7.2/url/struct.Url.html#method.parse) to parse the URI obtained from `request.uri().to_string()`. However the query string is [not added](https://github.com/awslabs/aws-lambda-rust-runtime/blob/0ac259a711d6486e7f6b7dc641539fbec1ee7b79/lambda-http/src/request.rs#L263) to the URI when converting from `LambdaRequest` to `Request`. I believe this is not intended, but would like to confirm that. 

Note 1: Eventually, I found the `request.query_string_parameters()` function, as mentioned in the [examples](https://docs.rs/lambda_http/0.1.1/lambda_http/#examples).
Note 2: I also mentioned this problem in [this](https://github.com/awslabs/aws-lambda-rust-runtime/issues/107#issuecomment-499523868) issue.

softprops:
Quick question, in your use case is this more of an inconvenience or a discovery issue? 

A goal of using the http::Request type is to work with the existing http crate ecosystem, in that regard if you are using lambda http with another crate that does query string parsing this nuance of http lambda is an inconvenience.

 A feature of api gateway is that these query 
string parameters have already been preparsed  for you meaning you'd end up needing to reparse what's already been parsed if you were to work with a raw http::Request uri. In that sense it may have been a case where discovery was an issue where it may not have been clear api gateway provides this for you and how that's exposed via lambda http.

I'm curious if it was one or the other or maybe an angle I'm not thinking of yet.

bernardobelchior:
At first, it was a discovery issue. Eventually, I stumbled upon `RequestExt` and managed to get the query string.
Nevertheless, I still believe this is not the intended behavior. If I have a use case where I need the whole URL, I'd need to obtain it from `request.uri()` and then add the query string manually. However, this is not how the vanilla `http::request` works, so this seems like unintended behavior.

seanpianka:
It would be helpful if there was an escape-hatch available to add the url-encoded query parameters back into the URI. For Lambda applications that include a separate HTTP routing library, the default behavior of removing the query parameters from the original URI is unexpected.

[...]

PR #395, opened 2022-01-09
Fix request query strings

calavera:

*Issue #, if available:*

Fixes https://github.com/awslabs/aws-lambda-rust-runtime/issues/109

*Description of changes:*

- Add query strings to the request uri. This ensures that we don't mangle information coming into lambda.
- Fix bug in StrMapIter that ignored multi-value query parameters. Until now, this iterator only looped over the first element of each key, even if the key had multiple values assigned.
- Add tests to cover these changes.


bahildebrand (approved):
LGTM

PR #401, opened 2022-01-14
Implement tower::Service trait

### LINKED ISSUE #96: Consider Alternate Handler Types
(Pulled from #94)

I wonder if we can support that through an optional extension for customers that _do_ care about this. To that end, I've been talking to @carllerche—he might be able to comment on this more—about the usage of Tower in underlying of the Lambda runtime, and my current thinking is that we can provide _several_ valid handler signatures as implemented by specialized tower services and a corresponding `Into<Service>` (for context, [this is the foundational Service trait](https://docs.rs/tower-service/0.2.0/tower_service/trait.Service.html)) that'll be executed by the the Lambda runtime itself. Given this definition, we can provide several specialized services along the lines of:

- `Service<(http::Request<T>, Context)> -> http::Response<U>`
- `Service<(http::Request<T>, Context)> -> Result<http::Response<U>, HandlerError>`
- `Service<(aws_lambda_events::event::sqs::SqsEvent, Context)> -> U`
- `Service<(aws_lambda_events::event::sqs::SqsEvent, Context)> -> Result<U, HandlerError>`

Note: _each_ service function signature would be implemented in terms of a sealed trait similar to [`HttpService`](https://github.com/tower-rs/tower-http/blob/6d7a9fdc8e2f7ec047dd24b39a06d5bedddb7ca1/tower-http-service/src/service.rs#L14). 

In the absence of a handler error, the `Into<Service>` implementation would place a default value for `HandlerError`. Handler construction will still be available using a something similar to a [`ServiceFn`](https://github.com/tower-rs/tower/blob/92f4a0cb729a82df01086fa3c8e0e4dd80d36947/tower-service-util/src/service_fn.rs#L5).

To address some potential concerns: 

- We might be able to cut down on our compile times by using more, smaller crates _and_ disabling some of the chunkier features in Tokio, Hyper, and by removing our usage of Failure.
- The internals are more easily testable using [tower-mock](https://github.com/tower-rs/tower/tree/master/tower-mock).
- We can provide support for Tide's [`HttpService`](https://github.com/rustasync/http-service) behind a feature flag.

### LINKED ISSUE #374: Replace lambda_runtime::Handler with tower::Service?

nmoutschen:
*Issue #, if available:* https://github.com/awslabs/aws-lambda-rust-runtime/issues/374

*Description of changes:*

Replaces all custom traits with `tower::Service<LambdaEvent>`:

* [x] Switch `lambda_runtime::Runtime`
* [x] Switch `lambda_extension::Extension`
* [x] Switch `lambda_http::Handler`
* [x] Add run helper `tower::Service<Request>`
* [x] Add integration test for `tower::Service<Request>`
* [x] Update `lambda_integration_tests`
* [x] Update documentation

Closes #374
Closes #96


calavera (approved):
I really like how this turned out, good job!

coltonweaver (approved):
Agree with @calavera. This is a very clean change and easy to follow. Great work.

nmoutschen:
Thanks for the check, folks! I'll go ahead and merge this.

PR #413, opened 2022-02-09
Log error chain to CloudWatch Logs

### LINKED ISSUE #348: Logging error chains to CloudWatch
I noticed that the Rust runtime currently logs handler errors using `{}` as the format:

https://github.com/awslabs/aws-lambda-rust-runtime/blob/3871cb4c8fa0319eb157c15eb9e443b1b7526dd8/lambda-runtime/src/lib.rs#L178

However, the widely used [anyhow crate](https://github.com/dtolnay/anyhow) will only include the error chain with `{:?}` or `{:#}`:

https://github.com/dtolnay/anyhow/blob/71caf88b78cc3faf2b99ac3d7397a66a7da5a930/src/fmt.rs#L7-L40

Among other things, this means that only the last error will be logged to CloudWatch and that all underlying causes are lost when using [with_context](https://docs.rs/anyhow/1.0.44/anyhow/trait.Context.html) to augment errors.

My current solution/workaround is to add an additional log statement like this:

```rust
#[tokio::main]
    env_logger::try_init()?;

    lambda_runtime::run(handler_fn(|_: Input, _: Context| async {
        let output = handler().await.map_err(|e| {
            error!("{:?}", e); // log error chain to CloudWatch
            e
        })?;
        Ok(output) as Result<Output>
    }))
    .await?;

    Ok(())
}
```

Of course, this duplicates error logging to some degree and would have to be done in every Lambda function. I wonder if there's a better way of telling the runtime _how_ to log errors?

mlafeldt:
The anyhow crate will only include the error chain with `{:?}` or `{:#}`.

Without this change, only the last error will be logged to CloudWatch and all underlying causes are lost when using `with_context` to augment errors.

Fixes #348

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -125,7 +125,7 @@ where
                         .into_req()
                     }
                     Err(err) => {
-                        error!("{}", err); // logs the error in CloudWatch
+                        error!("{:?}", err); // logs the error (chain) to CloudWatch Logs
                         EventErrorRequest {
                             request_id,
                             diagnostic: Diagnostic {


calavera (changes requested):
It looks like this change is not as simple as we thought. There are errors now because some traits need to implement the Debug trait 😔

calavera:
I've addressed the build errors in #418

PR #448, opened 2022-03-17
Be nice to more standard http servers.

calavera:
Handle HTTP status codes coming from the runtime server
in a way that plays nice with local development tools
like the runtime interface emulator.

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -97,15 +97,28 @@ where
     {
         let client = &self.client;
         tokio::pin!(incoming);
-        while let Some(event) = incoming.next().await {
+        while let Some(next_event_response) = incoming.next().await {
             trace!("New event arrived (run loop)");
-            let event = event?;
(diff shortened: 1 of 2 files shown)


lambda-runtime/src/lib.rs
```diff
@@ -97,15 +97,28 @@ where
...
         let client = &self.client;
         tokio::pin!(incoming);
-        while let Some(event) = incoming.next().await {
+        while let Some(next_event_response) = incoming.next().await {
             trace!("New event arrived (run loop)");
-            let event = event?;
+            let event = next_event_response?;
             let (parts, body) = event.into_parts();
 
+            if parts.status == http::StatusCode::NO_CONTENT {
```
rimutaka, line 105: Can an AWS server return this status? If not, should this be under an optional flag or for debug only?

calavera: why should we complicate it more than this?

nmoutschen (approved):
LGTM!

Issue #450, opened 2022-03-21
http request url should not include stage in lambda_http


bnusunny:
In lambda_http v0.5.0 and v0.5.1, stage is added to http request url. This is a breaking change from v0.4. 

Stage should not be included in the generated http request url. Including the stage in the url will couple the lambda function to the API GW deployment stage. The orginal behavior allows one lambda function to be deployed in different stages and works pertfectly fine. I think this is one of the benefit of seperating stage and path in API Gateway. 

@calavera could we revert this behavior?

calavera:
> Stage should not be included in the generated http request url.

As far as I understand the API GW docs, the stage is part of the URL, before the path. It makes sense to me that the URL that the lambda function receives is exactly the same that requests are sent to through API GW.

> Including the stage in the url will couple the lambda function to the API GW deployment stage

I don't think this is correct. The URL is generated when requests come in, so if you have the same lambda deployed in two different stages, when requests come in, the lambda function will get different URLs, which is expected.

This issue has more information about this breaking change: https://github.com/awslabs/aws-lambda-rust-runtime/issues/107

bnusunny:
Yes, the stage is in the frontend URL called by clients. But API Gateway does the work to parse the request and send the 
 proxy integration request to lambda in [this format](https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-input-format). In this format, path and stage are seperated properties. 

The lambda functions can do routing based on path without stage. This allows developers to deploy the same code to different stages (e.g. dev, test, staging, prod). Serverless Express and Serverless Java Container projects are working this way. [Here](https://github.com/vendia/serverless-express/tree/mainline/examples/basic-starter-api-gateway-v1) is an example from Serverless Express project. 

If developers want to get stage, she can access it from request context.

calavera:
> [Here](https://github.com/vendia/serverless-express/tree/mainline/examples/basic-starter-api-gateway-v1) is an example from Serverless Express project.

The reason why serverless-express works as you describe it is because they access the raw event coming from API GW, or ALB, they are not using an abstraction like lambda_http. In their case, they are using the basic javascript runtime, it's the equivalent of using the lambda_runtime directly, without the http abstraction on top.

bnusunny:
Yes, #453 works for me. Thanks.

bnusunny:
[...]

Issue #107, opened 2019-05-27
Request URI is wrong when using API Gateway in some common cases


iliana:
Without using a custom domain mapping in API Gateway, the URI looks like:

```
https://1234567890.execute-api.us-east-1.amazonaws.com/prod/a/b/c
```

where `prod` is the deployment stage. Currently lambda_http doesn't include the stage in the URL (but it looks like it's available in the event).

I tested it and it works fine with an empty custom domain mapping, but a non-empty custom domain mapping might also not work; I didn't test that, and I'm not sure if the domain mapping is present in the event.

(I'm running into this because I'm writing request validation for a Twilio endpoint, which involves concatenating key/value pairs onto the URL Twilio knows. So I'm relying on `Request::uri().to_string()`.)

bernardobelchior:
This is due to [this](https://github.com/awslabs/aws-lambda-rust-runtime/blob/0ac259a711d6486e7f6b7dc641539fbec1ee7b79/lambda-http/src/request.rs#L263) portion of code:

```rust
let mut builder = HttpRequest::builder();
        builder.method(http_method);
        builder.uri({
            format!(
                "{}://{}{}",
                headers
                    .get("X-Forwarded-Proto")
                    .map(|val| val.to_str().unwrap_or_else(|_| "https"))
                    .unwrap_or_else(|| "https"),
                headers
                    .get(HOST)
                    .map(|val| val.to_str().unwrap_or_default())
                    .unwrap_or_default(),
                path
            )
        });
```

You can see that not only are stage variables not available, but the query string is also being ignored. 
I have run into the same problems, as I am using `Uri::parse` to obtain values from the query string.

arahalevich-move:
Hey @bernardobelchior! I'm not 100% sure that this exactly what you need, but I'd suggest to take a look at [RequestExt](https://docs.rs/lambda_http/0.1.1/lambda_http/trait.RequestExt.html) trait. It provides few methods for accessing url and path params.

saks:
Hey @bernardobelchior! I'm not 100% sure that this exactly what you need, but I'd suggest to take a look at [RequestExt](https://docs.rs/lambda_http/0.1.1/lambda_http/trait.RequestExt.html) trait. It provides few methods for accessing url and path params.

bernardobelchior:
Thanks, @saks! That was the solution! 
Nevertheless, this still seems like a bug

softprops:
I can take a closer look at this

[...]

PR #451, opened 2022-03-21
remove stage from generated http request url

bnusunny:
*Issue #, if available:*

Issue #450 

*Description of changes:*

This PR removes stage from generated http request url. This is the behavior in v0.4.1. 


calavera (changes requested):
I'm not sure about this, tbh. I'd like to understand why this needs to be reverted when it was a fix for a customer problem.

see for context on this behavior,  https://github.com/awslabs/aws-lambda-rust-runtime/issues/107

See my comment about stage behavior here: https://github.com/awslabs/aws-lambda-rust-runtime/issues/450#issuecomment-1074019918

calavera:
@bnusunny please open a different pull request for the User-Agent change.

bnusunny:
Sorry. My mistake.

PR #453, opened 2022-03-22
Expose the raw http path coming from the lambda event.

calavera:
The API GW url includes th API GW stage information, so it's not easy to know what's the raw http path coming into the event.

With this change, we expose the raw http path in a general way, so people that need to know the exact raw path have a common interface, regardless of where the event comes from.

I believe this is a better solution to #450 than #451 because the AWS documentation explicitly says that the API GW URL includes the stage when requests come in, see https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-custom-domains.html, so our behavior is going to be more expected by API GW users with our current implementation. However, I see the need to access the raw path when you write integrations with other services. With this change, people can access that raw path in a consistent manner regardless where the event comes from.

/cc @bnusunny

### DIFF
--- a/lambda-http/examples/hello-raw-http-path.rs
+++ b/lambda-http/examples/hello-raw-http-path.rs
@@ -1,0 +1,13 @@
+use lambda_http::{service_fn, Error, IntoResponse, Request, RequestExt};
+
+#[tokio::main]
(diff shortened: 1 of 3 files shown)


nmoutschen (approved):
LGTM!

Issue #452, opened 2022-03-21
Allow customized User-Agent used by lambda-runtime-api-client


bnusunny:
In [AWS Lambda Adapter](https://github.com/aws-samples/aws-lambda-adapter), we want to send a custom User-Agent to lambda Runtime API. 

lambda-runtime-api-client has a default User-Agent string, e.g. "aws-lambda-rust/0.5.0". To allow custom User-Agent, we can set an environment variable CUSTOM_USER_AGENT at compile time, like this. 
```rust
const CUSTOM_USER_AGENT: Option<&str> = option_env!("CUSTOM_USER_AGENT");
```
And when building the http request, use it if it is configured, otherwise send the default User-Agent. e.g.

```rust
pub fn build_request() -> http::request::Builder {
    let user_agent = match CUSTOM_USER_AGENT {
        Some(value) => value,
        None => USER_AGENT,
    };
    http::Request::builder().header(USER_AGENT_HEADER, user_agent)
}
```

@calavera @nmoutschen What's your opinion?

calavera:
That makes sense to me. I'd make the variable name `LAMBDA_RUNTIME_USER_AGENT` to be explicit on what's used for.

rimutaka:
1. What is the use case for this? Just curious.
2. `user_agent` can also be a constant with the match statement that builds it

bnusunny:
Good suggestions. I will send a PR shortly. 

This will help us to understand the usage of different runtimes/custom clients.

PR #454, opened 2022-03-22
allow customized User_Agent for lambda runtime api client

bnusunny:
*Issue #, if available:*

#452 

*Description of changes:*

allow customized User_Agent for lambda runtime api client

### DIFF
--- a/lambda-runtime-api-client/README.md
+++ b/lambda-runtime-api-client/README.md
@@ -33,3 +33,8 @@ async fn main() -> Result<(), Error> {
     client.call(request).await
 }
 ```
+
+## Custom User Agent
+
+To customize User Agent header sent to Lambda Runtime API, you can configure an environment variable `LAMBDA_RUNTIME_USER_AGENT` at compiling time.
+This will overide the default User Agent. 
--- a/lambda-runtime-api-client/README.md
+++ b/lambda-runtime-api-client/README.md
@@ -33,8 +33,3 @@ async fn main() -> Result<(), Error> {
     client.call(request).await
 }
 ```
-
-## Custom User Agent
-
-To customize User Agent header sent to Lambda Runtime API, you can configure an environment variable `LAMBDA_RUNTIME_USER_AGENT` at compiling time.
-This will overide the default User Agent. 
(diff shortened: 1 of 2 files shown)


bnusunny:
We want to track the usage of AWS Lambda Adapter. Users of Lambda Adapter are not supposed to change the User-Agent.
I think customize User-Agent at compiling time is the right solution.

lambda-runtime-api-client/README.md
```diff
@@ -33,3 +33,8 @@ async fn main() -> Result<(), Error> {
     client.call(request).await
 }
 ```
+
+## Custom User Agent
+
+To customize User Agent header sent to Lambda Runtime API, you can configure an environment variable `LAMBDA_RUNTIME_USER_AGENT` at compiling time.
```
rimutaka, line 39: 1. Where will this value show up? CloudWatch logs?
2. Why would a user of the runtime do this?
3. Do I have to have to have my own value or is it OK to leave the default?

The purpose of it is not entirely clear.

bnusunny: Hmm, this paragragh is not clear. I will remove it. This is not a value users need to configure.

bnusunny:
Thanks for the suggestion. It is done. :)

nmoutschen:
> We want to track the usage of AWS Lambda Adapter. Users of Lambda Adapter are not supposed to change the User-Agent. I think customize User-Agent at compiling time is the right solution.

Agree with this. The use-case for changing the User-Agent is pretty niche, and not something end-users need to do.

nmoutschen (approved):

lambda-runtime-api-client/src/lib.rs
```diff
@@ -14,7 +14,8 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tower_service::Service;
 
 const USER_AGENT_HEADER: &str = "User-Agent";
-const USER_AGENT: &str = concat!("aws-lambda-rust/", env!("CARGO_PKG_VERSION"));
+const DEFAULT_USER_AGENT: &str = concat!("aws-lambda-rust/", env!("CARGO_PKG_VERSION"));
+const CUSTOM_USER_AGENT: Option<&str> = option_env!("LAMBDA_RUNTIME_USER_AGENT");
```
rimutaka, line 18: @bnusunny , Harold, sorry for nit-picking, but it would be nice to have the explanation in the comments for the future us.

```rust
// Use of custom value was requested by https://github.com/aws-samples/aws-lambda-adapter for internal AWS needs. It is not something an end user would want to change and the value is not propagated to CloudWatch. See #452
```

calavera (approved):

PR #460, opened 2022-04-03
Document option to run functions locally.

### LINKED ISSUE #77: Documentation: Running locally
Trying to run locally:

```
cargo run --example=basic
  Downloaded simple-error v0.1.13
   Compiling simple-error v0.1.13
   Compiling simple_logger v1.0.1
   Compiling lambda_runtime v0.2.0 (/Users/walther/git/aws-lambda-rust-runtime/lambda-runtime)
    Finished dev [unoptimized + debuginfo] target(s) in 3.87s
     Running `/Users/walther/git/aws-lambda-rust-runtime/target/debug/examples/basic`
thread 'main' panicked at 'Could not find runtime API env var: environment variable not found', lambda-runtime-core/src/runtime.rs:79:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

It would be useful to have some documentation on how to run Rust-based Lambda functions locally, for development & testing purposes.

calavera:
*Issue #, if available:*

Fixes #77 

*Description of changes:*

Cargo-Lambda includes a very straightforward command to run lambda locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,6 @@ This package makes it easy to run AWS Lambda Functions written in Rust. This wor
 - [![Docs](https://docs.rs/lambda-extension/badge.svg)](https://docs.rs/lambda-extension) **`lambda-extension`** is a library that makes it easy to write Lambda Runtime Extensions in Rust.
 - [![Docs](https://docs.rs/lambda_runtime_api_client/badge.svg)](https://docs.rs/lambda_runtime_api_client) **`lambda-runtime-api-client`** is a shared library between the lambda runtime and lambda extension libraries that includes a common API client to talk with the AWS Lambda Runtime API.
 
-
 ## Example function
 
 The code below creates a simple function that receives an event with a `firstName` field and returns a message to the caller. Notice: this crate is tested against latest stable Rust.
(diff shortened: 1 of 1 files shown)


brainstorm:
Thanks for this @calavera!

I'm back from holidays and I just tested `cargo-lambda` on my M1, thanks for this! One nitpick on the docs though... it seems that I have to pass `bootstrap` as the lambda name? See my rust-lambda example here: https://github.com/umccr/s3-rust-noodles-bam#with-cargo-lambda 

Namely:

```
$ cargo lambda start
$ cargo lambda invoke bootstrap --data-file event.json
```

I wonder if the `bootstrap` name could be obviated as it might confuse new users that don't know that's the actual lambda `bin` name?

calavera:
@brainstorm the first argument is the name of the binary to run as a function. Yes, in your case, it's bootstrap because that's the name you gave the binary in https://github.com/umccr/s3-rust-noodles-bam/blob/master/Cargo.toml#L26. This is necessary because a package might include multiple binaries, and we need to know then name of the binary to watch changes for. The invoke command can be run from any directory, so we need a way to know which lambda you want to run

You don't actually need that section in your Cargo.toml at all when working with cargo-lambda. You could remove it, and run this:

```
$ cargo lambda start
$ cargo lambda invoke s3-rust-noodles-bam --data-file event.json
```

Then `cargo lambda build` will create a binary named `bootstrap`, regardless which name you give the package or binary.

Don't hesitate to open issues in that project if you have ideas to improve the experience.

calavera:
@brainstorm  an option would be to make the name optional, and not pass the `--bin` flag to cargo if the name is missing. I don't know how well that'd work when people have multiple functions, but it might just be a matter of adding more documentation to explain all the possible use cases.

nmoutschen (approved):
LGTM!

Issue #438, opened 2022-03-06
Can not use TraceLayer from tower_http crate


yotaro-shimose:
Hi.
I'm enjoying this library to create http server using lambda with rust's low memory consumption and speed.
I'm new to rust so this might be stupid question.

I want to log any (successful or failed) response in cloud watch using layer like we can in actix_web just by adding a line of code like
```rust
App::new()
    .wrap(Logger::default())
    .route("/path, web::get().to(my_handler))
```

So I tried 
```rust
use lambda_http::{self, tower::ServiceBuilder, Error};
use lambda_http::{Request, Response};
use tower_http::trace::TraceLayer;

#[tokio::main]
async fn main() -> Result<(), Error> {
    let service = ServiceBuilder::new()
        .layer(TraceLayer::new_for_http())
        .service_fn(handler);
    lambda_http::run(service).await?;
    Ok(())
}

async fn handler(req: Request) -> Result<Response<String>, Error> {
    Ok(Response::new("Success".to_string()))
}
```
but this code does not compile.

The compiler says
```
the trait bound `lambda_http::Body: std::convert::From<tower_http::trace::ResponseBody<std::string::String, tower_http::classify::NeverClassifyEos<tower_http::classify::ServerErrorsFailureClass>, tower_http::trace::DefaultOnBodyChunk, tower_http::trace::DefaultOnEos, tower_http::trace::DefaultOnFailure>>` is not satisfied
the following implementations were found:
  <lambda_http::Body as std::convert::From<&'a [u8]>>
  <lambda_http::Body as std::convert::From<&'a str>>
  <lambda_http::Body as std::convert::From<()>>
  <lambda_http::Body as std::convert::From<std::borrow::Cow<'static, [u8]>>>
and 3 others
required because of the requirements on the impl of `std::convert::Into<lambda_http::Body>` for `tower_http::trace::ResponseBody<std::string::String, tower_http::classify::NeverClassifyEos<tower_http::classify::ServerErrorsFailureClass>, tower_http::trace::DefaultOnBodyChunk, tower_http::trace::DefaultOnEos, tower_http::trace::DefaultOnFailure>`
[...]

PR #466, opened 2022-04-14
feat!(lambda-http): accept http_body::Body in responses

nmoutschen:
*Issue #, if available:*

*Description of changes:*

Allows using any `http_body::Body` implementation in the response for `lambda_http`.

⚠️ At the moment, this PR removes the ability to return a `Response<String>` or any into value that implements [`Into<aws_lambda_events::encodings::Body>`](https://docs.rs/aws_lambda_events/latest/aws_lambda_events/encodings/enum.Body.html).

### DIFF
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -65,6 +65,7 @@ extern crate maplit;
 pub use http::{self, Response};
 use lambda_runtime::LambdaEvent;
 pub use lambda_runtime::{self, service_fn, tower, Context, Error, Service};
+use response::ResponseFuture;
 
 pub mod ext;
 pub mod request;
@@ -91,7 +92,8 @@ pub type Request = http::Request<Body>;
 #[doc(hidden)]
 pub struct TransformResponse<'a, R, E> {
     request_origin: RequestOrigin,
-    fut: Pin<Box<dyn Future<Output = Result<R, E>> + 'a>>,
+    fut_req: Pin<Box<dyn Future<Output = Result<R, E>> + 'a>>,
+    fut_res: Option<ResponseFuture>,
 }
 
 impl<'a, R, E> Future for TransformResponse<'a, R, E>
@@ -101,11 +103,22 @@ where
     type Output = Result<LambdaResponse, E>;
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut TaskContext) -> Poll<Self::Output> {
-        match self.fut.as_mut().poll(cx) {
-            Poll::Ready(result) => Poll::Ready(
-                result.map(|resp| LambdaResponse::from_response(&self.request_origin, resp.into_response())),
-            ),
-            Poll::Pending => Poll::Pending,
+        if let Some(fut_res) = self.fut_res.as_mut() {
+            match fut_res.as_mut().poll(cx) {
+                Poll::Ready(resp) => Poll::Ready(
+                    Ok(LambdaResponse::from_response(&self.request_origin, resp))
+                ),
+                Poll::Pending => Poll::Pending,
+            }
+        } else {
+            match self.fut_req.as_mut().poll(cx) {
+                Poll::Ready(Ok(resp)) => {
+                    self.fut_res = Some(resp.into_response());
+                    Poll::Pending
+                },
+                Poll::Ready(Err(err)) => Poll::Ready(Err(err)),
+                Poll::Pending => Poll::Pending,
+            }
         }
     }
 }
@@ -151,7 +164,7 @@ where
         let request_origin = req.payload.request_origin();
         let event: Request = req.payload.into();
         let fut = Box::pin(self.service.call(event.with_lambda_context(req.context)));
-        TransformResponse { request_origin, fut }
+        TransformResponse { request_origin, fut_req: fut, fut_res: None, }
     }
 }
 
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -8,7 +8,15 @@ use http::{
     header::{CONTENT_TYPE, SET_COOKIE},
     Response,
 };
+use http_body::Body as HttpBody;
+use hyper::body::to_bytes;
 use serde::Serialize;
+use std::future::ready;
+use std::{
+    any::{Any, TypeId},
+    pin::Pin,
+    future::Future,
+};
 
 /// Representation of Lambda response
 #[doc(hidden)]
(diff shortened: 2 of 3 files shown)


lambda-http/src/response.rs
```diff
@@ -87,71 +95,103 @@ impl LambdaResponse {
...
 where
-    B: Into<Body>,
+    B: IntoBody + 'static,
 {
-    fn into_response(self) -> Response<Body> {
+    fn into_response(self) -> ResponseFuture {
         let (parts, body) = self.into_parts();
-        Response::from_parts(parts, body.into())
+
+        let fut = async { Response::from_parts(parts, body.into_body().await) };
```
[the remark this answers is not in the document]
nmoutschen: Here, this is due to the `.await` on `into_body()`, which is needed to support `http_body::Body` implementations that stream the body back, as we need to collect the entire payload before we can return a Lambda response.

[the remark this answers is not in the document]
brainstorm: Is it possible to collect partial payloads? Thinking HTTP Ranges or terminate early if a piece of data has been seen in the response as it streams.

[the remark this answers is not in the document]
nmoutschen: @brainstorm Not immediately for this purpose, since here we just use this to transform it into a `aws_lambda_events::encoding::Body` enum, which we needs to generate the right response for API Gateway/ALB.

That said, if you're calling an API that returns an `http_body::Body` implementation, or some form of asynchronous stream, you could always implement that yourself in your function handler.

nmoutschen:
Marking this one as ready to review as there's nothing outstanding here and would help on https://github.com/awslabs/smithy-rs/pull/1338

calavera (approved):
LGTM

calavera:
@nmoutschen will this fix the problem described in #438 ?

lambda-http/src/lib.rs
```diff
@@ -91,7 +92,8 @@ pub type Request = http::Request<Body>;
 #[doc(hidden)]
 pub struct TransformResponse<'a, R, E> {
     request_origin: RequestOrigin,
-    fut: Pin<Box<dyn Future<Output = Result<R, E>> + 'a>>,
+    fut_req: Pin<Box<dyn Future<Output = Result<R, E>> + 'a>>,
+    fut_res: Option<ResponseFuture>,
```
david-perez, line 96: nit: since only one of these will be polled at a time, it might be better to model it as an enum with two variants.

lambda-http/src/response.rs
```diff
@@ -87,71 +95,103 @@ impl LambdaResponse {
...
+impl<S, B> IntoResponse for (S, B)
+where
+    S: TryInto<StatusCode> + 'static,
+    S::Error: fmt::Debug,
+    B: Into<Body> + 'static,
+{
+    fn into_response(self) -> ResponseFuture {
+        Box::pin(async move {
+            Response::builder()
+                .status(self.0.try_into().expect("unable to transform status code"))
```
david-perez, line 155: This will panic if the conversion fails. I'm not sure if this convenience blanket impl is worth panicking. At the very least it should be documented.

lambda-http/src/response.rs
```diff
@@ -87,71 +95,103 @@ impl LambdaResponse {
...
+    B: HttpBody + Unpin + 'static,
+    B::Error: fmt::Debug,
+{
+    fn into_body(self) -> BodyFuture {
+        if TypeId::of::<Body>() == self.type_id() {
+            let any_self = Box::new(self) as Box<dyn Any + 'static>;
+            // Can safely unwrap here as we do type validation in the 'if' statement
+            Box::pin(ready(*any_self.downcast::<Body>().unwrap()))
+        } else {
+            Box::pin(async move { Body::from(to_bytes(self).await.expect("unable to read bytes from body").to_vec()) })
```
david-perez, line 179: This will _always_ use the `impl From<Vec<u8>> for Body`, which always creates the `Body::Binary` variant. Is this correct? The body might just contain UTF-8-encoded text.

See https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html, which makes my head hurt.

nmoutschen:
@calavera Yep, I've also included a new example to fix that.

Fair points @david-perez ! I'm a bit busy with the Stockholm summit this week, but will update this PR either by the end of this week or the next with those changes.

Regarding panicking, this is to mimic the current behaviour, but while I'm making those changes, I might as well return a `Result` instead.

For the `Body::Binary` variant, yep, I was thinking of trying to transform into a `String` first, and only use Vec<u8> as a default. In the case of Lambda, it will always return a string (either proper string, or base64 encoded payload). Using `Body::Binary` basically does base64 encoding and set `isBase64Encoded` to true (see https://docs.aws.amazon.com/apigateway/latest/developerguide/set-up-lambda-proxy-integrations.html#api-gateway-simple-proxy-for-lambda-output-format). My assumption was that `Body::Binary` would be handled correctly, but I might have to do a few more tests there to validate this just to ensure we don't break anything.

eduardomourar:
Could this be merged and released, please?

lambda-http/src/response.rs
bnusunny: ALB integration use content-encoding and content-type headers to determine if the body should be base64 encoded. See it here: https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer 

Perhaps we could add utility methods to convert hyper::Response or reqwest::Reponse to http::Response<lambda_http::Body>. These methods can convert the Body to correct lambda_http::Body based on the same rules used by ALB.

lambda-http/src/lib.rs
hugobast: I need some help here @calavera, @eduardomourar or @david-perez 

The PR as it is now simply doesn't work. The lambdas using `lambda_http` get called (I can see my info logs with the event) but never return the response and time out.

I believe this `poll` function is at the core of the issue: https://github.com/awslabs/aws-lambda-rust-runtime/pull/466/files#diff-c62692e6b817476306b4d6966f541eec77a1768674e4dad5cf7f10c4faababa4R106-R119

I observed that the value of `self.fut_res` gets set then nothing else happens, the `poll` function never gets executed again with the value of `self.fut_res` set, it actually looks like it gets called once then that's it.

hugobast: Ok bear with me I think I know why this isn't working as-is: https://tokio.rs/tokio/tutorial/async#wakers

> When a future returns `Poll::Pending`, it must ensure that the waker is signaled at some point. Forgetting to do this results in the task hanging indefinitely.
> 
> Forgetting to wake a task after returning `Poll::Pending` is a common source of bugs.

Looks like by returning `Poll::Pending` from `Poll::Ready` and doing nothing else we are breaking the contract we also need to take an additional step

> Before returning `Poll::Pending`, we called `cx.waker().wake_by_ref()`. This is to satisfy the future contract. By returning `Poll::Pending`, we are responsible for signaling the waker.

The code should be:
```
Poll::Ready(Ok(resp)) => {
    self.fut_res = Some(resp.into_response());
    cx.waker().wake_by_ref()
    Poll::Pending
}
```

And I've confirmed now that this works correctly. Continuing on with testing.

hugobast:
@calavera here's the PR, happy to discuss https://github.com/awslabs/aws-lambda-rust-runtime/pull/491 please also check this commit that I have decided to drop since the API Gateway configuration to perform the tests per the table lead me into a dead-end (I could be wrong and I'm happy to pick it back up) https://github.com/hugobast/aws-lambda-rust-runtime/commit/83fa77dc0f4be23dbba9c2e6654fb00ec18d4e54

PR #474, opened 2022-05-05
Don't require Lambda `Service` to use `lambda_http::Error`

david-perez:
Currently, the `Service` that you pass to `lambda_http::run` is required
to use `lambda_http::Error` as its associated error type, a type-erased
boxed error type. However, all the runtime needs is for the error type
to implement `Debug` and `Display`. This commit relaxes this constraint,
allowing users to run a broader class of `Service`s.

### DIFF
--- a/lambda-http/examples/shared-resources-example.rs
+++ b/lambda-http/examples/shared-resources-example.rs
@@ -1,4 +1,4 @@
-use lambda_http::{service_fn, Error, IntoResponse, Request, RequestExt, Response};
+use lambda_http::{service_fn, Body, Error, IntoResponse, Request, RequestExt, Response};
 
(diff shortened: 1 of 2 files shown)


david-perez:
I don't understand how the tip of `origin/main` currently builds. CI fails because of the same reason; I've opened https://github.com/awslabs/aws-lambda-rust-runtime/pull/475.

calavera:
CI should work again if you rebase from main, sorry about that.

calavera (approved):
I'm good with this, but I'd like to hear if @nmoutschen has any concerns since he's changing some other behaviors in #466

nmoutschen (approved):
Sounds like a good idea and LGTM!

PR #491, opened 2022-06-21
feat(lambda-http): accept http_body::Body in responses

hugobast:
*Issue #, if available:*
This PR is meant to continue the work started here: https://github.com/awslabs/aws-lambda-rust-runtime/pull/466

*Description of changes:*

**Original description**:
 _Allows using any http_body::Body implementation in the response for lambda_http._

---

For context, the outstanding issue on the original PR stems from this comment: https://github.com/awslabs/aws-lambda-rust-runtime/pull/466#issuecomment-1121351819

I picked up this PR and worked on it to the point where I believe that I had a viable testing approach. Ultimately I am blocked from performing the testing that has been requested in the original PR.

Here's why

In order to configure `contentHandling` we have to put the integration in the `aws`  (ie. not `aws-proxy`) configuration.

The request is treated as a `LambdaRequest::WebSocket` by the lambda_http crate, why this is and why the rust code panics when operated in this way, I don't know.

On an `aws-proxy` configuration, binary formats are dealt with by encoding the body of the response as base64 and setting `isBase64Encoded: true` (documentation: [Working with binary media types for REST APIs - Amazon API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html))

I managed to configure an API Gateway + Lambdas to be tested with CONVERT_TO_BINARY and CONVERT_TO_TEXT and it doesn't appear to work. What does work is setting the integration as `aws-proxy` but then it operates completely differently.

If we test through API gateway we only have two type requests, one where the function returns a `Body::Binary` and one where it returns a `Body::Text` (maybe a 3rd one for `Body::Empty`).

I'm proposing a simpler verification where we rely on the current unit tests (I'm adding impl IntoResponse for binaries + tests) and pasting the Lambda console results (below) that I got while debugging.

Happy to dig in further if I'm wrong and misunderstanding something.

---

Here's the console output for both a lambda that returns text and one that returns bytes

#### Bytes

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "YmluYXJ5IGRhdGE=",
  "isBase64Encoded": true
}
```

#### Text

```
{
  "statusCode": 200,
  "headers": {},
  "multiValueHeaders": {},
  "body": "text data",
  "isBase64Encoded": false
}
```

---

Finally, because I was most of the way there I want to share the commit I removed from this PR where I setup an API gateway (non-proxy) and the makefile changes where I would have added the verifications. https://github.com/hugobast/aws-lambda-rust-runtime/commit/83fa77dc0f4be23dbba9c2e6654fb00ec18d4e54

---

### DIFF
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -8,6 +8,7 @@ use aws_lambda_events::event::alb::AlbTargetGroupResponse;
 use aws_lambda_events::event::apigw::ApiGatewayProxyResponse;
 #[cfg(feature = "apigw_http")]
 use aws_lambda_events::event::apigw::ApiGatewayV2httpResponse;
+use encoding_rs::Encoding;
 use http::header::CONTENT_ENCODING;
 use http::HeaderMap;
 use http::{
@@ -16,9 +17,10 @@ use http::{
 };
 use http_body::Body as HttpBody;
 use hyper::body::to_bytes;
+use mime::{Mime, CHARSET};
 use serde::Serialize;
+use std::borrow::Cow;
 use std::future::ready;
-use std::str::from_utf8;
 use std::{fmt, future::Future, pin::Pin};
 
 /// Representation of Lambda response
@@ -183,7 +185,7 @@ where
             value.to_str().unwrap_or_default()
         } else {
             // Content-Type and Content-Encoding not set, passthrough as utf8 text
-            return convert_to_text(self);
+            return convert_to_text(self, "utf-8".to_string());
         };
 
         if content_type.starts_with("text")
@@ -191,7 +193,7 @@ where
             || content_type.starts_with("application/javascript")
             || content_type.starts_with("application/xml")
         {
-            return convert_to_text(self);
+            return convert_to_text(self, content_type.to_string());
         }
 
         convert_to_binary(self)
@@ -206,15 +208,30 @@ where
     Box::pin(async move { Body::from(to_bytes(body).await.expect("unable to read bytes from body").to_vec()) })
 }
 
-fn convert_to_text<B>(body: B) -> BodyFuture
+fn convert_to_text<B>(body: B, content_type: String) -> BodyFuture
 where
     B: HttpBody + Unpin + 'static,
     B::Error: fmt::Debug,
 {
+    let mime_type = content_type.parse::<Mime>();
+
+    let encoding = match mime_type.as_ref() {
+        Ok(mime) => mime.get_param(CHARSET).unwrap_or(mime::UTF_8),
+        Err(_) => mime::UTF_8,
+    };
+
+    let label = encoding.as_ref().as_bytes();
+    let encoding = Encoding::for_label(label).unwrap_or(encoding_rs::UTF_8);
+
     // assumes utf-8
     Box::pin(async move {
         let bytes = to_bytes(body).await.expect("unable to read bytes from body");
-        Body::from(from_utf8(&bytes).expect("response body not utf-8"))
+        let (content, _, _) = encoding.decode(&bytes);
+
+        match content {
+            Cow::Borrowed(content) => Body::from(content),
+            Cow::Owned(content) => Body::from(content),
+        }
     })
 }
 
@@ -297,6 +314,24 @@ mod tests {
         assert_eq!(
             json,
             r#"{"statusCode":200,"headers":{"content-type":"application/json"},"multiValueHeaders":{"content-type":["application/json"]},"body":"000000","isBase64Encoded":false,"cookies":[]}"#
+        )
+    }
+
+    #[tokio::test]
+    async fn charset_content_type_header() {
+        // Drive the implementation by using `hyper::Body` instead of
+        // of `aws_lambda_events::encodings::Body`
+        let response = Response::builder()
+            .header(CONTENT_TYPE, "application/json; charset=utf-16")
+            .body(HyperBody::from("000000".as_bytes()))
+            .expect("unable to build http::Response");
+        let response = response.into_response().await;
+        let response = LambdaResponse::from_response(&RequestOrigin::ApiGatewayV2, response);
+
+        let json = serde_json::to_string(&response).expect("failed to serialize to json");
+        assert_eq!(
+            json,
+            r#"{"statusCode":200,"headers":{"content-type":"application/json; charset=utf-16"},"multiValueHeaders":{"content-type":["application/json; charset=utf-16"]},"body":"〰〰〰","isBase64Encoded":false,"cookies":[]}"#
         )
     }
 
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -185,7 +185,7 @@ where
             value.to_str().unwrap_or_default()
         } else {
             // Content-Type and Content-Encoding not set, passthrough as utf8 text
-            return convert_to_text(self, "utf-8".to_string());
+            return convert_to_text(self, "utf-8");
         };
 
         if content_type.starts_with("text")
@@ -193,7 +193,7 @@ where
             || content_type.starts_with("application/javascript")
             || content_type.starts_with("application/xml")
         {
-            return convert_to_text(self, content_type.to_string());
+            return convert_to_text(self, content_type);
         }
 
         convert_to_binary(self)
@@ -208,7 +208,7 @@ where
     Box::pin(async move { Body::from(to_bytes(body).await.expect("unable to read bytes from body").to_vec()) })
 }
 
-fn convert_to_text<B>(body: B, content_type: String) -> BodyFuture
+fn convert_to_text<B>(body: B, content_type: &str) -> BodyFuture
 where
     B: HttpBody + Unpin + 'static,
     B::Error: fmt::Debug,
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -126,6 +126,18 @@ impl IntoResponse for &str {
     }
 }
 
+impl IntoResponse for &[u8] {
+    fn into_response(self) -> ResponseFuture {
+        Box::pin(ready(Response::new(Body::from(self))))
+    }
+}
+
+impl IntoResponse for Vec<u8> {
+    fn into_response(self) -> ResponseFuture {
+        Box::pin(ready(Response::new(Body::from(self))))
+    }
+}
+
 impl IntoResponse for serde_json::Value {
     fn into_response(self) -> ResponseFuture {
         Box::pin(async move {
@@ -192,6 +204,15 @@ mod tests {
         let response = "text".into_response().await;
         match response.body() {
             Body::Text(text) => assert_eq!(text, "text"),
+            _ => panic!("invalid body"),
+        }
+    }
+
+    #[tokio::test]
+    async fn bytes_into_response() {
+        let response = "text".as_bytes().into_response().await;
+        match response.body() {
+            Body::Binary(data) => assert_eq!(data, "text".as_bytes()),
             _ => panic!("invalid body"),
         }
     }
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -14,8 +14,9 @@ use serde::Serialize;
 use std::future::ready;
 use std::{
     any::{Any, TypeId},
-    pin::Pin,
+    fmt,
     future::Future,
+    pin::Pin,
 };
 
 /// Representation of Lambda response
@@ -28,14 +29,11 @@ pub enum LambdaResponse {
     Alb(AlbTargetGroupResponse),
 }
 
-/// tranformation from http type to internal type
+/// Tranformation from http type to internal type
 impl LambdaResponse {
-    pub(crate) fn from_response<T>(request_origin: &RequestOrigin, value: Response<T>) -> Self
-    where
-        T: Into<Body>,
-    {
+    pub(crate) fn from_response(request_origin: &RequestOrigin, value: Response<Body>) -> Self {
         let (parts, bod) = value.into_parts();
-        let (is_base64_encoded, body) = match bod.into() {
+        let (is_base64_encoded, body) = match bod {
             Body::Empty => (false, None),
             b @ Body::Text(_) => (false, Some(b)),
             b @ Body::Binary(_) => (true, Some(b)),
@@ -95,7 +93,11 @@ impl LambdaResponse {
     }
 }
 
+/// Trait for generating responses
+///
+/// Types that implement this trait can be used as return types for handler functions.
 pub trait IntoResponse {
+    /// Transform into a Response<Body> Future
     fn into_response(self) -> ResponseFuture;
 }
 
@@ -106,9 +108,7 @@ where
     fn into_response(self) -> ResponseFuture {
         let (parts, body) = self.into_parts();
 
-        let fut = async {
-            Response::from_parts(parts, body.into_body().await)
-        };
+        let fut = async { Response::from_parts(parts, body.into_body().await) };
 
         Box::pin(fut)
     }
@@ -130,19 +130,18 @@ impl IntoResponse for serde_json::Value {
     fn into_response(self) -> ResponseFuture {
         Box::pin(async move {
             Response::builder()
-            .header(CONTENT_TYPE, "application/json")
-            .body(
-                serde_json::to_string(&self)
-                    .expect("unable to serialize serde_json::Value")
-                    .into(),
-            )
-            .expect("unable to build http::Response")
+                .header(CONTENT_TYPE, "application/json")
+                .body(
+                    serde_json::to_string(&self)
+                        .expect("unable to serialize serde_json::Value")
+                        .into(),
+                )
+                .expect("unable to build http::Response")
         })
     }
 }
 
-pub type ResponseFuture = Pin<Box<dyn Future<Output=Response<Body>>>>;
-
+pub type ResponseFuture = Pin<Box<dyn Future<Output = Response<Body>>>>;
 
 pub trait IntoBody {
     fn into_body(self) -> BodyFuture;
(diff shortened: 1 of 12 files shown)


lambda-http/src/response.rs
```diff
@@ -87,71 +93,99 @@ impl LambdaResponse {
...
+    B: HttpBody + Unpin + 'static,
+    B::Error: fmt::Debug,
+{
+    fn into_body(self) -> BodyFuture {
+        if TypeId::of::<Body>() == self.type_id() {
+            let any_self = Box::new(self) as Box<dyn Any + 'static>;
+            // Can safely unwrap here as we do type validation in the 'if' statement
+            Box::pin(ready(*any_self.downcast::<Body>().unwrap()))
+        } else {
+            Box::pin(async move { Body::from(to_bytes(self).await.expect("unable to read bytes from body").to_vec()) })
```
david-perez, line 173: I've experimented with this patch and the more I think about this issue, the more I think that it's impossible for `aws-lambda-rust-runtime` to provide a response body conversion that will work for all cases.

The issue I have with this line in particular is that it's effectively saying "if the user's `tower::Service` has a `Response` whose `Body` is not of type `lambda_http::Body`, then convert the response body [to_bytes](https://docs.rs/hyper/latest/hyper/body/fn.to_bytes.html), then to `Vec<u8>`, and use `lambda_http::Body::from` to yield a response whose body is of the `lambda_http::Body::Binary` variant".

This is problematic, because `lambda_http::Body::Binary` is signalling to API Gateway that the response contains binary data, and API Gateway does [all kinds of fun stuff](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html) before relaying the reponse to the client depending on several factors, including whether we signalled our payload contains binary data or not. For example, it might base64 encode the body before returning the response.

I deployed some Lambdas and was able to reproduce clearly undersirable behavior when the service returns a response body with `Data = Bytes` that contains a UTF-8 encoded string. Now, one might argue that the user should then use `Data = String`, or have their `Response` type be `String` directly (which would leverage the `impl IntoResponse for String` in `response.rs`, that correctly uses `lambda_http::Body::Text`). The problem is the user might not know or be in control of what body type they're returning because they are relying on a library. For example, Axum uses [`axum_core::body::BoxBody`](https://docs.rs/axum-core/latest/axum_core/body/type.BoxBody.html) as the unified response body type of all the routes, regardless of what data type your handler returns. `BoxBody` is a type-erased boxed body type that holds `Bytes`.

Here is an example repository that reproduces the problem: https://github.com/david-perez/lambda-pr-491. It has instructions to deploy a Rust Lambda function fronted by both a [HTTP API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api.html) and a [REST API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/apigateway-rest-api.html) using CDK.

The [Lambda function](https://github.com/david-perez/lambda-pr-491/blob/master/rust/src/bin/bootstrap.rs) uses Axum and contains a route that returns JSON:

```rust
#[tokio::main]
    let lambda_handler = Router::new().route("/prod/hello", get(hello));

    lambda_http::run(lambda_handler)
        .await
        .expect("something went wrong in the Lambda runtime");

    Ok(())
}

async fn hello() -> Json<Value> {
    Json(json!({ "message": "hello stranger!"}))
}
```

As explained above, this setup makes `aws-lambda-rust-runtime` return a `lambda_http::Body::Binary` body to the internal AWS Lambda service, signalling a binary payload.

When the response gets sent to the client, it seems like the behavior of API Gateway is dfferent depending on whether the API type is HTTP or REST.

```
$ curl https://<http-api-id>.execute-api.eu-west-1.amazonaws.com/prod/hello                             
{"message":"hello stranger!"}%
$ curl https://<rest-api-id>.execute-api.eu-west-1.amazonaws.com/prod/hello
eyJtZXNzYWdlIjoiaGVsbG8gc3RyYW5nZXIhIn0=%
```

The HTTP API works as we would expect, but the the REST API has "helpfully" base64 encoded the JSON text. In both cases the response has the `Content-Type: application/json` header.

hugobast: I'll continue with the information you've provided, I wasn't able to get to the core of the issue and was focusing too much on the issues I had with coming up with a test setup for the response conversion table, what you provided is straightforward and really helpful. Thank you, David!

hugobast: @david-perez I'm coming up empty-handed after a few hours of playing with this, the problem as I see it is that `axum::body::BoxBody` is a downcast (is that the correct term?) of `http_body::Body` and I don't see how we can get anything more out of it than the bytes it's holding. In other words, is it correct to say that it is essentially "just bytes" at this point and we have no choice but the `Body::Binary` variant?

That said I had an idea maybe this is what we are looking for

```suggestion
            Box::pin(async move {
                let bytes = to_bytes(self).await.expect("unable to read bytes from body");

                match from_utf8(&bytes) {
                    Ok(s) => Body::from(s),
                    Err(_) => Body::from(bytes.to_vec()),
                }
            })
```

To summarize the code we are commenting on

1. `lambda_http::Body` pass through as they are
2. Other `bytes::Bytes` types (of which `BoxBody` is)
  a. utf8 from bytes -> `Body::Text`
  b. if not -> `Body::Binary`

bnusunny: ALB uses `Content-Encoding` and `Content-Type` to determine if the body is binary (doc is [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer)). I also followed the same rules to convert the body in Lambda Adapter [here](https://github.com/awslabs/aws-lambda-web-adapter/blob/18db3940716d4cadcfb5cddc5326b3d11b70e1c1/src/main.rs#L90). 

> If the content-encoding header is present, the load balancer Base64 encodes the body and sets isBase64Encoded to true.
> 
> If the content-encoding header is not present, Base64 encoding depends on the content type. For the following types, the load balancer sends the body as is and sets isBase64Encoded to false: text/*, application/json, application/javascript, and application/xml. Otherwise, the load balancer Base64 encodes the body and sets isBase64Encoded to true.

This conversion probably needs to be done in one of into_response function, not in the into_body function.

david-perez: @bnusunny The documentation you linked is for when the Lambda _receives_ a request event from the ALB. This PR is for returning responses from the Lambda.

The documentation for responses for ALB is [here](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#respond-to-load-balancer). Again, it's the user who has to _manually_ set `isBase64Encoded`.

hugobast: Just a small update from playing around with your PR yesterday @david-perez 

The integration response from the Lambda is itself already base64 encoded. What I think we are experiencing for the RestApi; is it passes it through unchanged while the HttpApi base64 decodes it back into text.

I don't know where exactly this Body::Binary to base 64 conversion happens (you might know). I have a hunch that it's somewhere within the `lambda_runtime` crate since I tried a different approach (see further down)
.
```
{
  "statusCode": 200,
  "headers": {
    "content-type": "application/json",
    "content-length": "29"
  },
  "multiValueHeaders": {
    "content-type": [
      "application/json"
    ],
    "content-length": [
      "29"
    ]
  },
  "body": "eyJtZXNzYWdlIjoiaGVsbG8gc3RyYW5nZXIhIn0=",
  "isBase64Encoded": true
}
```

In another experiment I did, I went looking at alternatives to `lambda_http` and found `lambda_web`. I was looking for something to compare `lambda_http::run` against.

```
use axum::{routing::get, Json, Router};
use lambda_web::{run_hyper_on_lambda, LambdaError};
use serde_json::{json, Value};

#[tokio::main]
async fn main() -> Result<(), LambdaError> {
    let app = Router::new().route("/prod/hello", get(hello));

    run_hyper_on_lambda(app)
        .await
        .expect("something went wrong in the Lambda runtime");

    Ok(())
}

async fn hello() -> Json<Value> {
    Json(json!({ "message": "hello other stranger!"}))
}
```

And these were my results (spoiler they are the same as yours)

```
☁  curl https://<http-api-id>.execute-api.us-east-1.amazonaws.com/prod/hello
{"message":"hello other stranger!"}% 
☁  curl https://<rest-api-id>.execute-api.us-east-1.amazonaws.com/prod/hello
eyJtZXNzYWdlIjoiaGVsbG8gb3RoZXIgc3RyYW5nZXIhIn0=%
```

 ... I'm continuing on.

bnusunny: @hugobast @david-perez This is a working but not an optimal solution. 

It is totally valid to treat all Body data as binary data. aws_lambda_events crate will automatically serialize them with Base64 encoding (code [here](https://github.com/LegNeato/aws-lambda-events/blob/2f2221e8cd3c5f3cd584cd1548fb32a6cc874ba9/aws_lambda_events/src/encodings.rs#L297)). Why API Gateway REST API seems not working? You need to configure `binaryMediaTypes` for Lambda Proxy integration (doc [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html#api-gateway-payload-encodings-proxy)). API Gateway HTTP API is smarter. It will automatically decode the body if `isBase64Encoded` is 'true'. 

But this is not an optimal solution. Large part of Lambda responses are not binary data. We waste CPU cycles to encode on Lambda and decode on API Gateway. It hurts performance and increase latency. Another problem is the payload size. Lambda function has 6MB payload limit for both request and response. Base64 encoding increases the payload size by about 30%. This means Lambda will throw an error if the actual returned data is larger than ~4.6MB. 

My suggestion is to use ALB rules to detect the binary data and text data. For text data, pass them through as `Body::Text`, and for the rest, treat them as binary data. We can avoid the performance penalty and payload size issue for most common use cases.

david-perez: @bnusunny 

> You need to configure `binaryMediaTypes` for Lambda Proxy integration (doc [here](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html#api-gateway-payload-encodings-proxy))

`binaryMediaTypes` is indeed one setting that determines whether the client sees base64-encoded data or not, but it's not that simple. There's actually three "parameters" that come into play, [as I mentioned before](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings-workflow.html).

> The combination of your API's `binaryMediaTypes`, the headers in client requests, and the integration `contentHandling` property determine how API Gateway encodes payloads.

---

> It is totally valid to treat all body data as binary.

Not really. I don't think it's fair to impose on users of `aws-lambda-rust-runtime` that their responses will _always_ be signalled as binary, and force them into fiddling with their `binaryMediaTypes`/`contentHandling` REST API settings to get text responses working correctly again. They should only have to fiddle with those settings when they are _actually_ returning binary payloads.

There's also the performance penalty concern that you raise of course.

---

The crux of the problem is that `aws-lambda-rust-runtime` can only _guess_ as to what the user's intention is when they send a response from a `tower::Service`. `aws-lambda-rust-runtime` also does not know how the user's REST API is configured. For all of the above, I think there's no way for `aws-lambda-rust-runtime` to provide a "one-size-fits-all" solution.

david-perez: @hugobast @calavera @bnusunny 

To move forward with landing this valuable feature, I propose that we implement a "good enough" solution that will work for _most_ cases. As you note, following the ALB rules to determine whether the response contains binary data or not is one heuristic. So, to be more concrete, looking at [these rules](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/lambda-functions.html#receive-event-from-load-balancer) for _requests_, my proposal is that we implement the same algorithm to determine whether a _response_ contains binary data or not:

1. If the `content-encoding` header is present, base64-encode the body and set `isBase64Encoded` to `true`.
2. If the `content-encoding` header is _not_ present:
    1. If the `content-type` header is present:
        1. If it is equal to `text/*`, `application/json`, `application/javascript` or `application/xml`, leave the body as is and set `isBase64Encoded` to `false`.
        2. Otherwise, base64-encode the body and set `isBase64Encoded` to `true`.
    2. If the `content-type` header is _not_ present, leave the body as is and set `isBase64Encoded` to `false`.

We should detail the heuristic in the docs, and _loudly_ call out that this heuristic might not support _all_ use cases. But at least it will support _most_ use cases. I myself I'm struggling to come up with an example of a well-designed API whose responses would be broken by this heuristic while writing this text. But, there are plenty of badly-designed APIs out there and misbehaving clients.

In a follow-up PR, we can consider ways in which the service implementer can _explicitly_ convey whether they want `aws-lambda-rust-runtime` to base64-encode their response body and/or set `isBase64Encoded`. An idea that I proposed above is to use [`http::Extensions`](https://docs.rs/http/latest/http/struct.Extensions.html) as a "side-channel" to allow the user to communicate with the runtime and provide their intent.

hugobast: @david-perez @calavera I'll review this carefully tomorrow and go about implementing this solution, thank you

bnusunny: @david-perez 

`lambda-http` crate only supports Lambda Proxy integration, not non-proxy integrations. It makes the configuration for binary media type support simpler: adding `binaryMediaTypes=*/*` to the REST api will make it work. API GW developer guide has [the detail](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html).

bnusunny: For the rules. I suggest two changes: 

> 2.i.a If it is equal to text/*, application/json, application/javascript or application/xml, leave the body as is and set isBase64Encoded to false.

`content-type` may have suffixes, such as 'text/html; charset=UTF-8'. So it is better to use the following:

2.i.a If the `content-type` **begins** with `text/*`, `application/json`, `application/javascript` or `application/xml`, leave the body as is and set `isBase64Encoded` to `false`. 

> 2.ii If the `content-type` header is not present, leave the body as is and set `isBase64Encoded` to `false`.

This is problematic. If a bad client sends http response with binary data without `content-type` header, Lambda function will through error. We should default to binary here. ALB also uses binary as default. 

2.ii If the `content-type` header is not present, base64 encode the body and set `isBase64Encoded` to `true`.

bnusunny: > @bnusunny I think your point needs to be documented in the runtime documentation, but we should not make users depend on specific configurations in another service. binaryMediaTypes are for example not available when you call a function with `invoke`. People that use that should still be able to use http responses as expected.

Yes. We definitely need to document this configuration.

david-perez: > 2.ii If the `content-type` header is not present, leave the body as is and set `isBase64Encoded` to `false`.

> This is problematic. If a bad client sent a binary http response without the `content-type` header, we would treate it as text and Lambda service would throw an error.

However, if they send a text HTTP response without the `content-type` header, the client sends e.g. `Accept: text/plain` and the user has not configured `contentHandling` (which they shouldn't need to, since they are returning text payloads) we would treat it as binary and return a base64-encoded response. That is, the client would receive `eyJtZXNzYWdlIjoiaGVsbG8gc3RyYW5nZXIhIn0=` instead of `{"message":"hello stranger!"}`.

I think APIs that return text are far more common than those that return binary payloads and as such we should default to text in this case where we don't have any information. I also prefer the Lambda service throwing a loud error than the client being puzzled at why they received base64-encoded text.

bnusunny: Make sense. Let's default to text. 

By the way. `lambda_http` crate only supports Lambda Proxy integration with API GW. `contentHandling` doesn't apply to Lambda Proxy integration. APIGW Developer Guide has the [details](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-payload-encodings.html).

lambda-http/src/response.rs
```diff
@@ -87,71 +91,137 @@ impl LambdaResponse {
...
+{
+    Box::pin(async move { Body::from(to_bytes(body).await.expect("unable to read bytes from body").to_vec()) })
+}
+
+fn convert_to_text<B>(body: B) -> BodyFuture
+where
+    B: HttpBody + Unpin + 'static,
+    B::Error: fmt::Debug,
+{
+    // assumes utf-8
```
hugobast, line 203: I've pushed changes that should be close, at least in spirit, to what was talked about in the last few days in the other thread, I'm not that comfortable with Rust so please make suggestions if the code here is suboptimal.

I have concerns about encoding @david-perez @bnusunny @calavera. I'm guessing that most things will be utf-8 (JSON?) but that's far from saying that everything is utf-8.

@bnusunny commented with an [example code that's using reqwest](https://github.com/awslabs/aws-lambda-web-adapter/blob/18db3940716d4cadcfb5cddc5326b3d11b70e1c1/src/main.rs#L90-L118) so I looked at how the [`text` fn is implemented](https://docs.rs/reqwest/latest/src/reqwest/async_impl/response.rs.html#184-207) on `request::Response`. We could use the same approach here and parse the charset from the content-type if we'd like I'll defer to your comments/opinions on this.

hugobast: It's optional but here's a commit where encode based on the charset of the content-type, falling back to utf-8 in error cases: https://github.com/awslabs/aws-lambda-rust-runtime/pull/491/commits/2632aad27b92930a1267db6c03142a77ebc915f8

hugobast:
@calavera @bnusunny @david-perez is anything more needed here to get this merged?

hugobast:
> When Nicolas implemented this originally, he was thinking about a future where Lambda could stream responses, but that might be better to put it into a separated PR when/if that happens.

Right, forgot about this, I can push up a version that removes the async response, I agree that tackling this as a separate PR is gonna be clearer.

hugobast:
In my hasty response I'm afraid I won't get the point accross so I'll summarise here in a new comment.

What I mean by this previous reply is reading a body is an asynchronous operation and that's why it looks like Nicolas made the response into a Future.

bnusunny:
@hugobast You are right. It is an asynchronous operation. Thanks very much for getting this PR done.

bnusunny (approved):

PR #497, opened 2022-07-02
Feature flags for lambda_http

### LINKED ISSUE #481: Improving Compile Times And/OR Pluggable Deserializable Types
Hey! First let me say that I love this project. Clearly a lot of thoughtful design work has been put in. Kudos to all the contributors. Sorry about the wall of text in advance.

My scenario is this: I’ve noticed that even debug incremental recompiles of an absolutely barebones `lambda_http` app take upwards of 15 seconds. That doesn’t sound like much but multiple this by 10x lambdas in a workspace and it gets out of hand pretty quick.

I did some cursory investigation (the results of which I can post later). My investigation indicates that the slowness comes during the `codegen_crate` and subsequent LLVM optimize steps of the build.

Checking the number of emitted lines of IR, I see that it’s in the 4M+ lines range! This is a lot of code for LLVM to optimize, so it explains why both the steps take so long.

I took a lot at where there is so much IR being generated but I couldn’t find any obvious pathological cases. It does appear that the majority of the IR is `serde::Deserialize` impls on the `aws_lambda_events` types used by this crate.

I took a look at those types, and they have a heck of a lot of `serde_deserialize_with`. Which isn’t too expensive on its own, but multiplied across every field of every struct that’s a lot of codegen. There was also some duplicate code via the fact that `#[derive(Deserialize)]` implements `visit_seq` along with `visit_map`. The former is unnecessary because we’re dealing with JSON and it is strictly invalid in most cases to allow users to deserialize a JSON array into a struct.

So I just cut the `visit_seq` code out entirely via a custom `#[derive(Deserialize)]` impl. That plus stripping down this crate’s `LambdaRequest` enum to only cover the cases I care about let me bring the LLVM IR count down from 4.5M to about 3.5.

That was it for the low-hanging fruit. The remaining deserialization code seems necessary, but is a bit bloated. I could probably write it out by hand to be more efficient, or improve the existing codegen of the `#[derive(Deserialize)]` macro.

But I’ve been rethinking my approach. **The problem is that we’re re-running the `#[derive(Deserialize)]` macro every time the lambda binary compiles.** There are some efforts to [disable derives when running cargo check](https://github.com/async-graphql/async-graphql/issues/783#issuecomment-1067460045). This is clever for the purposes of improving type hint resolution times, etc. But I don’t think it’s sufficient. What’s needed is a true way to cache the results of prof macro evaluations, if said proc macros are pure. This finally brings me to my first question: **Does anyone know of a way to cache the results of a proc-macro evaluation between builds?**

Alternatively: **Is it possible to force rustc to evaluate a proc macro for a struct, *before* said struct is used in a leaf crate?**

cc @LegNeato  @calavera as maintainers of the aws-lambda-events crate, any ideas? Perhaps we could **Selectively cargo expand the `Deserialize` impls in the aws-lambda-events crate?** It would result in a much larger source file, that could cut out at least some of the `crate_codegen` time. Not sure how much but it’d be a decent chunk of the 40% it seems to take now.

cc @calavera @nmoutschen as maintainers of this crate, any ideas? Perhaps we could refactor things a bit so that users can provide their own APIGW types and deserialize on their end? That way there’s a bit more flexibility. For example I could fork aws-lambda-events, strip it to just the types I need, and hand roll `Deserialize` impls. At the moment I’m locked in to the types this crate provides. I understand that this crate does this because it handles a lot of the boilerplate when it comes to talking to AWS, but I’m fine with my crate handling some of that.

calavera:
Add feature flags to lambda_http so consumer can decide which service their events come from. This makes compilation much faster when you don't put the same Lambda function behind multiple services.

I've tested these changes in a simple function behind APIGW, and compilation time for a pre-compiled function drops from 18 seconds to 5 seconds.

*Issue #, if available:*

Fixes #481 

### DIFF
--- a/README.md
+++ b/README.md
@@ -374,7 +374,7 @@ fn main() -> Result<(), Box<Error>> {
 
 ## Feature flags in lambda_http
 
-`lambda_http` is a wrapper for HTTP events coming from three different services, Amazon Load Balancer (ALB), Amazon Api Gateway (APIGW), and AWS Lambda Function URLs. Amazon Api Gateway can also send events from three different endpoints, REST APis, HTTP APIs, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
+`lambda_http` is a wrapper for HTTP events coming from three different services, Amazon Load Balancer (ALB), Amazon Api Gateway (APIGW), and AWS Lambda Function URLs. Amazon Api Gateway can also send events from three different endpoints, REST APIs, HTTP APIs, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
 
 By default, `lambda_http` compiles your function to support any of those services. This increases the compile time of your function because we have to generate code for all the sources. In reality, you'll usually put a Lambda function only behind one of those sources. You can choose which source to generate code for with feature flags.
 


README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
 }
 ```
 
+## Feature flags in lambda_http
+
+`lambda_http` is a wrapper for HTTP events coming from two different services, Amazon Load Balancer (ALB), and AWS Api Gateway (APIGW). AWS Api Gateway can also send events from three different endpoints, Proxy V1, Proxy V2, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust http semantics into your Lambda functions.
```
david-perez, line 377:
```suggestion
`lambda_http` is a wrapper for HTTP events coming from two different services, Amazon Load Balancer (ALB), and Amazon API Gateway (APIGW). Amazon API Gateway can also send events from three different endpoints, Proxy V1, Proxy V2, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
```

The official name is "Amazon API Gateway".

README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
...
+
+By default, `lambda_http` compiles your function to support any of those services, this increases the compile time of your function because we have to generate code for all the sources. In reality, you'll usually put a Lambda function only behind of those sources. You can choose which source to generate code for with feature flags.
```
david-perez, line 379:
```suggestion
By default, `lambda_http` compiles your function to support any of those services. This increases the compile time of your function because we have to generate code for all the sources. In reality, you'll usually put a Lambda function only behind one of those sources. You can choose which source to generate code for with feature flags.
```

README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
...
+
+The available features flags for `lambda_http` are the following:
+
+- `alb`: for events coming from Amazon Load Balancer.
```
david-perez, line 383: ELB Application Load Balancer?

It'd be nice if we added links to each product.

README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
...
+- `apigw_v1`: for events coming from API Gateway Proxy V1.
+- `apigw_v2`: for events coming from API Gateway Proxy V2.
+- `apigw_websockets`: for events coming from API Gateway WebSockets.
+
+If you only want to support one of these sources, you can disable the default features, and enable only the source that you care about in your package's Cargo.toml file. Substitute the dependency line for `lambda_http` for the snippet below, changing the feature that you want to enable:
```
david-perez, line 388:
```suggestion
If you only want to support one of these sources, you can disable the default features, and enable only the source that you care about in your package's `Cargo.toml` file. Substitute the dependency line for `lambda_http` for the snippet below, changing the feature that you want to enable:
```

README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
...
+
+```toml
+[dependencies.lambda_http]
+version = "0.5.3"
+default-features = false
+features = ["apigw_v1"]
+```
+
+This will make your function to compile much faster.
```
david-perez, line 397:
```suggestion
This will make your function compile much faster.
```

david-perez (approved):
Implementation looks good. Minor comments about wording in the readme.

README.md
```diff
@@ -378,6 +372,30 @@ fn main() -> Result<(), Box<Error>> {
 }
 ```
 
+## Feature flags in lambda_http
+
+`lambda_http` is a wrapper for HTTP events coming from two different services, Amazon Load Balancer (ALB), and Amazon Api Gateway (APIGW). Amazon Api Gateway can also send events from three different endpoints, Proxy V1, Proxy V2, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
```
bnusunny, line 377: `lambda_http` actually supports three sources: ALB, APIGW, and the new [Lambda Function URLs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html). Lambda Function URLs use the same event format as APIGW HTTP API [payload format version 2.0](https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html#http-api-develop-integrations-lambda.proxy-format). So it works out of box.

bnusunny (changes requested):

README.md
```diff
@@ -374,15 +374,15 @@ fn main() -> Result<(), Box<Error>> {
 
 ## Feature flags in lambda_http
 
-`lambda_http` is a wrapper for HTTP events coming from two different services, Amazon Load Balancer (ALB), and Amazon Api Gateway (APIGW). Amazon Api Gateway can also send events from three different endpoints, Proxy V1, Proxy V2, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
+`lambda_http` is a wrapper for HTTP events coming from three different services, Amazon Load Balancer (ALB), Amazon Api Gateway (APIGW), and AWS Lambda Function URLs. Amazon Api Gateway can also send events from three different endpoints, REST APis, HTTP APIs, and WebSockets. `lambda_http` transforms events from all these sources into native `http::Request` objects, so you can incorporate Rust HTTP semantics into your Lambda functions.
```
bnusunny, line 377: minor one: `REST APis` -> `REST APIs`

bnusunny (changes requested):

bnusunny (approved):

PR #520, opened 2022-08-31
Report error when we cannot deserialize the payload.

### LINKED ISSUE #517: Invalid body panics the runtime
[this line](https://github.com/awslabs/aws-lambda-rust-runtime/blob/fd2ea235f4adb166422ca5b612c4a1a42a626daa/lambda-runtime/src/lib.rs#L124) makes the process exit with status 1 when the request body is malformed (e.g. there are missing fields).

it should instead send back an `EventErrorRequest` with the appropriate error details and continue processing

calavera:
*Issue #, if available:*

Fixes #517 

*Description of changes:*

Instead of panic, capture the error, and report it to the Lambda api. This is more friendly to operate.


bnusunny (approved):
LGTM

PR #538, opened 2022-10-06
Check that `lambda_http`s inner service is ready before calling it

dcormier:
This allows for service [backpressure](https://docs.rs/tower/0.4.13/tower/trait.Service.html#backpressure).

Additionally:

> Services are permitted to panic if `call` is invoked without obtaining `Poll::Ready(Ok(()))` from `poll_ready`.

[Source](https://docs.rs/tower/0.4.13/tower/trait.Service.html#be-careful-when-cloning-inner-services)

### DIFF
--- a/lambda-integration-tests/src/bin/runtime-trait.rs
+++ b/lambda-integration-tests/src/bin/runtime-trait.rs
@@ -1,9 +1,11 @@
-use lambda_runtime::{Error, LambdaEvent, Service};
-use serde::{Deserialize, Serialize};
 use std::{
     future::{ready, Future},
     pin::Pin,
+    sync::atomic::{AtomicBool, Ordering},
 };
+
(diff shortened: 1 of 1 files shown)


dcormier:
Alternatively, I could tweak the existing `http-trait` test to include checking that `.poll_ready()` is properly called.

dcormier:
`lambda_runtime` already does this properly, but should tests be added to ensure it?

dcormier:
`lambda_extension` does not do this properly.

dcormier:
I can take a swing at it. Want it in this PR or another one?

dcormier:
I added it here. I updated the existing integration test to account for this.

dcormier:
Alright. That's all done. Thank you for the feedback.

PR #562, opened 2022-11-05
http-dynamodb example

jknoetzke:
*Issue #, if available:*
This is an example of a rust lambda function that takes in a JSON payload and writes to a Dynamodb table.

*Description of changes:*
First commit.

### DIFF
--- a/examples/http-dynamodb/Cargo.toml
+++ b/examples/http-dynamodb/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "skunks"
+name = "http-dynamodb"
 version = "0.1.0"
 edition = "2021"
 
--- a/examples/http-dynamodb/Cargo.toml
+++ b/examples/http-dynamodb/Cargo.toml
@@ -12,6 +12,7 @@ edition = "2021"
 
 [dependencies]
 serde_json = "1.0"
+serde = { version = "1.0", features = ["derive"] }
 lambda_http = { path = "../../lambda-http" }
 lambda_runtime = { path = "../../lambda-runtime" }
 aws-sdk-dynamodb = "0.21.0"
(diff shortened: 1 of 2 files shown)


jknoetzke:
Hello

Thanks for all the useful comment.

I have attempted to implement all of the requested changes.
There is one requested change I had difficulty with and this is regards to returning the struct as JSON to the user after it was pushed into the table.

In the end, I created two Item structs using serde, one that gets sent to Dynamo and one that is sent back to the user. I doubt this is the correct way. However, add_item requires ownership of the struct and with serde (serialize , deserialize) I could not figure out a way to implement the Copy or Move trait.. The compiler was not allowing me to reuse Item to return to the user as it had been changed.. So, I implemented this hack instead.

I would also like to return a 400 error if the rest call made by the user is missing parameters.. Now the function crashes and returns a 500. Any ideas how to do this ? I see examples that you wrote in how to return errors.. But I can't figure out how to determine when serde realizes the request is missing a field and to return the missing field(s) back to the user.. 

Thanks.. getting there.

jknoetzke:
Code was formatted.
Implemented clone on the struct.
Removed duplicate variable for the struct.

jknoetzke:
Thanks. I had tried adding Clone to derive, and the compiler wasn't having it.. I thought maybe serde was somehow blocking Clone.. but it looks like I just had the syntax wrong. ;-)

I've made the change.

There is one last thing, should I be checking if the input request is correctly formatted and missing fields ? Right now, if you send a request with a missing field, the function crashes and returns a 500. 

I can see how to return an error from your examples, but I can't figure out how to pull out the errors from serde when calling let item: Item = serde_json::from_str(s).map_err(Box::new)?;

jknoetzke:
Check if JSON is valid. If not, return 400 and abort. 
Returns 200 if record is inserted into table.

PR #577, opened 2022-12-15
Add tracing span with request id to the handler

llgerard:
When logging errors, it is very useful to log the request Id (and xray id if available). But, not every one might want it, so we probably don't want to include it by default in every log.

The solution proposed in this PR is to use a `span` with fields containing the request id and xray id. Then, to get the request id part of the log messages, one has to enable logging spans in the logger setup (using [tracing_subscriber](https://docs.rs/tracing-subscriber/latest/tracing_subscriber/fmt/struct.Layer.html#method.with_span_list) for example).

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -26,7 +26,7 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_stream::{Stream, StreamExt};
 pub use tower::{self, service_fn, Service};
 use tower::{util::ServiceFn, ServiceExt};
-use tracing::{error, trace};
+use tracing::{error, span, trace, Instrument};
 
 mod requests;
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


lambda-runtime/src/lib.rs
```diff
@@ -138,55 +129,80 @@ where
...
-                Ok(body) => body,
-                Err(err) => {
-                    let req = build_event_error_request(request_id, err)?;
-                    client.call(req).await.expect("Unable to send response to Runtime APIs");
-                    return Ok(());
-                }
+            let request_span = match xray_trace_id {
+                Some(trace_id) => tracing::span!(
+                    tracing::Level::TRACE,
+                    "Lambda request",
```
llgerard, line 135: I am really open to a better span name!

llgerard:
@calavera I had it formatted already. I reran `cargo fmt` when changing the name (commit just above), but nothing was changed.

Which part looks improperly formatted?

PR #583, opened 2023-01-06
Added `Context::deadline(&self) -> SystemTime`

dcormier:
*Description of changes:*

Added a function to get the `SystemTime` value for the deadline of the context.

There have been multiple times that I've needed a `SystemTime` (or similar) value, rather than simply the epoch ms `u64` that `Context` currently exposes. And I've watched other devs trip over that existing `u64` value. This should make it easier to use the deadline.


dcormier:
Thanks @calavera!

Issue #594, opened 2023-02-02
lambda_http::request::from_str() ignores URI paths and Params


bsodmike:
I just tried running some tests but the request.uri only contains the TLD entry. Even the first path, /foo/ is missing.

Test setup:

```rs
#[cfg(test)]
mod test_bug {
    use crate::errors::Error;
    use lambda_http::{Body, Request, Response};
    use tracing_test::traced_test;

    async fn test_handler(event: Request) -> Result<Response<Body>, Error> {
        let uri = event.uri();

        // FIXME: Tests do not see the full URI
        tracing::info!("Request URI: {:?}", uri);

        Ok(Response::builder()
            .status(200)
            .body(format!("URI: {}", uri).into())
            .map_err(Box::new)?)
    }

    #[tokio::test]
    #[traced_test]
    async fn test_params_missing() {
        let proxy_request = format!(
            r#"
        {{
            "method": "GET",
            "uri": "https://example.com/test?name=bob",
            "version": "HTTP/1.1",
            "headers": {{
              "accept": "*/*"
            }},
            "body": ""
          }}
        "#
        );

        let request =
            lambda_http::request::from_str(proxy_request.as_str()).expect("created request");

        let response = test_handler(request).await.expect("provided response");
        match response.body() {
            Body::Text(text) => {
                assert_eq!(text, &r#"URI: /"#.to_string())
            }
            _ => tracing::error!("Err: Body missing"),
        };

        // This fails
        assert!(logs_contain("Request URI: /test?name=bob"));
    }
}
```

bsodmike:
@calavera Sorry, I didn't have time to troubleshoot.  I've updated the test case above with a working example. Thanks!

bsodmike:
Please ignore. the format of the JSON request was incorrect - thanks!

PR #595, opened 2023-02-03
Add test to check if params are missing

bsodmike:
https://github.com/awslabs/aws-lambda-rust-runtime/issues/594

Added a test to check if it fails within the test suite, however, it passes.  In https://github.com/awslabs/aws-lambda-rust-runtime/issues/594 we see it fails.


bsodmike:
@calavera I'm missing something, as I can't seem to get this to fail within the existing tests?

bsodmike:
Please ignore, this has been resolved.

PR #598, opened 2023-02-08
add example for Lambda + HTTP + Axum

greenwoodcm:
this example bridges Lambda's HTTP support, which is based on the `tower::Service` trait, and the axum web framework, which builds applications that are also based on the `tower::Service` trait.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/examples/http-axum/Cargo.toml
+++ b/examples/http-axum/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "http-basic-lambda"
+version = "0.1.0"
+edition = "2021"
+
+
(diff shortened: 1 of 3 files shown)


hxk1633:
How do I access the request context of lambda events in the Axum handlers?

greenwoodcm:
> How do I access the request context of lambda events in the Axum handlers?

you should be able to use the `lambda_http::RequestExt` trait and then use the following method to extract the Lambda context from the `Request` struct.  maybe i'll push an example that demonstrates this.

https://docs.rs/lambda_http/latest/lambda_http/ext/extensions/trait.RequestExt.html#tymethod.lambda_context_ref

Issue #599, opened 2023-02-12
data did not match any variant of untagged enum LambdaRequest


winstxnhdw:
> I have rewritten the issue using your examples instead of my implementation to help with reproducibility.

Hi, I have been trying to write a simple Lambda function in Rust with the Function URL according to one of your examples [here](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/basic-lambda/src/main.rs).

Testing the Lambda locally with `cargo lambda` correctly returns `Command do something executed.`

```bash
cargo lambda invoke something --data-ascii '{ "command": "do something" }'
```

There were no issues when executing the Lamba Test Event in the web GUI either. The problem only comes after deployment. Sending the exact POST request with Insomnia always returns an InternalServerError. CloudWatch complains of the following.

```rs
ERROR Lambda runtime invoke{requestId="5fb1fd9b-e8cd-48a4-b641-caf1ff16529b" xrayTraceId="Root=1-63e9481d-6db35e9e5220a0d061dd7904;Parent=0d074a1c13e2a3e8;Sampled=0"}: lambda_runtime: Error("missing field `command`", line: 1, column: 1118)
```

Okay fine, so I tried the example written in the `lambda-http` documentation [here](https://github.com/awslabs/aws-lambda-rust-runtime/tree/main/lambda-http). I tested this again with Insomnia and CloudWatch now complains of the following.

```rs
thread 'main' panicked at 'called Option::unwrap() on a None value'
```

At this point, I was getting pretty desperate and did a fresh generation of `cargo lambda new` with the Function URL template selected. Unsurprisingly, this was just an exact copy of your `http-basic-lambda` example [here](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/http-basic-lambda/src/main.rs). I did the same test with `cargo lambda invoke` as above and this time it returned the following.

```rs
Error: serde_json::error::Error

  × data did not match any variant of untagged enum LambdaRequest
```

This is my Cargo.toml.

```toml
[dependencies]
tokio = { version = "^1.25.0", features = ["macros"] }
reqwest = { version = "^0.11.14", default-features = false, features = ["json", "rustls-tls"] }
lambda_http = { version = "^0.7.3", default-features = false, features = ["apigw_http"] }
tracing-subscriber = { version = "^0.3", default-features = false, features = ["fmt"] }
tracing = { version = "^0.1", features = ["log"] }
[...]

PR #601, opened 2023-02-17
docs/refactor: add documentation for invoking HTTP functions locally

winstxnhdw:
# Summary

As mentioned in #599, I have added some light documentation for invoking the HTTP Lambda functions locally. Additionally, I have fixed all of the non-visual markdown lint warnings/errors. I did not fix all of them to preserve the styling of the original authors in the case that it was done intentionally.

I have also converted `Utc.ymd` to the non-deprecated `Utc.with_ymd_and_hms`. Tested and functional.

## Explanation

When a reader looks at [Deploying with Cargo Lambda](https://github.com/awslabs/aws-lambda-rust-runtime#22-deploying-with-cargo-lambda), the reader has likely not read the official `cargo-lambda` documentation. It is then possible that there may be an assumption, through disjunction, that it is possible to execute `cargo lambda invoke` without the `--remote` flag.

The user may then try this command on the first given [example](https://github.com/awslabs/aws-lambda-rust-runtime#22-deploying-with-cargo-lambda) to find that the command runs successfully. Later on, they may write a HTTP function and assume that the command is a signal of correctness, only to find that the command fails.

Inevitably, after several futile attempts at debugging,  the user may come back to the documentation and to notice that there is no documentation that describes their issue until they finally choose to look at the `cargo-lambda` documentation where they must notice the following in a glance, along with the context of the first paragraph.

image

With the updated documentation, the user will likely perform a search for `data did not match any variant of untagged enum LambdaRequest` and quickly realise their blunder.

# Disclaimer


README.md
```diff
@@ -331,9 +328,38 @@ fn test_my_lambda_handler() {
...
+cargo lambda watch -a 127.0.0.1 -p 9001
+```
+
+Now you can use the `cargo lambda invoke` to send requests to your function. For example:
+
+```bash
+cargo lambda invoke <lambda-function-name> --data-ascii '{ "command": "hi" }'
+```
+
+Running the command on a HTTP function (Function URL, API Gateway, etc) will not work, presenting you with the following error.
```
bnusunny, line 343: This could work if the event data matches the right schema. Here is a list of events used for testing in lambda-http crate. 
https://github.com/awslabs/aws-lambda-rust-runtime/tree/main/lambda-http/tests/data

bnusunny (changes requested):

README.md
winstxnhdw: You're right. I have the amendments.

bnusunny (approved):

PR #604, opened 2023-02-18
Fix invoke example for Windows CMD

ed-muthiah:
*Description of changes:*

Currently the invoke example `cargo lambda invoke rust-lambda --data-ascii '{ "command": "hi" }'` does not work on Windows and returns the following error because of the way that Windows handles single quotes. Trying to invoke the command as shown in the readme returns: `error: Found argument 'command:' which wasn't expected, or isn't valid in this context`

This change the example invoke command to run for Windows, Mac and Linux.

Edit: Updated the PR name to include CMD

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,7 +354,7 @@ An simpler alternative is to cURL the following endpoint based on the address an
 curl -v -X POST \
   'http://127.0.0.1:9001/lambda-url/<lambda-function-name>' \
   -H 'content-type: application/json' \
-  -d '{ "command": "hi" }'
+  -d "{\"command\": \"hi\" }"
 ```
 
 > **warning** Do not remove the `content-type` header. It is necessary to instruct the function how to deserialize the request body.
--- a/README.md
+++ b/README.md
@@ -69,8 +69,6 @@ async fn func(event: LambdaEvent<Value>) -> Result<Value, Error> {
 ```
 
 ## Building and deploying your Lambda functions
-> **Note**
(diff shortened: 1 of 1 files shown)


winstxnhdw:
This isn't a Windows-specific issue, but a shell-specific issue. If we do go ahead with this change, please update the rest of the examples to follow this convention.

ed-muthiah:
Done 👍🏽

bnusunny:
I recommend to keep them seperated, and show examples for both Linux and Windows.

winstxnhdw:
> I recommend to keep them seperated, and show examples for both Linux/MacOS and Windows.

What’s the benefit to doing this? Escaping the quotations will work on most modern shells.

bnusunny:
Linux and MacOs users could use a simpler format.

winstxnhdw:
How can we do this without cluttering the docs?

bnusunny:
Add one section to talk about it. There is an example in [AWS CLI doc](https://docs.amazonaws.cn/en_us/cli/latest/userguide/cli-usage-parameters-quoting-strings.html).

ed-muthiah:
Done 👍🏽 Let me know if it's in an appropriate spot.

bnusunny:
I suggest to put this note close to the first apperance of `cargo lambda invoke`.

ed-muthiah:
Done 👍🏽

bnusunny (approved):

PR #609, opened 2023-02-28
make tracing init consistent across examples

greenwoodcm:
this just copies the code across examples so that we're initializing the tracing subscriber in a consistent way.  we want to set a max level, disable printing the name of the module every time, disable ANSI color codes (because CloudWatch prints them in a non-readable way), and disable logging the time at this layer because CloudWatch is going to add the ingestion time itself.


bnusunny (approved):
LGTM

Issue #614, opened 2023-03-07
Memory allocation during LambdaEvent::new()


timClicks:
We currently convert a `&str` to a `String` when a `LambdaEvent` is created. Presumably this doesn't need to be re-created each invocation of the Lambda function.

https://github.com/awslabs/aws-lambda-rust-runtime/blob/0f6e2a285677c00c35d65250166fa99cd56d3d68/lambda-extension/src/events.rs#L67

maxday:
👋 here is a proposal to fix this issue, coming with a breaking change unfortunately
I could not find a way to avoid copying and still using a `String` in `LambdaEvent`

timClicks:
Thanks for your input @maxday :D

PR #615, opened 2023-03-09
feat: use rc to avoid allocating a new string for extension_id

maxday:
*Issue #, if available:* potential fix for #614

*Description of changes:*
Using a RC around the extension id. 
Now it clones the RC for each event which is way lighter than cloning the full extension_id string

⚠️ this is a breaking change as LambdaEvent is public

### DIFF
--- a/lambda-extension/src/extension.rs
+++ b/lambda-extension/src/extension.rs
@@ -1,5 +1,6 @@
 use std::{
-    convert::Infallible, fmt, future::ready, future::Future, net::SocketAddr, path::PathBuf, pin::Pin, sync::Arc,
+    convert::Infallible, fmt, future::ready, future::Future, net::SocketAddr, path::PathBuf, pin::Pin, rc::Rc,
+    sync::Arc,
 };
 
 use hyper::{server::conn::AddrStream, Server};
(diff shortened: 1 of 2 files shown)


lambda-extension/src/extension.rs
```diff
@@ -221,6 +222,7 @@ where
 
         let extension_id = register(client, self.extension_name, self.events).await?;
         let extension_id = extension_id.to_str()?;
+        let extension_id_rc = Rc::new(String::from(extension_id));
```
greenwoodcm, line 225: isn't `String::from(&str)` an allocation anyways?  can we make the `extension_id` field an `&str`?

maxday: - yes you're right but it's done only once (outside the loop) instead of doing it for every received events
- I tried to make it a `&str` but this introduces a lot of lifetime-related code, was a bit less readable (IMHO) and also introduces a breaking change that's why I went with the `Rc<String>` path

maxday: what do you think @greenwoodcm?

greenwoodcm: ok i think that seems reasonable.  is this change non-breaking?

maxday: no it's not, I put a warning in the description but this basically modifies the public `extension_id` field in `LambdaEvent` from `String` to `Rc<String>`

maxday: does it make sense to you @greenwoodcm ?

Issue #616, opened 2023-03-12
how to initialize SDK client outside of function handler


peterborkuti:
Dear Developers,

I created an example for s3:
https://github.com/peterborkuti/aws-lambda-rust-runtime/commit/dbc5d3dffa6b7fb1504702c4a4f103d2cd8aa687

It works, but as I read aws best practices for lambda:
https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html
**"Initialize SDK clients and database connections outside of the function handler"**

I could not do it (I am absolute beginner in Rust but I have solid knowledge in Java and JavaScript)

### creating static/const  but client creation is async
```const client: Client = get_client().await;```
### static mut
```
static mut client : Option<Client> = None;
...
// in main:
client = Some(get_client().await); // but it is unsafe
```
### using lambdas context in main
```
pub(crate) async fn function_handler<F1, Fut1, F2, Fut2>(fn_get_file: F1, fn_put_file: F2, event: LambdaEvent<S3Event>) -> Result<(), Error> 
    where
        F1: Fn(&str, &str) -> Fut1,
        Fut1: Future<Output = Option<Cursor<Vec<u8>>>>,
        F2: Fn(&str, &str, ByteStream) -> Fut2,
        Fut2: Future<Output = ()>
     {
```
(With the above handler testing would be very easy, I think, because the client is "embedded" in the fn_get_file and fn_put_file functions, so when testing we do not need sdk client at all)

in main I tried to binding the client parameters to fn_get_file and fn_put_file:

```
    let client = get_client().await;

    let clientRef = |x:Client| &x;

    let clientBroker = move || clientRef(client); 

    let fn_get_file = move |bucket: &str, key: &str| get_file(&clientBroker(), bucket, key);
    let fn_put_file = |bucket: &str, key: &str, bytes: ByteStream| put_file(&clientBroker(), bucket, key, bytes);
    let fn_function_handler = |event: LambdaEvent<S3Event>| function_handler(fn_get_file, fn_put_file, event);  

    run(service_fn(fn_function_handler)).await
    
```
First I had lifetime issues (tried to solve with "move" and now I got:"expected a closure that implements the `Fn` trait, but this closure only implements `FnOnce`":

Do you have any ideas how to do it?

Thank you in advance
Péter

greenwoodcm:
the following example does something similar, can you model your approach off of this?

[...]

PR #619, opened 2023-03-14
basic-sdk example

greenwoodcm:
adds an example for a basic Lambda function that calls S3.  this demonstrates the best practice of instantiating an SDK client during function initialization (main method) and reusing that client across multiple invokes.

*Issue #, if available:*
#616

### DIFF
--- a/examples/basic-sdk/Cargo.toml
+++ b/examples/basic-sdk/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "basic-sdk"
+version = "0.1.0"
+edition = "2021"
+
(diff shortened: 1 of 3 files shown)


bnusunny:
Could you add a simple README file?

timClicks:
Is it possible to add a README to the example to make it obvious that arguments need to be passed when calling it?

> // This example requires the following input to succeed:
> // { "bucket": "bucket-to-list" }

(I know that the other examples don't have thorough READMEs, but it would be worthwhile practice to start)

bnusunny (approved):

PR #620, opened 2023-03-18
disable connection pool timeout

richarddd:
*Issue #, if available:*

*Description of changes:*
Disable the default 90s connection pool timeout to increase connection reuse


bnusunny:
I'm not sure if Lambda Runtime API has a keep-live timeout for client connections. Do you test this will long idle time?

bnusunny:
Will this work with SnapStart?

richarddd:
I think this might be a premature optimization since the request is "blocked" until a new event arrives. So we'll probably never hit the connect timeout unless it takes 90 seconds to process the event which makes the localhost connection latency seem neglishable

Issue #613, opened 2023-03-06
s3 examples


peterborkuti:
Dear Developers,

I would like to add some S3 examples so I opened this ticket.
If you do not think it is useful, just tell me.

I already have two very basic:
1) listen to an S3 event
2) listen to only S3 Create event
I would like to add three more:
1) download an image to local from s3 bucket when the image is uploaded
2) create a thumbnail from it
3) upload a thumbnail to the same bucket with a different name
If you already have some other ideas, just tell me, I am learning AWS and rust, so I am willing to work on other ideas.

Thank you in advance
Peter

greenwoodcm:
Some S3 examples sound great!  Feel free to open a PR when you have something mergeable.  In doing so, I wonder if one full-featured S3 example would suffice.  Let's say that example istens to the event, filters for Create events, downloads, creates thumbnail, and uploads into a different bucket/key.  That's one example function that the user could easily strip down and simplify if they only need a subset of that stuff.  I think that's clearer than having 5 different examples.  What do you think?

peterborkuti:
One bigger example: makes sense. I am doing that.

greenwoodcm:
cool, open a PR when you've got something

PR #621, opened 2023-03-19
s3 example - thumbnail creator (#613)

peterborkuti:
*Issue #613, if available:*

*Description of changes:*
S3 example: donwload + create thumbnail + upload
It uses a shared client, based on the example #619


examples/basic-s3-thumbnail/src/s3client.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn get_thumbnail(&self, reader: Cursor<Vec<u8>>) -> ByteStream {
+        let mut thumbnails = create_thumbnails(reader, mime::IMAGE_PNG, [ThumbnailSize::Small]).unwrap();
+
+        let thumbnail = thumbnails.pop().unwrap();
+        let mut buf = Cursor::new(Vec::new());
+        thumbnail.write_png(&mut buf).unwrap();
+
+        return ByteStream::from(buf.into_inner());
+    }
+}
```
greenwoodcm, line 31: it doesn't seem like the thumbnail logic belongs in this module, since it simply translates `Cursor<Vec<u8>> -> ByteStream`.  maybe this is best placed in the top level module.

examples/basic-s3-thumbnail/src/s3client.rs
```diff
@@ -0,0 +1,78 @@
...
+
+#[async_trait]
+impl GetFile for S3Client {
+    async fn get_file(&self, bucket: &str, key: &str) -> Option<Cursor<Vec<u8>>> {
+        println!("get file bucket {}, key {}", bucket, key);
```
greenwoodcm, line 36: np: what do you think about making these `tracing::info` calls?

examples/basic-s3-thumbnail/src/s3client.rs
```diff
@@ -0,0 +1,78 @@
...
+
+        let output = self.get_object().bucket(bucket).key(key).send().await;
+
+        let mut reader = None;
+
+        if output.as_ref().ok().is_some() {
```
greenwoodcm, line 42: this is more clearly written as a match statement
```
match output {
    Ok(response) => {
        ...do stuff
    }
    Err(err) => {
        ...do other stuff
    }
}

File: examples/basic-s3-thumbnail/src/s3client.rs [hunk 4]
```diff
@@ -0,0 +1,78 @@
+use std::io::Cursor;
+
+use async_trait::async_trait;
+use aws_sdk_s3::{types::ByteStream, Client as S3Client};
+use thumbnailer::{create_thumbnails, ThumbnailSize};
+
+#[async_trait]
+pub trait GetFile {
+    async fn get_file(&self, bucket: &str, key: &str) -> Option<Cursor<Vec<u8>>>;
```
greenwoodcm, line 9: I'd recommend this trait also return a `Result` and let the upper layer handle the error as it sees fit

examples/basic-s3-thumbnail/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+        return empty_response;
+    }
+
+    println!("Bucket: {}, Object key: {}", bucket_name, object_key);
+
+    return (bucket_name, object_key);
+}
+
+async fn get_client() -> S3Client {
+    let region_provider = RegionProviderChain::default_provider().or_else("us-east-2");
```
greenwoodcm, line 84: under what circumstances will the `default_provider` not return a provider?  i think the `from_env` in the next line down will figure out the region by itself.

examples/basic-s3-thumbnail/src/main.rs
```diff
@@ -0,0 +1,227 @@
...
+    #[tokio::test]
+    async fn response_is_good() {
+        let mut context = Context::default();
+        context.request_id = "test-request-id".to_string();
+
+        let bucket = "test-bucket";
+        let key = "test-key";
+
+        mock! {
+            FakeS3Client {}
```
greenwoodcm, line 146: i think you'll be able to automock the traits.  i'd recommend the following so that the mock trait impls are only generated in test (i think this is the syntax):
```
#[cfg_attr(test, automock)]
```

peterborkuti:
Thank you for the thorough review. Give me some days to fix your findings.

examples/basic-s3-thumbnail/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+# to install the `add` subcommand.
+#
+# Running `cargo add DEPENDENCY_NAME` will
+# add the latest version of a dependency to the list,
+# and it will keep the alphabetic ordering for you.
+
+[dependencies]
+aws_lambda_events = "0.7.2"
+lambda_runtime = { path = "../../lambda-runtime" }
+serde = "1.0.136"
```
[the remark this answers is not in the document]
timClicks: I would go further and go with `serde = "1"`

examples/basic-s3-thumbnail/src/main.rs
peterborkuti: I have read here: https://docs.rs/mockall/latest/mockall/
That automock can mock traits only one impl and I have two.

peterborkuti: I fixed the others (hopefully), now I am testing and will send the PR.

peterborkuti:
Dear @greenwoodcm and @timClicks,
I improved the PR based on your remarks.

examples/basic-s3-thumbnail/src/s3client.rs
peterborkuti: I think, I "copy pasted" the code from here:
https://github.com/awslabs/aws-sdk-rust/blob/main/examples/s3/src/bin/client.rs
So unfortunately, I do not understand it well. I just trusted in them, shame on me.

examples/basic-s3-thumbnail/src/main.rs
peterborkuti: I think, I "copy pasted" the code from here:
https://github.com/awslabs/aws-sdk-rust/blob/main/examples/s3/src/bin/client.rs
So unfortunately, I do not understand it well. I just trusted in them, shame on me.

peterborkuti:
I checked the performance, so I uploaded the same image with different names:

`for i in `seq 1 10`; do aws s3 cp image.png s3://bpimage1/image-run-slow2-$i.png; done`

It created different events, so every upload initiated one handler run.
The duration was between **40ms and 100ms**.

I moved the client instantiation into the handler and checked the same way, the duration was between **122ms and 185ms**.
So instantiating client in the main is a must have.


lambda-fast


lambda-slow

examples/basic-s3-thumbnail/src/main.rs
```diff
@@ -0,0 +1,240 @@
...
+        // So it needs file creation permission into that bucket
+
+        let _ = client.put_file(&thumbs_bucket, &key, thumbnail).await;
+    }
+
+    Ok(())
+}
+
+fn get_file_props(record: &S3EventRecord) -> (String, String) {
+    let empty_response = ("".to_string(), "".to_string());
```
greenwoodcm, line 61: instead of using the empty string to represent nothing here, i recommend having this function return `(Option<String>, Option<String>)`

peterborkuti:
Hi @greenwoodcm,
I modified the code based on your second review, thank you.
I think, the code now is more "rusty" than before.
Péter

greenwoodcm (approved):
looks great, thanks for the contribution!

PR #640, opened 2023-04-14
Provide context when deserialization fails.

### LINKED ISSUE #639: Explain Serde deserialization issues clearly in error messages
When a function cannot deserialize the incoming data into the type that the handler receives. The runtime returns a Serde deserialization error. We've seen many reports showing that those messages are confusing because they are very sparse. I believe improving the those error messages would remove a lot of the friction that people bump into when using the runtime.

These are some examples of those messages:

- `"field_error": "expected value at line 1 column 1",`
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/599
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/559
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/528

calavera:
Use [serde_path_to_error ](https://docs.rs/serde_path_to_error/latest/serde_path_to_error/index.html) to provide a little bit more information about the serde failure.

*Issue #, if available:*

Fixes #639 

*Description of changes:*

### DIFF
--- a/lambda-runtime/src/deserializer.rs
+++ b/lambda-runtime/src/deserializer.rs
@@ -4,7 +4,7 @@ use serde::Deserialize;
 
 use crate::{Context, LambdaEvent};
 
-const ERROR_CONTEXT: &str = "failed to deserialize the incoming data into the function payload's type";
+const ERROR_CONTEXT: &str = "failed to deserialize the incoming data into the function's payload type";
 
 /// Event payload deserialization error.
 /// Returned when the data sent to the function cannot be deserialized


lambda-runtime/src/deserializer.rs
```diff
@@ -0,0 +1,43 @@
+use std::{error::Error, fmt};
+
+use serde::Deserialize;
+
+use crate::{Context, LambdaEvent};
+
+const ERROR_CONTEXT: &str = "failed to deserialize the incoming data into the function payload's type";
```
greenwoodcm, line 7: np: function's payload type

greenwoodcm (approved):

PR #653, opened 2023-05-12
Fix doc about curl command

### LINKED ISSUE #651: Curl example in the read me doc does not work without a `/` at the end
Hi,
While running cargo lambda locally with the cli, when the run time is online using the following code 

```
curl -X POST http://localhost:9001/lambda-url/test-lambda-http -H "Content-Type: text/html" -d '{"username": "abc", "age": "100"}'
```
Does not work but by adding a `/` after the test-lambda-http it works. As follows:

```
curl -X POST http://localhost:9001/lambda-url/test-lambda-http/ -H "Content-Type: text/html" -d '{"username": "abc", "age": "100"}'
```

I spent quite a lot of time to find the issue because it does not produce any log that what is the problem. I will better to have the readme file updated of the project with the suggested changes if that is an expected scenario.

calavera:
*Issue #, if available:*

Fixes #651 

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -358,7 +358,7 @@ An simpler alternative is to cURL the following endpoint based on the address an
 
 ```bash
 curl -v -X POST \
-  'http://127.0.0.1:9001/lambda-url/<lambda-function-name>' \
+  'http://127.0.0.1:9001/lambda-url/<lambda-function-name>/' \
   -H 'content-type: application/json' \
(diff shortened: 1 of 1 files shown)


DataPsycho (approved):

DataPsycho:
Thanks for quick update. I though I will have to do a pr but you already did.

bnusunny (approved):

PR #676, opened 2023-07-14
Serialize APIGW queryStringParameters properly

mrautio:
**This has a dependency on a PR of query-map-rs:** https://github.com/calavera/query-map-rs/pull/11

API Gateway Lambda proxy payload's `queryStringParameters` is serialized incorrectly to format like `"queryStringParameters":{"key":["value"]}` when it should be like `"queryStringParameters":{"key":"value"}`
Reference: https://docs.aws.amazon.com/apigateway/latest/developerguide/http-api-develop-integrations-lambda.html 

*Description of changes:*

- Corrected serialization of `queryStringParameters` for API Gateway Lambda proxy payload format version 1.0 and 2.0

### DIFF
--- a/lambda-events/Cargo.toml
+++ b/lambda-events/Cargo.toml
@@ -30,7 +30,7 @@ chrono = { version = "0.4.23", default-features = false, features = [
   "serde",
   "std",
 ], optional = true }
-query_map = { version = "^0.6", features = ["serde", "url-query"], optional = true }
+query_map = { version = "^0.7", features = ["serde", "url-query"], optional = true }
 flate2 = { version = "1.0.24", optional = true }
 
 [features]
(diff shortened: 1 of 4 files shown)


mrautio:
> What's the V1 format when there are multiple parameters with the same name?

Whoops, I misread the AWS doc. Now V1 value is correct. In V1 **first** value will be used if there are multiple values.

Example:
Source data: `"key": ["value1", "value2"]`
V1:  `"key": "value1"`
V2: `"key": "value1,value2"`

mrautio:
@calavera I've split the change to two PRs, query-map-rs PR here: https://github.com/calavera/query-map-rs/pull/11

mrautio:
thank you for the quick release! I've now bumped the query_map version to 0.7 in this PR

PR #692, opened 2023-09-04
Fix streaming prelude serialization

calavera:
*Description of changes:*

HTTP headers can be multi-value, but the current implementation ignores this fact and only serializes the first value for each header.

This changes uses http-serde to serialize the prelude correctly.


bnusunny:
Thanks for spotting this. Lambda Function URLs doesn't support multiple value HTTP headers. The original version was designed with that in mind. But for direct API calls or SDK invokes, this limit should not apply.

bnusunny (approved):

Issue #691, opened 2023-09-04
Add the possibility to flush traces/metrics after a handler loop has finished


purkhusid:
The problem I currently have is that there is no good way to flush traces/logs/metrics/errors once the runtime has finished handling a message. 

Currently the handling code is structured something like this:

```
let request_span = <create span for the request>

async {
    <do parsing, call the user provider handler, catch panics etc>
}.instrument(tracing_span)
```

There are a couple of issues with this approach:
1. The user provided handler is the only way or the end user to affect the loop and thus he has to e.g. flush spans inside the handler
2. Since the handler can panic it is possible that the flushing is never done since it has to happen after the work has been done.
3. The `request_span` span can not be flushed since it's finished after the user provided handler is provided.

I think it would be very valuable that the runtime would provide a hook for the end user so that they can do cleanup after each invocation.

purkhusid:
I patched the runtime with this to fix tracing in our lambdas:
```
diff --git lambda-runtime/src/lib.rs lambda-runtime/src/lib.rs
index e3ffd49..a01d2cb 100644
--- lambda-runtime/src/lib.rs
+++ lambda-runtime/src/lib.rs
@@ -20,7 +20,7 @@ use std::{
     env,
     fmt::{self, Debug, Display},
     future::Future,
-    panic,
+    panic, sync::Arc,
 };
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_stream::{Stream, StreamExt};
@@ -101,6 +101,7 @@ where
         &self,
         incoming: impl Stream<Item = Result<http::Response<hyper::Body>, Error>> + Send,
         mut handler: F,
+        callback: Option<Arc<impl Fn()>>
     ) -> Result<(), Error>
     where
         F: Service<LambdaEvent<A>>,
@@ -202,7 +203,13 @@ where
             }
             .instrument(request_span)
             .await?;
+        
+            if let Some(callback) = callback.clone() {
+                callback();
+            }
         }
+
+
         Ok(())
     }
 }
@@ -258,7 +265,52 @@ where
 
     let client = &runtime.client;
     let incoming = incoming(client);
-    runtime.run(incoming, handler).await
+    let callback : Option<Arc<fn()>>= None;
+    runtime.run(incoming, handler, callback).await
+}
+
+/// Starts the Lambda Rust runtime and begins polling for events on the [Lambda
[...]

PR #694, opened 2023-09-05
Allow running handler with a callback

purkhusid:
## What?
This is a pretty naive approach that allows the end user to supply a callback that will be invoked once a single event loop has been finished.

## Why?
Because currently there is no good way(that I know of) to flush logs/traces/metrics. By having a callback that is called after all event handling is finished allows the end user to do a cleanup at the end of each lambda invocation.

See #691 for further context

### DIFF
--- a/examples/basic-lambda-with-callback/Cargo.toml
+++ b/examples/basic-lambda-with-callback/Cargo.toml
@@ -1,0 +1,19 @@
+[package]
+name = "basic-lambda-with-callback"
+version = "0.1.0"
+edition = "2021"
+
+
+# Use cargo-edit(https://github.com/killercup/cargo-edit#installation)
+# to manage dependencies.
+# Running `cargo add DEPENDENCY_NAME` will
+# add the latest version of a dependency to the list,
+# and it will keep the alphabetic ordering for you.
+
+[dependencies]
+lambda_runtime = { path = "../../lambda-runtime" }
+serde = "1.0.136"
+tokio = { version = "1", features = ["macros"] }
+tracing = { version = "0.1", features = ["log"] }
+tracing-subscriber = { version = "0.3", default-features = false, features = ["fmt"] }
+tokio-test = "0.4.2"
--- a/examples/basic-lambda-with-callback/README.md
+++ b/examples/basic-lambda-with-callback/README.md
@@ -1,0 +1,11 @@
+# AWS Lambda Function example
+
+## Build & Deploy
+
+1. Install [cargo-lambda](https://github.com/cargo-lambda/cargo-lambda#installation)
+2. Build the function with `cargo lambda build --release`
+3. Deploy the function to AWS Lambda with `cargo lambda deploy --iam-role YOUR_ROLE`
+
+## Build for ARM 64
+
+Build the function with `cargo lambda build --release --arm64`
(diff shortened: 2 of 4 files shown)


greenwoodcm:
Thanks for the submission!  In general I like this change, I think this is functionality we want to support.  The current approach feels a bit disjointed - the callback is passed as an `Arc<impl Fn()>` and exists outside of the "awareness" of the `Service` impl the developer provides.  I'd like to find a way to make the callback more of a first-class construct down at that level of the type system.  the goal here for me is allowing the developer to reason over the function handler and the cleanup callback closer together in code.  A few alternatives I can think of:
1. right now the `Service` impl must return back a response of type `B: Serialize`, which is wired up as `Service::Future: Future<Output = Result<B, F::Error>>`.  we could make the response type optionally be a wrapper type that includes both the response payload (still of type `Serialize`) and additionally includes a callback `Fn()`.  in that model the function handler is responsible for returning the callback on every invoke response, the contract being "here's the response, and here's the way to call me back to clean up after you've responded".
2. there's the concept of [middleware and layering](https://docs.rs/tower/latest/tower/trait.Service.html#middleware--layer) in tower.  this is the idea that you can layer one service on top of another.  the example they give in the docs is to build a timeout service that wraps another service and enforces a timeout.  we could make the callback infrastructure essentially middleware around the existing `Service` impl.  then what I would do is construct a `Service` using `service_fn`, then wrap it in a `callback_fn` that returns it's wrapped `Service`.  something like that.

thoughts?  my initial impression is that i like option two better.

BMorinDrifter:
See `MetricsService` in  https://github.com/BMorinDrifter/metrics-cloudwatch-embedded/blob/main/src/lambda.rs for how I did this for the embedded metrics crate.

That said, working with Tower ended up being quite a chore between figuring out how to write `MetricsServiceFuture` and dealing with metaprogramming because the tower service is coupled to response that's either a thing for serde to reserialize of an http response depending on if you're using lambda_http or not.

greenwoodcm:
Yeah I started prototyping what I had in mind, and one of the annoying things is that you can't build Tower middleware with standard async/await semantics.  so we'd probably end up doing our own `Future` stuff like you did, @BMorinDrifter .  i don't think we have the response type problem you refer to, as I _think_ where I'd take this is the `Service` that `run` uses has a response type of `()`.  that service internally would be responsible not just for invoking the handler, but also for posting the response back to the runtime API.  it would essentially terminate the response type itself because the response isn't needed after it's serialized and sent.  this would simplify the type signature and the body of `Runtime::run` and would leave the door open for something to wrap that new `Service` in a middleware that does callbacks.

BMorinDrifter:
> Yeah I started prototyping what I had in mind, and one of the annoying things is that you can't build Tower middleware with standard async/await semantics. so we'd probably end up doing our own `Future` stuff like you did, @BMorinDrifter .

My bad, I had that backwards.  Yeah the hook after the lambda function is what required writing my own Future.  So yeah, doing this right now does require quite a bit of code.  There might be a an easier way with [tower](https://docs.rs/tower/latest/tower/index.html)::[ServiceBuilder](https://docs.rs/tower/latest/tower/struct.ServiceBuilder.html#) however when I was unable to get that to work at the library scope.

Issue #680, opened 2023-07-24
CFN Response


brianheineman-aver:
I have developed a CloudFormation custom resources and did not see any existing support in the crate(s) to send a response to CFN.   For the time being, I created a solution using the `reqwest` crate for my immediate needs as I did not see an existing solution.

[Feature Request]
Please consider adding support for sending CFN Responses similar to the Python [cfnresponse](https://pypi.org/project/cfnresponse/) library.

PR #695, opened 2023-09-10
Add event definitions for CloudFormation custom resources

chris-leach:
*Issue #, if available:*
Related to #680 - see below

*Description of changes:*
This PR adds event definitions for CloudFormation [custom resource request objects](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requests.html). Requests are either a [Create](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requesttypes-create.html), [Update](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requesttypes-update.html), or [Delete](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/crpg-ref-requesttypes-update.html) event, implemented as enum variants with internally tagged representation.

Note that, in addition to the fields documented in the links above, in practice CloudFormation sends a `ServiceToken` field in every request. Since this isn't documented I've encoded it as an `Option<String>` with `#[serde(default)]`, so deserialisation will tolerate it being missing, but I can change this if desired.

---

This is intended as an initial building block of a framework for implementing Lambda custom resource handlers, of the kind requested in #680. These work by invoking a Lambda with one of the above event types, all of which include an S3 presigned URL to which a response must be posted. Failing to post a response can result in long delays, so tools to make it easy to send a response, and a fallback mechanism to send a failure response if the Lambda is about to time out, are desirable.

I'd appreciate feedback on whether further PRs to add this functionality (probably in a `lambda_cfn` crate) would be appreciated, but either way I think the event definitions are a useful addition.

### DIFF
--- a/lambda-events/Cargo.toml
+++ b/lambda-events/Cargo.toml
@@ -42,6 +42,7 @@ default = [
   "autoscaling",
   "chime_bot",
   "clientvpn",
+  "cloudformation",
   "cloudwatch_events",
   "cloudwatch_logs",
   "code_commit",
@@ -81,6 +82,7 @@ appsync = []
 autoscaling = ["chrono"]
 chime_bot = ["chrono"]
 clientvpn = []
+cloudformation = []
 cloudwatch_events = ["chrono"]
 cloudwatch_logs = ["flate2"]
 code_commit = ["chrono"]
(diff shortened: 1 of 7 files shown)


chris-leach:
@calavera Done - I've removed the `let ... else` statements that I now realise were stabilised one release after the MSRV.

chris-leach:
@calavera pushed a fix for this - please review

PR #704, opened 2023-10-12
Eventbridge Event Processor

nichmorgan:
*Issue #, if available:*

*Description of changes:*
New event created to support Eventbridge.

### DIFF
--- a/lambda-events/Cargo.toml
+++ b/lambda-events/Cargo.toml
@@ -30,7 +30,10 @@ chrono = { version = "0.4.23", default-features = false, features = [
   "serde",
   "std",
 ], optional = true }
-query_map = { version = "^0.7", features = ["serde", "url-query"], optional = true }
+query_map = { version = "^0.7", features = [
+  "serde",
+  "url-query",
+], optional = true }
 flate2 = { version = "1.0.24", optional = true }
 
 [features]
@@ -72,6 +75,7 @@ default = [
   "sns",
   "sqs",
   "streams",
+  "eventbridge",
 ]
 
 activemq = []
@@ -112,3 +116,4 @@ ses = ["chrono"]
 sns = ["chrono", "serde_with"]
 sqs = ["serde_with"]
 streams = []
+eventbridge = ["chrono", "serde_with"]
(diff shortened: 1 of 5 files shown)


lambda-events/Cargo.toml
```diff
@@ -30,7 +30,10 @@ chrono = { version = "0.4.23", default-features = false, features = [
   "serde",
   "std",
 ], optional = true }
-query_map = { version = "^0.7", features = ["serde", "url-query"], optional = true }
+query_map = { version = "^0.7", features = [
+  "serde",
+  "url-query",
+], optional = true }
```
BMorinDrifter, line 36: nit: whitespace only change

lambda-events/src/event/eventbridge/mod.rs
```diff
@@ -0,0 +1,87 @@
+use chrono::{DateTime, Utc};
+use serde::de::DeserializeOwned;
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct EventBridgeEvent {
```
BMorinDrifter, line 8: Could this be an alias for `EventBridgeEventObj<Option<String>>` to avoid code duplication?

BMorinDrifter (commented):
Some minor polish feedback, otherwise looks good to me.

lambda-events/src/event/eventbridge/mod.rs
nichmorgan: We are having some problems creating this Alias due to the use of this setting in the detail field:

```rust
#[serde_as(as = "serde_with::json::JsonString")]
#[serde(bound(deserialize = "T: DeserializeOwned"))]
pub detail: T,
```

As the input of this field will always be a String, we expect it to be a JsonString as an EventBridgeEventObj, but if you create an alias EventBridgeEventObj<Option<String>> and its contents are a simple String, not a JsonString, the test will fail.

We're still working on solutions to avoid duplication, but so far we've had no success. Any suggestions?

Issue #712, opened 2023-10-26
lambda_http::RequestExt is missing with_stage_variables method


bassmanitram:
The with_* methods are great for mocking - is it an oversight to not have one for Stage Variables?

greenwoodcm:
i don't know of any reason this is missing.  if you want to submit a PR for adding i'd be happy to look at it.

bassmanitram:
Well, it's already there - but masked by `#cfg[(test)]` with an explicit comment that its available under `#[cfg(test)]` configurations - so for its intended use it IS available. Obviously, as such, it's missing in the docs. Is there any objection to _removing_ the test-only constraint (the other with_ methods don't have that)?

PR #713, opened 2023-10-27
Remove cfg(test) on with_stage_variables

bassmanitram:
*Issue #, if available:*
#712 

*Description of changes:*
Tis PRt removes the cf(test) constraint from the `with_stage_variables` method. constraint.

None of the other with_ methods are constrained to test-only, although this method did have a specific comment indicating test-only, so there may have been a reason.

### DIFF
--- a/lambda-http/src/ext/extensions.rs
+++ b/lambda-http/src/ext/extensions.rs
@@ -108,10 +108,9 @@ pub trait RequestExt {
     /// These will always be `None` for ALB triggered requests.
     fn stage_variables_ref(&self) -> Option<&QueryMap>;
 
-    /// Configures instance with stage variables under `#[cfg(test)]` configurations
+    /// Configures instance with stage variables
     ///
     /// This is intended for use in mock testing contexts.
-    #[cfg(test)]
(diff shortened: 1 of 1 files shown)


lambda-http/src/ext/extensions.rs
```diff
@@ -108,10 +108,9 @@ pub trait RequestExt {
     /// These will always be `None` for ALB triggered requests.
     fn stage_variables_ref(&self) -> Option<&QueryMap>;
 
-    /// Configures instance with stage variables under `#[cfg(test)]` configurations
+    /// Configures instance with stage variables
     ///
     /// This is intended for use in mock testing contexts.
```
greenwoodcm, line 113: is this part of the comment still accurate?

greenwoodcm (approved):

lambda-http/src/ext/extensions.rs
bassmanitram: I removed the added comments "under ... " because I removed the config annotation - the comment is now in the same form as the other `with_` methods.

PR #733, opened 2023-11-21
Turn request_id into a reference.

calavera:
*Description of changes:*

This change prevents us from allocating an extra string in every invocation.


greenwoodcm (approved):
do you think it's worth the (minor) complexity to avoid the (minor) allocation overhead?

Issue #737, opened 2023-11-25
Upgrade to hyper/http 1.0


jonhoo:
Now that version 1.0 of the hyper and http* crates are out, the lambda* crates should probably follow suit. Unfortunately, it's a decently heavy lift, especially due to `Body` changing from a concrete type to a trait (https://github.com/hyperium/hyper/issues/2345). I took a stab at it, but quickly got lost in a sea of `<B: Body>`s, and figured I'd leave it to those who know the library best.

A few bits I did pick up that may come in handy for someone wanting to take this on:

- `body::to_bytes(body).await?` becomes `body.collect().await?.to_bytes()` via [`http_body_util::BodyExt`](https://docs.rs/http-body-util/latest/http_body_util/trait.BodyExt.html)
- The old `Body` becomes `BoxBody<hyper::body::Bytes, hyper::Error>` via [`http_body_util::combinators::BoxBody`](https://docs.rs/http-body-util/latest/http_body_util/combinators/struct.BoxBody.html) _if_ you want to keep it fully type-erased like it is today. Otherwise, `hyper::body::Incoming` is the right replacement type for at least most `Response`s.
- There is no more pooling http client, nor a `Connector` trait (or rather, they're relegated to [`hyper_util::client::legacy`](https://docs.rs/hyper-util/latest/hyper_util/client/legacy/index.html)). Instead, there is now [`hyper::client::conn::http1::Connection`](https://docs.rs/hyper/latest/hyper/client/conn/http1/struct.Connection.html).

davidbarsky:
> What are the benefits of migrating to the new version?

Saying this for any readers—`http` is a public dependency of the Lambda runtime, which will manifest in (indirectly related) errors like https://github.com/awslabs/aws-lambda-rust-runtime/issues/738.

simbleau:
> What are the benefits of migrating to the new version?
> 
> I tried to do it too, and I found the lack of documentation about all the deprecated things troubling. So I'm reticent to spend quite a lot time without knowing that's worth the work.
> 
> If anyone wants to do the heavy lifting, we appreciate the contributions.

A benefit would be to play nicely with other libraries, e.g. Axum, who have followed suit and updated `http` to 1.0. Hence, anyone making a Lambda service using Axum might run into this:

<img width="1408" alt="image" src="https://github.com/awslabs/aws-lambda-rust-runtime/assets/48108917/be15e4fe-1725-4967-98eb-ca34901fa13b">

[...]

PR #740, opened 2023-11-29
Draft: update hyper/http

simbleau:
*Issue #, if available:*
#737 

*Description of changes:*
Updates a good bit of #737 -
Hit a snag, though. I don't fully understand `hyper::Client` and how it worked, likewise for `Connector`s, and it seems they were removed with the update to 1.0.

This means attention is needed in `lambda-runtime-api-client/lib.rs`.
Perhaps @calavera is able to update that file?

### DIFF
--- a/lambda-extension/Cargo.toml
+++ b/lambda-extension/Cargo.toml
@@ -17,13 +17,23 @@ readme = "README.md"
 async-stream = "0.3"
 bytes = "1.0"
 chrono = { version = "0.4", features = ["serde"] }
-http = "0.2"
-hyper = { version = "0.14.20", features = ["http1", "client", "server", "stream", "runtime"] }
+http = "1.0"
+hyper = { version = "0.14.20", features = [
+    "http1",
+    "client",
+    "server",
+    "stream",
+    "runtime",
+] }
 lambda_runtime_api_client = { version = "0.8", path = "../lambda-runtime-api-client" }
 serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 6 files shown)


simbleau:
I've put up the bat symbol, an issue to hyper, to help speed things up.

lambda-extension/Cargo.toml
```diff
@@ -17,13 +17,23 @@ readme = "README.md"
...
-http = "0.2"
-hyper = { version = "0.14.20", features = ["http1", "client", "server", "stream", "runtime"] }
+http = "1.0"
+hyper = { version = "0.14.20", features = [
+    "http1",
+    "client",
+    "server",
+    "stream",
+    "runtime",
+] }
```
Oliboy50, line 27: this should also be upgraded to `hyper v1`, no?

(for the "server" part, `hyper-util` may be required though)

simbleau: Fixed

simbleau:
I was able to go a bit further after some clarity in https://hyper.rs/guides/1/upgrading/ and https://github.com/hyperium/hyper/issues/3450.

Now `lambda-runtime-api-client` has no compile issues, and should generally serve as a guide on how to upgrade the rest.

@calavera It would be better if you took a look at how I updated the `lambda-runtime-api-client`, made sure there were no issues with the upgrades, and then tried to take a stab at `lambda-runtime` and `lambda-extension` which still have compile issues.

PR #744, opened 2023-12-02
Support internal Lambda extensions

### LINKED ISSUE #743: Support for internal Lambda extensions
First off, thanks for these useful crates! It's nice to see official Rust support from AWS.

The `lambda-extension` crate currently only supports **external** [Lambda extensions](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html) (i.e., those running in a separate binary from the main runtime). AWS Lambda also supports **internal** Lambda extensions that run in the same process as the runtime, and it would be great if this crate also supported those.

The main requirement for supporting internal extensions is that each internal extension must register itself with the Lambda Runtime API [*before* the runtime thread calls `Next`](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-extensions-api.html#runtimes-extensions-api-reg):
> The `Init` phase completes after the runtime and each registered extension indicate completion by sending a `Next` API request.

Once the main runtime calls `Next`, it's too late to register any additional extensions. The Lambda Runtime API returns an error if an extension tries to register itself at that point.

The current implementation of `Extension::run()` [registers the extension](https://github.com/awslabs/aws-lambda-rust-runtime/blob/3c53476206482535f6b73f118e09258aeeb1d6f8/lambda-extension/src/extension.rs#L222) with the Lambda Runtime API and then proceeds to [call `Next`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/3c53476206482535f6b73f118e09258aeeb1d6f8/lambda-extension/src/extension.rs#L314). Since `Extension::run()` returns a single future that includes both registration and invocation of the extension, there's no way for the caller to determine when registration is complete, and that it's safe to call `Runtime::run()` (which immediately [calls `Next`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/3c53476206482535f6b73f118e09258aeeb1d6f8/lambda-runtime/src/lib.rs#L121) and ends the `Init` phase). In other words, calling `Extension::run()` and `Runtime::run()` concurrently creates a race condition between registering the extensions and the runtime calling `Next`.

A straightforward solution would be to replace (or augment) the `Extension::run()` function with an API that looks like the following (note the new `Extension::register()` function and a new type representing a *registered* extension that's ready to be run):
```rust
let extension = Extension::new()
  .with_events(&["INVOKE"])
  .with_events_processor(...);
let registered_extension = extension.register().await;

futures::future::try_join(
  registered_extension.run(),
  Runtime::run(...),
).await
```

A potential downside to this API is that `.with_events` will allow a user to attempt to register the `SHUTDOWN` event, which will fail at runtime since it's not permitted for internal extensions. Also, I'm not sure whether the telemetry and logs processing functionality works for internal extensions or not. These issues could be solved by having separate types for internal and external extensions, along with an enum to represent the supported events instead of accepting raw strings. However, that's a more involved API redesign that may not be worth it at this point.

I'm happy to contribute a PR once there's consensus on the desired API!

ramosbugs:
Internal Lambda extensions must be registered during the Lamba lifecycle Init phase, which ends when the runtime calls Next to await the first event. Since the `Extension::run` function both registers and executes the extension, there was previously no way for the runtime to determine that all internal extensions have been registered and that it's safe to proceed to the Invoke lifecycle phase.

This change introduces an `Extension::register` method that registers the extension and begins any logs/telemetry handlers. It then returns a new `RegisteredExtension` abstraction that can be used to invoke the extension's run loop concurrently with the runtime's run loop.

This change maintains backward compatibility by having the existing `Extension::run` method perform both steps. External Lambda extensions can use either API, and internal extensions should use the new API.

Resolves #743.

- [X ] I confirm that my contribution is made under the terms of the Apache 2.0 license.
- [X ] I confirm that I've made a best effort attempt to update all relevant documentation.

### DIFF
--- a/lambda-extension/src/extension.rs
+++ b/lambda-extension/src/extension.rs
@@ -215,14 +215,21 @@ where
         }
     }
 
-    /// Execute the given extension
-    pub async fn run(self) -> Result<(), Error> {
+    /// Register the extension.
+    ///
+    /// Performs the
+    /// [init phase](https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtime-environment.html#runtimes-lifecycle-ib)
+    /// Lambda lifecycle operations to register the extension. When implementing an internal Lambda
+    /// extension, it is safe to call `lambda_runtime::run` once the future returned by this
+    /// function resolves.
+    pub async fn register(self) -> Result<RegisteredExtension<E>, Error> {
         let client = &Client::builder().build()?;
 
         let extension_id = register(client, self.extension_name, self.events).await?;
         let extension_id = extension_id.to_str()?;
-        let mut ep = self.events_processor;
 
+        // Logs API subscriptions must be requested during the Lambda init phase (see
+        // https://docs.aws.amazon.com/lambda/latest/dg/runtimes-logs-api.html#runtimes-logs-api-subscribing).
         if let Some(mut log_processor) = self.logs_processor {
             trace!("Log processor found");
 
@@ -262,6 +269,8 @@ where
             trace!("Registered extension with Logs API");
         }
 
+        // Telemetry API subscriptions must be requested during the Lambda init phase (see
+        // https://docs.aws.amazon.com/lambda/latest/dg/telemetry-api.html#telemetry-api-registration
         if let Some(mut telemetry_processor) = self.telemetry_processor {
             trace!("Telemetry processor found");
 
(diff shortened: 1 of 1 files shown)


lambda-extension/src/extension.rs
```diff
@@ -301,6 +310,42 @@ where
...
+    }
+
+    /// Execute the given extension.
+    pub async fn run(self) -> Result<(), Error> {
+        self.register().await?.run().await
+    }
+}
+
+/// An extension registered by calling [`Extension::register`].
+pub struct RegisteredExtension<E> {
```
ramosbugs, line 326: This new type seemed unavoidable since `Extension::register` has to pass ownership of all the logging and telemetry fields to the spawned background tasks. These are the remaining fields to be used in the invoke phase.

ramosbugs:
The `cargo lambda` runtime emulator doesn't seem to support internal extensions, but I tested the example by deploying it to AWS via:

```bash
cargo lambda deploy
cargo lambda invoke extension-internal-flush --data-example sqs-event-obj --verbose --remote
```

This produced the following CloudWatch logs showing both the runtime and extension behaving as expected:
```
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
|   timestamp   |                                                                                 message                                                                                 |
|---------------|-------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| 1701561431609 | INIT_START Runtime Version: provided:al2023.v9 Runtime Version ARN: arn:aws:lambda:us-west-2::runtime:68313de216b90a01e19138b97177c40478ef50b63a22b29f152b960070428ae8  |
| 1701561431636 | EXTENSION Name: bootstrap State: Ready Events: [INVOKE]                                                                                                                 |
| 1701561431637 | START RequestId: 34bcc601-2f2d-4ad8-af6e-124968936a5a Version: $LATEST                                                                                                  |
| 1701561431637 | [runtime] received event Data { a: "Test", b: 123 }                                                                                                                     |
| 1701561431637 | [extension] waiting for event to be processed                                                                                                                           |
| 1701561431637 | [extension] flushing logs and telemetry                                                                                                                                 |
| 1701561431648 | END RequestId: 34bcc601-2f2d-4ad8-af6e-124968936a5a                                                                                                                     |
| 1701561431648 | REPORT RequestId: 34bcc601-2f2d-4ad8-af6e-124968936a5a Duration: 11.06 ms Billed Duration: 38 ms Memory Size: 128 MB Max Memory Used: 14 MB Init Duration: 26.63 ms     |
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

PR #748, opened 2023-12-06
change some properties to optional value

KUrushi:
*Issue #, if available:*

*Description of changes:*
When I actually created an Event and started it from a Lambda function, I noticed that some parameters were Nullable.

Agents for Amazon Bedrock takes into account the HTTP Method of the configured schema and sets values for the properties.
So I noticed that, for example, if the GET method is configured in the schema, there is usually no Request body.

I don't know if this is correct because the schema page of the document does not specify whether it is Nullable or not, but I think it is safest to set parameters and request_body to Optional.

If anyone knows the details, I would appreciate it if you could let me know.


KUrushi:
Thank you for your code suggestion! I have applied your suggestion.

PR #775, opened 2024-01-14
Add events pass-through feature in lambda-http crate

bnusunny:
*Issue #, if available:*

Related to https://github.com/awslabs/aws-lambda-web-adapter/issues/216 and https://github.com/awslabs/aws-lambda-web-adapter/issues/317 

*Description of changes:*

To support addtional event triggers in lambda web adapter, such as SQS or Bedrock Agents, we need to pass-through events from those triggers. This PR adds a `pass-through` feature to `lambda-http` crate. When enabled, the events payloads will be translated to a HTTP post request (event payloads in the HTTP body). 

This PR also improves the deserializer.rs. It uses `RawValue` to get the data from `deserializer`, not relay on private serde_json functions. 

### DIFF
--- a/lambda-http/Cargo.toml
+++ b/lambda-http/Cargo.toml
@@ -21,6 +21,7 @@ apigw_rest = []
 apigw_http = []
 apigw_websockets = []
 alb = []
+pass_through = []
 
 [dependencies]
 base64 = { workspace = true }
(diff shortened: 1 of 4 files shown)


bnusunny:
Thanks!

Issue #792, opened 2024-01-24
Deserialization of Lambda HTTP requests broke between 0.8 and 0.9 ... kinda


bassmanitram:
I don't _really_ think I'm gonna get a lot of sympathy for this one _but_ for a few months I had my own implementations of `lambda_http::request::from_reader` and `lambda_http::request::from_str` ... instead of:

```
serde_json::from_reader(rdr).map(LambdaRequest::into)
```

I used

```
simd_json::from_reader(rdr).map(LambdaRequest::into)
```

Performance-wise, simd_json blows serde_json out of the water - especially for complex deserializations like this. And it's supposed to be a drop-in replacement for situations like this. So the above works great at 0.8, buuuut at 0.9 I now get this for the simd_json version...

```
Err(
    Error {
        index: 0,
        character: None,
        error: Serde(
            "invalid type: newtype struct, expected any valid JSON value",
        ),
    },
)
```
(this is true for any `from` function in SIMD targeting the LambdaRequest)

[Here's a succinct tester GIST that shows the issue](https://gist.github.com/bassmanitram/fc58b50915e3266004a808431e34a050) - use lambda_http v0.8 and both have the same output. Use v0.9 and the second one fails.

Any pointers into how I might be able to debug this would be greatly received.

bassmanitram:
I've shown that [this line](https://github.com/awslabs/aws-lambda-rust-runtime/blob/baeaaec58aad8efd426119612ae69cffcef38b7b/lambda-http/src/deserializer.rs#L27) works with both simd_json and serde_json, so the heavy deserialization still works

I _suspect_ that it's something to do with the RawValue introduced into the [deserializer](https://github.com/awslabs/aws-lambda-rust-runtime/blob/baeaaec58aad8efd426119612ae69cffcef38b7b/lambda-http/src/deserializer.rs#L23). 

The reason I suspect RawValue is that searches on `invalid type: newtype struct, expected any valid JSON value` mention strange issues with that type. Anyway, I'm going to try and figure out precisely _what_ the issue is.

(I don't have passthru enabled)

I'm also going to have a go at an unstable feature to use simd_json instead of serde_json - and then try some proper benchmarks.

bassmanitram:
YUP that's where it is - on the Box::deserialize - something in simd_json isn't compatible with that ... still looking.

bassmanitram:
[...]

PR #795, opened 2024-01-25
Lambda http/optimize deserialization

bassmanitram:
*Issue #, if available:*
#792 

*Description of changes:*
Identifying different Lambda events that pertain to HTTP requests is notoriously
hard since they don't contain a simple `kind` property to help out. This is generally
true - not just a problem in Rust.

In lambda-http a full deserialization of each enabled request type is attempted in
turn, trying the next on failure of the previous, which has the possibility of being a
bit of a performance hog when multiple features are enabled.

Up to 0.8 of lambda-http, the deserialization was made in two stages - firstly using a 
private deserializer available in the serde crate to deserialize to an interim value. That
value was then passed to each enabled types's own `deserialize` associated function.

In 0.9 this was modified to be far cleaner, deserializing into a serde_json::RawValue
to check the general structure of the request JSON, then deserializing the contained
string slice into the type using standard serde_json functions..

However, this latter implementation entails two more downsides:

* The attempted type deserializations use the original string rather than already partially parsed-data
* The unintended consequence of disabling the possibility of using simd_json as the
  incoming deserializer - not exactly a supported feature anyway, but simd_json has a problem
  with RawValue.

This PR attempts to solve all these issues by:

* Deserializing into a serde_json::Value using the incoming deserializer - this fixes the simd_json issue
* Testing for "type determinants" in the Value instead of trying full deserialization - this avoids invoking the
  the full Serde deserialization framework multiple times until one works.
* Completing the deserialization to the discovered type from the Value - i.e. a partially -parsed internal 
  structure is used for for the second deserialization phase - much faster using using a string slice.
* This also allows for a much simpler "adoption" of simd_json, at least in the lambda_http package, because
  simd_json gets to do what it is design for - the string->Value parsing, while serde_json completes the 
  Value->event parsing.

The one possible drawback I can see in this PR is that if `Passthru` is enabled, rather than passing through
the original JSON string, that string is reconstructed from the parsed Value. You still get a JSON string but
it won't be _exactly_ the same as the input string. Others who know the project better than I can chime in on that.

### DIFF
--- a/lambda-http/Cargo.toml
+++ b/lambda-http/Cargo.toml
@@ -38,7 +38,7 @@ mime = "0.3"
 percent-encoding = "2.2"
 pin-project-lite = { workspace = true }
 serde = { version = "1.0", features = ["derive"] }
-serde_json = { version = "1.0", features = ["raw_value"] }
+serde_json = { version = "1.0" }
 serde_urlencoded = "0.7"
 tokio-stream = "0.1.2"
 url = "2.2"
--- a/lambda-http/src/deserializer.rs
+++ b/lambda-http/src/deserializer.rs
@@ -19,13 +19,13 @@ const ERROR_CONTEXT: &str = "this function expects a JSON payload from Amazon AP
 //
 enum RequestType {
     #[cfg(feature = "apigw_rest")]
-    ApiGatewayProxyRequest,
+    ApiGatewayProxy,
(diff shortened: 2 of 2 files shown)


bassmanitram:
Closing because the expected performance gains were not realized - v0.9 is, in fact, faster than v0.8 - by a long way.

PR #800, opened 2024-01-28
Support serialize UTF-8 String in HTTP Header Values

### LINKED ISSUE #786: thread 'main' panicked at 'lambda runtime failed: Error("failed to convert header to a str", line: 0, column: 0)'
We use the ARM Lambda runtime and https://github.com/awslabs/aws-lambda-web-adapter in combination with a Docker container to resize images on the fly. This works great, but occasionally (very rarely though) we see the following error in CloudWatch

```
thread 'main' panicked at 'lambda runtime failed: Error("failed to convert header to a str", line: 0, column: 0)', src/main.rs:25:2
```

image

Any pointers on how we might debug this are appreciated.

bnusunny:
*Issue #, if available:*

close #786

*Description of changes:*

Support serialize UTF-8 String in HTTP Header Values

### DIFF
--- a/lambda-events/src/custom_serde/headers.rs
+++ b/lambda-events/src/custom_serde/headers.rs
@@ -13,7 +13,7 @@ where
     for key in headers.keys() {
         let mut map_values = Vec::new();
         for value in headers.get_all(key) {
-            map_values.push(value.to_str().map_err(S::Error::custom)?)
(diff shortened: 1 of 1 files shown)


bnusunny:
Added a test to cover both `serialize_headers` and `serialize_multi_value_headers`.

PR #801, opened 2024-01-29
Make trigger reason for Cognito events Enum instead of String

ivan-kiselev:
*Description of changes:*

I was building on top of this library and it felt awkward to do the following: 

```Rust
    let payload = event.payload;
    match payload
        .cognito_event_user_pools_header
        .trigger_source
        .as_ref()
        .unwrap()
        .as_str()
    {
        "PostConfirmation_ConfirmSignUp" => { todo!() }
        "It's awkward and error prone to match on strings" => { todo!() }
    }

```

And so I thought I might just suggest a PR making it an enum? 

Let me know what you think.

I didn't really find relevant documentation, so let me know if there any in case this PR is a go.

Thanks!

### DIFF
--- a/lambda-events/src/event/cognito/mod.rs
+++ b/lambda-events/src/event/cognito/mod.rs
@@ -44,9 +44,20 @@ pub struct CognitoDatasetRecord {
 pub struct CognitoEventUserPoolsPreSignup {
     #[serde(rename = "CognitoEventUserPoolsHeader")]
     #[serde(flatten)]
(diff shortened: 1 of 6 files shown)


ivan-kiselev:
@calavera addressed in this commit https://github.com/awslabs/aws-lambda-rust-runtime/pull/801/commits/f8476b0f22eae228c19e5d28c5fa6cbb59edcee8

PR #809, opened 2024-02-05
Update lib.rs: Response vs Request in trace! log

lmammino:
*Issue #, if available:*

*Description of changes:* If I understand the code correctly, the tracing line reports the content of the **request**, NOT the response.

If that's correct, I believe we should apply this change to make debugging a little bit easier.

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -126,7 +126,7 @@ impl Runtime {
             // Group the handling in one future and instrument it with the span
             async {
                 let body = body.collect().await?.to_bytes();
-                trace!("response body - {}", std::str::from_utf8(&body)?);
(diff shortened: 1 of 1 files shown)


greenwoodcm:
the terminology is a little ambiguous.  this is the event payload that the Lambda function was invoked with (so in that regard it's a request).  but function code (or runtime code here) polls the Lambda runtime APIs for work and receives one payload at a time, as responses to that poll.  so it's a request to the Lambda function, but a response from the runtime API.  maybe we should change this line to say neither request nor response, but something like "event" or "payload".

greenwoodcm (approved):

greenwoodcm:
looks like you need to reformat

lmammino:
Sorry, read a bit too late. Thanks for taking care of it, @calavera

PR #827, opened 2024-02-21
Api Gateway authorizer improvements

calavera:
*Issue #, if available:*

The discussion in https://github.com/awslabs/aws-lambda-rust-runtime/discussions/823 made me think that we can provide much better ergonomics to work with Api Gateway Authorizers, since it looks like a common use case.

*Description of changes:*

There are changes in the events crate that are not backwards compatible, but I believe they provide much better user experience. I've divided this PR in three commits so you can check the changes out in isolation. There are more details in the commit messages.


bnusunny (approved):
LGTM

PR #828, opened 2024-02-21
Allow error response customization

kikuomax:
*Issue #, if available:*
- #824

*Description of changes:*
- `F::Error` of `Runtime::run` and `run` is required to implement `Into<Diagnostic<'a>>` instead of `std::fmt::Display`
- `Into<Diagnostic<'a>>` is implemented for any types that implement `std::fmt::Display` as a fallback and for backward compatibility
- `Diagnostic` and its fields become public so that users can customize error responses
- `error_type` and `error_message` of `Diagnostic` are wrapped in `std::borrow::Cow` so that they can hold both borrowed and owned strings. An example of the situation where an owned string is necessary is in the fallback implementation of `From<Display>` for `Diagnostic`, which generates `error_message` inside `from` method and it has to outlive the method call.

### DIFF
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -14,8 +14,9 @@ use hyper::{body::Incoming, http::Request};
 use lambda_runtime_api_client::{body::Body, BoxError, Client};
 use serde::{Deserialize, Serialize};
 use std::{
+    borrow::Cow,
     env,
-    fmt::{self, Debug, Display},
+    fmt::{self, Debug},
     future::Future,
     panic,
     sync::Arc,
@@ -33,7 +34,9 @@ pub mod streaming;
 mod types;
 
 use requests::{EventCompletionRequest, EventErrorRequest, IntoRequest, NextEventRequest};
-pub use types::{Context, FunctionResponse, IntoFunctionResponse, LambdaEvent, MetadataPrelude, StreamResponse};
+pub use types::{
+    Context, Diagnostic, FunctionResponse, IntoFunctionResponse, LambdaEvent, MetadataPrelude, StreamResponse,
+};
 
 use types::invoke_request_id;
 
@@ -96,7 +99,7 @@ impl Runtime {
     where
         F: Service<LambdaEvent<A>>,
         F::Future: Future<Output = Result<R, F::Error>>,
-        F::Error: fmt::Debug + fmt::Display,
+        F::Error: for<'a> Into<Diagnostic<'a>> + fmt::Debug,
         A: for<'de> Deserialize<'de>,
         R: IntoFunctionResponse<B, S>,
         B: Serialize,
@@ -173,7 +176,14 @@ impl Runtime {
                                 } else {
                                     "Lambda panicked".to_string()
                                 };
-                                EventErrorRequest::new(request_id, error_type, &msg).into_req()
+                                EventErrorRequest::new(
+                                    request_id,
+                                    Diagnostic {
+                                        error_type: Cow::Borrowed(error_type),
+                                        error_message: Cow::Owned(msg),
+                                    },
+                                )
+                                .into_req()
                             }
                         }
                     }
@@ -224,7 +234,7 @@ pub async fn run<A, F, R, B, S, D, E>(handler: F) -> Result<(), Error>
 where
     F: Service<LambdaEvent<A>>,
     F::Future: Future<Output = Result<R, F::Error>>,
-    F::Error: fmt::Debug + fmt::Display,
+    F::Error: for<'a> Into<Diagnostic<'a>> + fmt::Debug,
     A: for<'de> Deserialize<'de>,
     R: IntoFunctionResponse<B, S>,
     B: Serialize,
@@ -249,15 +259,12 @@ fn type_name_of_val<T>(_: T) -> &'static str {
     std::any::type_name::<T>()
 }
 
-fn build_event_error_request<T>(request_id: &str, err: T) -> Result<Request<Body>, Error>
+fn build_event_error_request<'a, T>(request_id: &'a str, err: T) -> Result<Request<Body>, Error>
 where
-    T: Display + Debug,
+    T: Into<Diagnostic<'a>> + Debug,
 {
     error!("{:?}", err); // logs the error in CloudWatch
-    let error_type = type_name_of_val(&err);
-    let msg = format!("{err}");
-
-    EventErrorRequest::new(request_id, error_type, &msg).into_req()
+    EventErrorRequest::new(request_id, err).into_req()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


lambda-runtime/src/lib.rs
```diff
@@ -14,8 +14,9 @@ use hyper::{body::Incoming, http::Request};
 use lambda_runtime_api_client::{body::Body, BoxError, Client};
 use serde::{Deserialize, Serialize};
 use std::{
+    borrow::Cow,
```
evbo, line 17: Why use Cow? Why not just use `&str`?

kikuomax: @evbo I introduced `Cow` to overcome a lifetime issue rather than performance concern. Since the fallback conversion from `Error` (`Display`) into `Diagnostic` dynamically formats `error_message`, I was not able to simply return `&str`: https://github.com/awslabs/aws-lambda-rust-runtime/blob/e0e95e61cba0a8a695d4766d9215cbe839caf3a7/lambda-runtime/src/types.rs#L67-L77

However, regarding `error_type`, it could have been `&str`. I did not notice that until you pointed it out.

evbo: `error_type` can't always be `&str` in the case where you're borrowing the variant name (such as with strum::AsRefStr), which is helpful for auto-converting Variants to error types.

I'm not sure why both can't just be `String`? Just seems like a lot of extra typing using `Cow`, sorry to nitpick

evbo:
I think I see the underlying problem now... You can't derive `Display` for the custom error enum. If you do, you will conflict with the fallback `Into` `impl` for `Diagnostic`...

This is a pretty big issue for anyone wanting a custom `Display` implemented for their errors, especially if using `thiserror`

Any decent workarounds?

kikuomax:
@evbo Have you tried something like the following?
```rust
#[derive(thiserror::Error, Debug)]
pub enum MyError {
    #[error("some error: {0}")]
    SomeError(String),
    // ... other variants
}

impl<'a> Into<Diagnostic<'a>> for MyError {
    fn into(self) -> Diagnostic<'a> {
        match self {
            Self::SomeError(s) => Diagnostic {
                error_type: Cow::Borrowed("SomeError"),
                error_message: Cow::Owned(s),
            },
            // ... other variants
        }
    }
}
```

The above will cause conflict between `From<Display> for Diagnostic` and `Into<Diagnostic> for MyError`, because `MyError` implements `Display`. So you cannot directly implement `Diagnostic` for `MyError`; more generally, any type implementing [`std::error::Error`](https://doc.rust-lang.org/std/error/trait.Error.html).

A workaround in this case may be to define a wrapper struct for your error type and implement `Diagnostic` for the wrapper like in the example of `Diagnostic`:
```rust
struct ErrorResponse(MyError);

impl<'a> Into<Diagnostic<'a>> for ErrorResponse {
    fn into(self) -> Diagnostic<'a> {
        match self.0 {
            ErrorResponse::SomeError(s) => Diagnostic {
                error_type: Cow::Borrowed("SomeError"),
                error_message: Cow::Owned(s),
            },
            // ... other variants
        }
    }
}
```
`Display` must not be implemented for `ErrorResponse` to avoid the same conflict.

Obviously, I should refine the documentation.

lambda-runtime/src/lib.rs
kikuomax: I think some might say the opposite "why `String`? they are `&'static str` most of the time."

Btw, you may avoid explicit use of `Cow` as long as you are dealing with `&str` or `String`:
```rust
let msg = "message".to_string();
Diagnostic {
    error_type: "SomeError".into(), // &str → Cow::Borrowed
    error_message: msg.into(), // String → Cow::Owned
};
```

evbo:
@kikuomax I'm sorry, your example did show a top-level struct wrapper.

I guess the issue is all of my business logic would then need to instantiate `ErrorResponse`, or I would need to derive `From`/`Into` for `ErrorResponse`

So if I have to implement `From` for `ErrorResponse` it becomes the same problem I was trying to use `thiserror` for on `MyError`. Do you see my point?

My current workaround is just don't use `thiserror` (since it requires `Display`) and rely on pretty formatted `Debug`, `{:#?}`, instead of Display. It's not ideal but it enables the most ergonomics since I can easily auto-convert using `?`

Oh, and I use `strum` to get the enum variant as a string for `error_type` - instead of `"SomeError"` hard-coded like you've shown. In that case `error_type` must be owned, hence why I feel `error_type` and `error_message` might as well both just be `String`

lambda-runtime/src/lib.rs
evbo: Thanks for the `into` - that's much cleaner. I hate to burn up so much time, but why would someone say: "why String"? If we're creating millions of exceptions, then I understand the cost of heap allocation would eventually cost noticeably, but maybe at most 1 exception gets generated once in a blue moon, so isn't it fair using `String`?

evbo:
@kikuomax I take back my earlier messages. I think you're right - this can be solved with better docs.

I think the best ergonomics are achieved when `function_handler` returns the error and then you `map_err` to the top-level struct by intercepting what gets returned to `lambda_runtime::run` as shown here:
https://docs.aws.amazon.com/lambda/latest/dg/rust-handler.html#rust-shared-state

kikuomax:
@evbo I am planning a PR to improve the documentation. How about having the following in the API documentation? Does it make better sense?

`Diagnostic` is automatically generated from any types that implement [`Display`](https://doc.rust-lang.org/1.76.0/core/fmt/trait.Display.html); e.g., [`Error`](https://doc.rust-lang.org/1.76.0/core/error/trait.Error.html). The default behavior puts the type name of the original error obtained with [`std::any::type_name`](https://doc.rust-lang.org/1.76.0/core/any/fn.type_name.html) into `error_type`, which may not be reliable for conditional error handling. You can define your own error container that implements `Into<Diagnostic>` if you need more accurate error types.

Example:

```rust
use lambda_runtime::{service_fn, Diagnostic, Error, LambdaEvent};

#[derive(Debug)]
struct ErrorResponse(Error);

impl<'a> Into<Diagnostic<'a>> for ErrorResponse {
    fn into(self) -> Diagnostic<'a> {
        Diagnostic {
            error_type: "MyError".into(),
            error_message: self.0.to_string().into(),
        }
    }
}

async fn function_handler(_event: LambdaEvent<()>) -> Result<(), Error> {
   // ... do something
   Ok(())
}

#[tokio::main]
    lambda_runtime::run(service_fn(|event| async {
        function_handler(event).await.map_err(ErrorResponse)
    })).await
}
```

The container `ErrorResponse` in the above snippet is necessary because you cannot directly implement `Into<Diagnostic>` for types that implement [`Error`](https://doc.rust-lang.org/1.76.0/core/error/trait.Error.html). For instance, the following code will not compile:

ⓘ compile_fail
```rust
use lambda_runtime::Diagnostic;

#[derive(Debug)]
struct MyError(String);

impl std::fmt::Display for MyError {
        write!(f, "MyError: {}", self.0)
    }
}

impl std::error::Error for MyError {}

// ERROR! conflicting implementations of trait `Into<Diagnostic<'_>>` for type `MyError`
impl<'a> Into<Diagnostic<'a>> for MyError {
    fn into(self) -> Diagnostic<'a> {
        Diagnostic {
            error_type: "MyError".into(),
            error_message: self.0.into(),
        }
    }
}
```

PR #852, opened 2024-04-01
fix: do not duplicate headers (single val, multi val) on return

jreppnow:
*Issue*

This is the Rust equivalent to these changes made in the Go runtime library for lambda HTTP: https://github.com/awslabs/aws-lambda-go-api-proxy/pull/73 

The issue here is that the API Gateway does not properly merge single-value headers and multi-value returned by the Lambda and instead returns both to the client. Since this library just puts the same headers into both, this means that every header is returned twice (more or less, in the case of actual multi-value headers, only the first element is duplicated). In the case of CORS, this is fatal, as browsers rejects duplicate CORS headers. Unless this change can go through, we would be forced to turn off application-level CORS handling and rely on the API gateway instead. 

*Description of changes:*

Adapts the code to return all headers as multi-value headers only, returning an empty `HeaderMap` for the single-value headers. 

### DIFF
--- a/lambda-http/src/response.rs
+++ b/lambda-http/src/response.rs
@@ -69,7 +69,7 @@ impl LambdaResponse {
                 body,
                 is_base64_encoded,
                 status_code: status_code as i64,
-                headers: headers.clone(),
+                headers: HeaderMap::new(),
                 multi_value_headers: headers,
             }),
             #[cfg(feature = "apigw_http")]
@@ -91,7 +91,7 @@ impl LambdaResponse {
                     is_base64_encoded,
                     status_code: status_code as i64,
                     cookies,
-                    headers: headers.clone(),
+                    headers: HeaderMap::new(), 
                     multi_value_headers: headers,
                 })
             }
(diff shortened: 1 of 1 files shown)


jreppnow:
If required, I could also provide logic that actually splits the headers into single-value and multi-value headers and returns them as they should be returned.

jreppnow:
Yup, that's the specification, but the real implementation does not behave that way unfortunately.

jreppnow:
In fact, you can inspect the issue in action if you navigate to this endpoint, which is served via the Lambda in question: 

https://compass.api.labbase.jp/health_check 

You will note how a lot of the response header (including the CORS-relevant allow-credentials) are duplicated. The ones not duplicated are the ones added or updated by CloudFlare, which is sitting in-between.

jreppnow:
@calavera Thanks a lot for the fast response! We have re-built our application with a git dependency on this crate (and the intermittent one) and deployed it. You can confirm that the header duplication has indeed been resolved by visiting the above link again.

PR #860, opened 2024-04-17
ApiGatewayV2CustomAuthorizerV2Request identity_source as optional

HVKukkonen:
*Issue #, if available:* None

*Description of changes:*
Fixes typing issue with `ApiGatewayV2CustomAuthorizerV2Request` not allowing for `None` as the `identity_source`. I suspect this being an issue as per the [docs](https://docs.aws.amazon.com/apigateway/latest/developerguide/configure-api-gateway-lambda-authorization-with-console.html)

> When caching is turned off, it is not necessary to specify an identity source. API Gateway directly passes the request to the authorizer Lambda function.

Also, I have run into issues with this in the wild as not having the identity source in the API Gateway Terraform configuration results in `DeserializeError { inner: Error { path: Path { segments: [Map { key: \"identitySource\" }] }, original: Error(\"invalid type: null, expected a sequence\"` in my Authorizer code. And after specifying an arbitrary field from the API Gateway event as the identity source it worked fine. 

### DIFF
--- a/lambda-events/src/event/apigw/mod.rs
+++ b/lambda-events/src/event/apigw/mod.rs
@@ -1011,6 +1011,16 @@ mod test {
 
     #[test]
     #[cfg(feature = "apigw")]
+    fn example_apigw_v2_custom_authorizer_v2_request_without_identity_source() {
+        let data = include_bytes!("../../fixtures/example-apigw-v2-custom-authorizer-v2-request-without-identity-source.json");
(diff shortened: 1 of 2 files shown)


bnusunny:
Could you add a test to cover this case?

HVKukkonen:
Added. Of course, nothing guarantees that the fixture used exactly matches the event from the API Gateway but that's the same issue with all of the fixtures, right? :smiling_face_with_tear:

bnusunny:
Could you fix the linting errors?

bnusunny (approved):

Issue #889, opened 2024-06-04
Error: Records[0].dynamodb: missing field `ApproximateCreationDateTime`


vstepanyuk:
According documentation https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_streams_StreamRecord.html
this field is optional

PR #891, opened 2024-06-06
[DynamoDB event] Makes ApproximateCreationDateTime optional in StreamRecord

jfkisafk:
*Issue #, if available:* closes https://github.com/awslabs/aws-lambda-rust-runtime/issues/889

*Description of changes:* Makes the DDB event [StreamRecord](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-events/src/event/dynamodb/mod.rs#L214-L247) more robust, b/c it appears that `ApproximateCreationDateTime` can be absent. The DDB docs are ambiguous in this matter as they suggest it is not required but also say:

> In the following list, the required parameters are described first.

Maybe we can get more guidance from DDB team?

- [x] I confirm that I've run the tests locally using `cargo test`
- [x] I confirm that I've run the lint locally using `cargo fmt`

### DIFF
--- a/lambda-events/src/event/dynamodb/mod.rs
+++ b/lambda-events/src/event/dynamodb/mod.rs
@@ -213,9 +213,11 @@ pub struct UserIdentity {
 #[serde(rename_all = "camelCase")]
 pub struct StreamRecord {
     /// The approximate date and time when the stream record was created, in UNIX
-    /// epoch time (http://www.epochconverter.com/) format.
+    /// epoch time (http://www.epochconverter.com/) format. Might not be present in
+    /// the record: https://github.com/awslabs/aws-lambda-rust-runtime/issues/889
     #[serde(rename = "ApproximateCreationDateTime")]
     #[serde(with = "float_unix_epoch")]
+    #[serde(default)]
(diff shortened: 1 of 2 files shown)


lambda-events/src/event/dynamodb/mod.rs
```diff
@@ -213,9 +213,11 @@ pub struct UserIdentity {
 #[serde(rename_all = "camelCase")]
 pub struct StreamRecord {
     /// The approximate date and time when the stream record was created, in UNIX
-    /// epoch time (http://www.epochconverter.com/) format.
+    /// epoch time (http://www.epochconverter.com/) format. Might not be present in
+    /// the record: https://github.com/awslabs/aws-lambda-rust-runtime/issues/889
     #[serde(rename = "ApproximateCreationDateTime")]
     #[serde(with = "float_unix_epoch")]
+    #[serde(default)]
     pub approximate_creation_date_time: DateTime<Utc>,
```
jfkisafk, line 221: Did not want to change contracts by making this `Option<DateTime<Utc>>`, but if that makes more sense please let me know.

PR #896, opened 2024-06-19
Make child-spans work as expected when using the lambda-runtime

TethysSvensson:
*Issue #, if available:* N/A

*Description of changes:* Implementing a sub-span directly in the `service_fn` does not currently set the parent-span in the expected way. Example:

```rust
let func = service_fn(move |event: LambdaEvent<T>| {
    let event_span = create_event_span(outer_variable);
    inner_handler(event, state.clone()).instrument(event_span)
});
```

The problem is, that at the time where the `create_event_span()` function is called, the span returned by `request_span()` has not been entered, so it will be not be set as the parent.

The solution is to enter the span before calling the inner service. This also matches e.g. how [`tower_http` does it](https://github.com/tower-rs/tower-http/blob/a1e3f8a28bdb118c98627b8fa2d05180049dc981/tower-http/src/trace/service.rs#L308).

### DIFF
--- a/lambda-runtime/src/layers/trace.rs
+++ b/lambda-runtime/src/layers/trace.rs
@@ -43,7 +43,11 @@ where
 
     fn call(&mut self, req: LambdaInvocation) -> Self::Future {
         let span = request_span(&req.context);
-        self.inner.call(req).instrument(span)
+        let future = {
+            let _guard = span.enter();
+            self.inner.call(req)
+        };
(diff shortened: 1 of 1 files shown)


TethysSvensson:
And thank you for accepting it!

Are you by any chance planning to do a release for this?

I'm asking because if you are, then I would rather wait for your release than create a work-around in our internal code.

PR #897, opened 2024-06-22
Relax blanket implementation of Diagnostic

### LINKED ISSUE #880: Custom `impl Into<Diagnostic>` for types that already implement `Display`
The `Diagnostic` docs say:

> You can define your own error container that implements `Into<Diagnostic>` if you need to handle errors based on error types.

I'm trying to do just that.  However my error container already has a `Display` implementation (via `thiserror`).  Because of the blanket `impl<'a, T> From<T> for Diagnostic<'a> where T: Display`, I run into a conflicting trait implementations error.

The error makes perfect sense, but I also see no way around it without removing the `Display` implementation from my code, which I'd like to keep.  I feel like it's not that unusual to want a custom `Into<Diagnostic>` for `Display` types.  Do I have other options here?

calavera:
*Issue #, if available:*

Fixes #880 

*Description of changes:*

Instead of implementing Diagnostic for everything that implements Display, implement the trait only for a few well known types. This gives people more flexibility to implement Diagnostic.


lambda-runtime/src/diagnostic.rs
```diff
@@ -0,0 +1,123 @@
+use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
+
+use crate::{deserializer::DeserializeError, Error};
+
+/// Diagnostic information about an error.
+///
+/// `Diagnostic` is automatically derived for types that implement
+/// [`Display`][std::fmt::Display]; e.g., [`Error`][std::error::Error].
```
tannerntannern, line 9: Is this comment true anymore?

bnusunny:
@calavera is this ready for review?

bnusunny (approved):

TethysSvensson:
This is quite an intrusive change, as it means you can no longer use e.g. anyhow or eyre in your handlers, at least without doing additional boxing.

Would you be open to doing `From<...>` impls for those crates specifically, as I don't think they would be interesting in adding a dependency for `lambda_runtime` themselves.

TethysSvensson:
Also, the requirement `for<'b> Into<Diagnostic<'b>>` is effectively the same as the requirement `Into<Diagnostic<'static>>` but less ergonomic.

For instance you will not be able to specify `Diagnostic<'static>`. This type does implement `Into<Diagnostic<'static>>`, but does not fulfill the requirement `for<'b> Into<Diagnostic<'b>>`.

TethysSvensson:
For anyone else having a hard time upgrading to v0.12 because of this PR, here is my work-around:

```rust
struct DiagnosticWrapper(Diagnostic<'static>);

impl<'a> From<DiagnosticWrapper> for Diagnostic<'a> {
    fn from(value: DiagnosticWrapper) -> Self {
        value.0
    }
}

impl std::fmt::Debug for DiagnosticWrapper {
        self.0.fmt(f)
    }
}

impl DiagnosticWrapper {
    fn new<E>(err: E) -> Self
    where
        E: Display,
    {
        DiagnosticWrapper(Diagnostic {
            error_type: Cow::Borrowed(std::any::type_name::<E>()),
            error_message: Cow::Owned(err.to_string()),
        })
    }
}
```

I then use `handler_future.map_err(DiagnosticWrapper::new)` inside my service function.

TethysSvensson:
@calavera I've created #901 to continue this discussion.

PR #898, opened 2024-06-25
Add OpenTelemetry layer as optional

### LINKED ISSUE #866: Add OpenTelemetryLayer to lambda_runtime layers
Hi, I have started testing OpenTelemetry integration and I found the example given here really useful: https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/examples/opentelemetry-tracing/src/lib.rs

I have copied the code into my project but I am wondering if it could be packaged into the distributed crate, maybe behind an opentelemetry feature flag?

I have also tried using nested layers and I would expect this to create child spans but this is not what I am observing. Is this a feature or a bug?

Given this setup I would expect the span of Tracing-layer to be a child of the OpenTelemetry-layer span:
```rust
 let runtime = Runtime::new(handler)
            .layer(layers::TracingLayer::new())
            .layer(OpenTelemetryLayer::new(|| {
                tracer
                    .provider()
                    .expect("Unable to get tracer provider!")
                    .force_flush();
            }));
```

But this is the result (function_name is customer-lookup):
Screenshot 2024-05-02 at 08 11 20

calavera:
*Issue #, if available:*

Closes #866 

*Description of changes:*

- Add the OpenTelemetryLayer as an optional module to the runtime.
- Fixes issue creating multiple spans instead of assigning one as parent.

### DIFF
--- a/lambda-runtime/src/layers/otel.rs
+++ b/lambda-runtime/src/layers/otel.rs
@@ -2,7 +2,7 @@ use std::future::Future;
 use std::pin::Pin;
 use std::task;
 
-use lambda_runtime::LambdaInvocation;
+use crate::LambdaInvocation;
 use opentelemetry_semantic_conventions::trace as traceconv;
 use pin_project::pin_project;
 use tower::{Layer, Service};
@@ -19,6 +19,7 @@ impl<F> OpenTelemetryLayer<F>
 where
     F: Fn() + Clone,
 {
+    /// Create a new [OpenTelemetryLayer] with the provided flush function.
     pub fn new(flush_fn: F) -> Self {
         Self { flush_fn }
     }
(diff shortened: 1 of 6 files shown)


bnusunny (approved):

lambda-runtime/src/layers/otel.rs
```diff
@@ -71,9 +72,14 @@ where
         // After the first execution, we can set 'coldstart' to false
         self.coldstart = false;
 
-        let fut = self.inner.call(req).instrument(span);
+        let future = {
+            // Enter the span before calling the inner service
+            // to ensure that it's assigned as parent of the inner spans.
+            let _guard = span.enter();
```
Oliboy50, line 78: I didn't know that using `.instrument(span)` was not enough when calling a future, where did you find docs about that?

And since the span is entered before starting to poll the future, it will increase the span duration, right?

lambda-runtime/src/layers/otel.rs
Oliboy50, line 1: line `67`: 

```
{ traceconv::FAAS_TRIGGER } = "http",
```

I think it should be configurable because the `lambda-runtime` crate can be used without using HTTP triggers (it could defaults to `http` when using the `lambda_http` crate though)

Oliboy50: 👍 ➡️  https://github.com/awslabs/aws-lambda-rust-runtime/pull/903

PR #903, opened 2024-06-28
feat(otel): allow to configure the faas.trigger attribute of the span

Oliboy50:
📬 *Issue #, if available:*
https://github.com/awslabs/aws-lambda-rust-runtime/pull/898#discussion_r1654283275

✍️ *Description of changes:*
Allow to configure the `faas.trigger` attribute of the OpenTelemetry layer span which is currently hardcoded to `http`

### DIFF
--- a/lambda-runtime/src/layers/otel.rs
+++ b/lambda-runtime/src/layers/otel.rs
@@ -10,6 +10,7 @@ use tracing::{instrument::Instrumented, Instrument};
 /// a function to flush OpenTelemetry after the end of the invocation.
 pub struct OpenTelemetryLayer<F> {
     flush_fn: F,
+    otel_attribute_trigger: Option<String>,
 }
 
 impl<F> OpenTelemetryLayer<F>
@@ -18,7 +19,20 @@ where
 {
     /// Create a new [OpenTelemetryLayer] with the provided flush function.
     pub fn new(flush_fn: F) -> Self {
-        Self { flush_fn }
+        Self {
+            flush_fn,
+            otel_attribute_trigger: None,
+        }
+    }
+
+    /// Configure the `faas.trigger` attribute of the OpenTelemetry span.
+    /// Defaults to `http` if not set.
+    /// See https://opentelemetry.io/docs/specs/semconv/attributes-registry/faas/ for the list of possible triggers.
+    pub fn with_trigger<T: Into<String>>(self, trigger: T) -> Self {
+        Self {
+            otel_attribute_trigger: Some(trigger.into()),
+            ..self
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


examples/opentelemetry-tracing/src/main.rs
```diff
@@ -24,10 +24,15 @@ async fn main() -> Result<(), BoxError> {
...
-        tracer_provider.force_flush();
-    }));
+    let runtime = Runtime::new(service_fn(echo)).layer(
+        // Create a tracing span for each Lambda invocation
+        OtelLayer::new(|| {
+            // Make sure that the trace is exported before the Lambda runtime is frozen
+            tracer_provider.force_flush();
+        })
+        // Set the "faas.trigger" attribute of the span (defaults to "http" otherwise)
+        .with_trigger("datasource"),
```
Oliboy50, line 34: Capture d’écran 2024-06-28 à 15 37 34

lambda-runtime/src/layers/otel.rs
```diff
@@ -33,6 +47,10 @@ where
             inner,
             flush_fn: self.flush_fn.clone(),
             coldstart: true,
+            otel_attribute_trigger: self
+                .otel_attribute_trigger
+                .clone()
+                .unwrap_or_else(|| "http".to_string()),
```
Oliboy50, line 53: I have a doubt here, I don't know if the `OpenTelemetryLayer::layer()` is executed each time the lambda is invoked... or if it's call only on coldstarts during the `Runtime` construction

because if it's called for each lambda invocation, maybe we should do best than that to avoid performance issues?

Oliboy50: > It makes me wonder if datasource should be the default, it's more generic than http

Yes `datasource` or even `other`, but this would be a breaking change, so let me know if you really want me to change the current default value 🙏 

> looking at the spec, maybe using an enum for the datasource makes more sense since there are only a few trigger types defined

The OpenTelemetry specs do not seem to move fast, but if they do, don't you think that this could be an issue for the users to wait for the aws-lambda-runtime releases to keep up with the OpenTelemetry specs evolution? 
Also note that the current specs for `faas.trigger` are still `experimental`... and even the `opentelemetry_semantic_conventions` does not expose such an enum yet 🤷 
That being said, let me know if we should expose an enum here (should we name it `OpenTelemetryFaasTrigger`?) 🙏

examples/opentelemetry-tracing/src/main.rs
Oliboy50: updated to use "pubsub" since "datasource" is the new default value
Capture d’écran 2024-06-30 à 15 25 35

lambda-runtime/src/layers/otel.rs
Oliboy50: done 👍

PR #907, opened 2024-07-05
Error handling improvements

### LINKED ISSUE #901: Regression in v0.12.0: External error types no longer work with lambda_runtime::run
With the release of v0.12.0, it is no longer possible to use error types defined externally to your own crate. This is caused by the inclusion of #897.

Example:

```rust
// Works in both v0.11.3 and v0.12.0
type Error = Box<dyn std::error::Error>;

// Works in v0.11.3 but not in v0.12.0
// type Error = std::io::Error;

// Works in v0.11.3 but not in v0.12.0
// type Error = anyhow::Error;

// Works in neither v0.11.3 nor v0.12.0
// type Error = lambda_runtime::Diagnostic<'static>;

async fn handler(_event: lambda_runtime::LambdaEvent<()>) -> Result<(), Error> {
    Ok(())
}

#[tokio::main]
async fn main() {
    lambda_runtime::run(lambda_runtime::service_fn(handler))
        .await
        .unwrap();
}
```

Note that it is not possible to implement `Into<Diagnostic<'a>>` yourself for any of these types because of the orphan rules.

calavera:
📬 *Issue #, if available:*

Fixes #901

✍️ *Description of changes:*

Change how we handle error types to be more ergonomic:

- Replace Cow with String: Diagnostics are serialized into JSON as soon as the function returns, which means that their value is copied right away. The performance improvement of using Cow is minimal in this case, but it has some ergonomic implications because we have to handle their lifetimes. By removing the explicit lifetimes, people can return Diagnostic values with static lifetimes which was not possible before.

- Add features to implement `From<T> for Diagnostic` for anyhow, eyre, and miette error types. This helps people that use those creates to transform their errors into `Diagnostic` without double boxing their errors.

### DIFF
--- a/lambda-runtime/Cargo.toml
+++ b/lambda-runtime/Cargo.toml
@@ -15,11 +15,11 @@ readme = "../README.md"
 
 [features]
 default = ["tracing"]
-tracing = ["lambda_runtime_api_client/tracing"]
-opentelemetry = ["opentelemetry-semantic-conventions"]
-anyhow = ["dep:anyhow"]
-eyre = ["dep:eyre"]
-miette = ["dep:miette"]
+tracing = ["lambda_runtime_api_client/tracing"] # enables access to the Tracing utilities
+opentelemetry = ["opentelemetry-semantic-conventions"] # enables access to the OpenTelemetry layers and utilities
+anyhow = ["dep:anyhow"] # enables From<T> for Diagnostic for anyhow error types, see README.md for more info
+eyre = ["dep:eyre"] # enables From<T> for Diagnostic for eyre error types, see README.md for more info
+miette = ["dep:miette"] # enables From<T> for Diagnostic for miette error types, see README.md for more info
 
 [dependencies]
 anyhow = { version = "1.0.86", optional = true }
--- a/lambda-http/Cargo.toml
+++ b/lambda-http/Cargo.toml
@@ -22,10 +22,11 @@ apigw_http = []
 apigw_websockets = []
 alb = []
 pass_through = []
-tracing = ["lambda_runtime/tracing"]
-anyhow = ["lambda_runtime/anyhow"]
-eyre = ["lambda_runtime/eyre"]
-miette = ["lambda_runtime/miette"]
+tracing = ["lambda_runtime/tracing"] # enables access to the Tracing utilities
+opentelemetry = ["lambda_runtime/opentelemetry"] # enables access to the OpenTelemetry layers and utilities
+anyhow = ["lambda_runtime/anyhow"] # enables From<T> for Diagnostic for anyhow error types, see README.md for more info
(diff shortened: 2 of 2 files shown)


TethysSvensson:
I think this is a overall pretty good trade-off.

* I like the improved ergonomics by getting rid of the lifetime on Diagnostic. This also means that you can now directly return a diagnostic from your handlers.
* I like the improved docs
* I am a bit confused about why you would need to use both `IntoDiagnostic` and `Into<Diagnostic>`. Wouldn't it be possible to just use `IntoDiagnostic` for everything?

README.md
```diff
@@ -71,6 +71,61 @@ async fn func(event: LambdaEvent<Value>) -> Result<Value, Error> {
...
+
+To get more descriptive `error_type` fields, you can implement `Into<Diagnostic>` for your error type. That gives you full control on what the `error_type` is:
+
+```rust
+use lambda_runtime::{Diagnostic, Error, LambdaEvent};
+
+#[derive(Debug)]
+struct ErrorResponse(&'static str);
+
+impl Into<Diagnostic> for ErrorResponse {
```
TethysSvensson, line 97: You should probably recommend that people do `impl From<ErrorResponse> for Diagnostic` instead, since this is not normal rust convertion, because then you get both the `Into` impl for free.

tim3z:
Hi, thanks for improving this, looks a lot clearer and better to me. Docs helped me as a naive user a lot to understand the point of the breaking change.

One thing confuses me still though, the `impl<T> From<Box<T>> for Diagnostic where T: std::error::Error` blanket impl. Why should it be impossible to convert some `T: std::error::Error` but possible for `Box<T> where T: std::error::Error`? This was already my primary confusion when the breaking change came up. Returning an error didn't work anymore, but slapping a `Box` on it would fix it. But what has a `Box` to do with the error handling? This blanket impl makes it also impossible to implement `impl From<Box<MyErrorType>> for Diagnostic` (just tried with the thiserror example). 

I think, it would make more sense to implement neither. And if the user really wants to use the simpel catchall conversion it is possible to go through `Box<dyn std::error::Error>`. That makes it explicit that you get to an impl by going through the trait object, not by slapping on a `Box`.

bnusunny (approved):

lambda-runtime/Cargo.toml
```diff
@@ -17,11 +17,16 @@ readme = "../README.md"
 default = ["tracing"]
 tracing = ["lambda_runtime_api_client/tracing"]
 opentelemetry = ["opentelemetry-semantic-conventions"]
+anyhow = ["dep:anyhow"]
```
rimutaka, line 20: Since there is no standard way of documenting features yet (https://github.com/rust-lang/rfcs/pull/3485), shall we add `# enables From<T> for Diagnostic for anyhow error types, see ReadMe for more info` here?
Add the same for eyre, and miette or merge into a single comment.

The features are described in the ReadMe, but that's likely to be missed on the first read.

Issue #917, opened 2024-08-18
[RFC] Ergonomic improvements to payload deserialization for HTTP use cases


mlaota:
## Summary
This RFC proposes ergonomic improvements to payload deserialization for HTTP usecases.

## Motivation
The [`RequestPayloadExt`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/d7c53b2d448421a4a4680c9d77eebb25ccfb9e55/lambda-http/src/ext/request.rs) trait is used in `lambda_http` to simplify extracting a `Request`'s body into
a user-provided struct. The current version of the [`RequestPayloadExt::payload`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/d7c53b2d448421a4a4680c9d77eebb25ccfb9e55/lambda-http/src/ext/request.rs#L102) has the following docstring:

>Return the result of a payload parsed into a type that implements [serde::Deserialize](vscode-file://vscode-app/c:/Users/laota/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html)
>
>Currently only application/x-www-form-urlencoded and application/json flavors of content type are supported
>
>A [PayloadError](vscode-file://vscode-app/c:/Users/laota/AppData/Local/Programs/Microsoft%20VS%20Code/resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) will be returned for undeserializable payloads. If no body is provided, Ok(None) will be returned.

The interpretation of how `payload` should behave when the content-type header is missing/unsupported is somewhat ambiguous. For example, there's no mention of what happens when the content-type header is missing. According to [RFC 9110: HTTP Semantics](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.3-5),

> If a Content-Type header field is not present, the recipient MAY [...] examine the data to determine its type.

so, it's not unreasonable to assume that a generic `payload` function on an HTTP `Request` struct would also do a best-effort estimation of the content-type. The signature of `payload` asks to handle multiple `PayloadError` variants for different content types which may reinforce that assumption.

The actual behavior is that when the content-type header is either missing or unsupported, the implementation assumes there is no content and returns `Ok(None)`; however, the docstring only calls out `Ok(None)` being returned "if no body is provided" which can lead to frustration when trying to figure out why the payload is not being recognized.

[...]

PR #921, opened 2024-09-02
feat: ergonomic improvements to HTTP payload deserialization

mlaota:
📬 *Issue #, if available:*
#917 

✍️ *Description of changes:*
Ergonomic improvements to HTTP payload deserialization with intent-driven helper methods (`json` and `form_url_encoded` are the initial implementations). Details in RFC (#917).

Note: there was a bug in `RequestPayloadExt::payload` where if the content-type header was present but the body was empty, it would try to parse it anyways, resulting in a PayloadError. The bug was fixed when I updated the function to delegate to `json` and `form_urlencoded`. I added a couple unit tests in this PR for that scenario to ensure the behavior doesn't regress if its implementation drifts from the new functions in the future.

- [X] I confirm that I've ran `cargo +nightly fmt`.
- [X] I confirm that I've ran `cargo clippy --fix`.

### DIFF
--- a/lambda-http/src/ext/request.rs
+++ b/lambda-http/src/ext/request.rs
@@ -20,6 +20,36 @@ pub enum PayloadError {
     WwwFormUrlEncoded(SerdeError),
 }
 
+/// Indicates a problem processing a JSON payload.
+#[derive(Debug)]
+pub enum JsonPayloadError {
+    /// Problem deserializing a JSON payload.
+    Parsing(serde_json::Error),
+}
+
+/// Indicates a problem processing an x-www-form-urlencoded payload.
+#[derive(Debug)]
+pub enum FormUrlEncodedPayloadError {
+    /// Problem deserializing an x-www-form-urlencoded payload.
+    Parsing(SerdeError),
+}
+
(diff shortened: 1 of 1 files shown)


calavera (approved):
Thanks for adding all those tests and new documentation! ✨

lambda-http/src/ext/request.rs
```diff
@@ -20,6 +20,39 @@ pub enum PayloadError {
     WwwFormUrlEncoded(SerdeError),
 }
 
+/// Alias for a type-erased error type.
+pub type BoxError = Box<dyn std::error::Error + Send + Sync>;
```
calavera, line 24: I believe this is not used, can you remove it? We also have a similar type already exposed in lambda-http/lib.rs.

mlaota: Good catch! Removed. 

(I was considering a generic `Other` variant for the new error types that would keep a box error. Decided to keep the errors as a single-variant enum to leave the option open without forcing users to handle an error path would have been currently impossible, but let me know if that's something you'd actually prefer)

calavera (approved):

PR #925, opened 2024-09-16
Use event filter

bassmanitram:
📬 *Issue #, if available:*

✍️ *Description of changes:*
Originally a `Level` was parsed from one of two environment variables (or defaulted) and then converted into a `LevelFilter` before initializing the subscriber.

However, this precludes using `RUST_LOG=off` since `Level` does not recognize that as valid, resulting in `Level::INFO` (the default) being used.

`LevelFilter` (to which the above is converted anyway) _does_ allow the value to be `off` - so it seems a little more flexible (and very very minutely faster) to parse the env var, or default value, directly into a `LevelFilter`.


calavera (approved):
Thanks!

PR #933, opened 2024-10-17
Only add URL query if it's not empty

RaeesBhatti:
📬 *Issue #, if available:*

✍️ *Description of changes:*
In the current state, the runtime adds a `?` to every request URL even if the query was not set. This change fixes that behavior for AWS API Gateway V2 requests.


calavera (approved):
thanks!

Issue #945, opened 2024-11-30
Provide ability to define span.kind for OpenTelemetry Layer


alessandrobologna:
Hi,
I wanted to propose a small enhancement to the OpenTelemetry tracing layer implementation in the AWS Lambda Rust runtime.

Currently, in [otel.rs](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-runtime/src/layers/otel.rs#L75-L81), the tracing span for Lambda invocations doesn't include a way to specify the span kind (e.g., server, client). Adding this capability could improve how systems are represented in observability platforms.

My suggestion: we could add an `otel.kind` field to the span creation, initialized as empty. This would allow users to set the span kind in their handlers if needed. For example:

```
let span = tracing::info_span!(
    "Lambda function invocation",
    "otel.name" = req.context.env_config.function_name,
    "otel.kind" = field::Empty,  // New field for span kind
    { traceconv::FAAS_TRIGGER } = &self.otel_attribute_trigger,
    { traceconv::FAAS_INVOCATION_ID } = req.context.request_id,
    { traceconv::FAAS_COLDSTART } = self.coldstart
);
```

Users could then set the span kind in their handler like this:

```
use tracing::Span;

async fn handler(event: Value, _: Context) -> Result<Value, Error> {
    let span = Span::current();
    span.record("otel.kind", "server");
    
    // Handler logic...
}
```

This change *should* be backward compatible and provide more flexibility for users. I'd love to hear your thoughts on this, and happy to propose a PR!

maxday:
closing this as https://github.com/awslabs/aws-lambda-rust-runtime/pull/946 as been merged.

PR #946, opened 2024-12-01
Feature/otel span kind support

alessandrobologna:
📬 *Issue #945

✍️ *Add OpenTelemetry span kind support to Lambda invocation tracing. This enhancement allows users to specify the span kind (e.g., SERVER, CLIENT) in their Lambda handlers, improving how functions are represented in distributed traces.
In this way:
- Users can set the span kind in their handlers using `span.record("otel.kind", "SERVER")` or `span.record("otel.kind", "CONSUMER") etc.
- Allows to follow [OpenTelemetry semantic conventions for span kind](https://opentelemetry.io/docs/specs/otel/trace/api/#spankind) while maintaining backward compatibility (the field can be left empty)

**Changes:**
- Added `otel.kind` field to the Lambda invocation span, initialized as `field::Empty` as per the [tracing_opentelemetry docs](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/#special-fields)
- Updated the example in `examples/opentelemetry-tracing/main.rs`to demonstrate setting the span kind

Please note that since tracing_opentelemetry 0.28.0, [they have provided a method to set the status on a span](https://github.com/tokio-rs/tracing-opentelemetry/pull/176), therefore setting `otel.status_code` and `otel.status_message` in advance on the span is not required anymore.

### DIFF
--- a/examples/opentelemetry-tracing/src/main.rs
+++ b/examples/opentelemetry-tracing/src/main.rs
@@ -1,6 +1,7 @@
 use lambda_runtime::{
     layers::{OpenTelemetryFaasTrigger, OpenTelemetryLayer as OtelLayer},
-    LambdaEvent, Runtime, tracing::Span
+    tracing::Span,
+    LambdaEvent, Runtime,
 };
 use opentelemetry::trace::TracerProvider;
 use opentelemetry_sdk::{runtime, trace};
--- a/examples/opentelemetry-tracing/src/main.rs
+++ b/examples/opentelemetry-tracing/src/main.rs
@@ -1,6 +1,6 @@
 use lambda_runtime::{
     layers::{OpenTelemetryFaasTrigger, OpenTelemetryLayer as OtelLayer},
-    LambdaEvent, Runtime,
+    LambdaEvent, Runtime, tracing::Span
 };
 use opentelemetry::trace::TracerProvider;
 use opentelemetry_sdk::{runtime, trace};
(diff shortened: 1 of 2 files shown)


maxday:
Thanks for your contribution! 
What do you think about using the enum from `https://docs.rs/opentelemetry/0.27.0/opentelemetry/trace/enum.SpanKind.html` instead of passing an hardcoded string like "server"?

alessandrobologna:
@maxday that's really depending on how the tracing_opentelemetry crate handles it. As it is today, it handles string to SpanKind conversion internally: see https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L443 and  https://github.com/tokio-rs/tracing-opentelemetry/blob/v0.1.x/src/layer.rs#L102-L111.

maxday:
While testing it seems that the "kind" is passed as a integer (enum id I guess)
for instance 4 for`producer` `"parentSpanId":"0000000000000000","name":"helloooo","kind":4,"startTimeUnixNano":1733334175621925826`
It seems that's what really want, I think it would be better to see the full name of the kind?

alessandrobologna:
(oh, and the o11y platforms will typically, or at least all those that i tried, convert it back to a string for visualization)

maxday:
It seems that our CI is red, I've just pushed a PR to fix it, I'll let you know once it gets reviewed/merge so you can rebase from that. Thanks!

alessandrobologna:
Yes, I saw, no worries

maxday:
https://github.com/awslabs/aws-lambda-rust-runtime/pull/948 has just been merged and CI is now green, could you please merge main into your branch, this should resolve the failing checks? Thanks!

maxday (approved):
LGTM! Thanks 👍

PR #980, opened 2025-05-02
fix(docs): examples/extension-internal-flush: don't refer to a standalone layer arn for internal extension

jlizen:
📬 *Issue #, if available:*
n/a

✍️ *Description of changes:*

Small tweak to the docs for our example on internal extensions. We currently refer to retrieving a layer arn from `cargo lambda deploy`. This was probably copy-pasted from other examples where we actually are calling `cargo lambda deploy` with the `--extension` target.

In this case, we are deploying the function directly, and there is no standalone extension - it is all inside our primary binary.

Tweaking the language to remove a reference to a deployed layer ARN.

If we want to add a guide to invoking the remote function deployment, we could. The other examples don't have that though.

- [x ] I confirm that I've ran `cargo +nightly fmt`.
- [ x] I confirm that I've ran `cargo clippy --fix`.

### DIFF
--- a/examples/extension-internal-flush/README.md
+++ b/examples/extension-internal-flush/README.md
@@ -18,9 +18,10 @@ the Lambda service returns the response to the caller immediately. Extensions ma
 without introducing an observable delay.
 
 ## Build & Deploy
+Two deploy options for internal extensions: 
 
 1. Install [cargo-lambda](https://github.com/cargo-lambda/cargo-lambda#installation)
-2. Build the extension with `cargo lambda build --release`
+2. Build a function with the internal extension embedded with `cargo lambda build --release`
 3. Deploy the function to AWS Lambda with `cargo lambda deploy --iam-role YOUR_ROLE`
 
 The last command will give you an ARN for the extension layer that you can use in your functions.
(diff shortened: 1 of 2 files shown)


jlizen:
Actually, this isn't working properly in my local testing. I think the original command was wrong. I'll open a new one later.

jlizen:
Reopening, seems to be a bug in `cargo lambda`: https://github.com/cargo-lambda/cargo-lambda/issues/848

This was actually called out when this example was first added:
https://github.com/awslabs/aws-lambda-rust-runtime/pull/744#issuecomment-1837286161

Will see if I have time to fix that as well.

jlizen:
Have a fix up for the `cargo lambda watch` bug: https://github.com/cargo-lambda/cargo-lambda/pull/849

This example does successfully run locally with that fix in place.

jlizen:
As another side note, I'd like to update the example to show a more realistic per-event flush via an extension using a non-blocking writer, if I implement this feature in `tracing-appender`:
- https://github.com/tokio-rs/tracing/issues/3267

(that's assuming this change lands: https://github.com/awslabs/aws-lambda-rust-runtime/pull/982)

PR #981, opened 2025-05-04
Add alternative to pip3 install of cargo lambda

Bryson14:
Another way to manage executable pip packages is with uv which has been very popular and helpful in the python community.

```

📬 *Issue #, if available:*

Small issue, When running `pip3 install .....` it shows this error about system vs user installation:

```
rror: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

```

✍️ *Description of changes:*

Experienced python users might already know that they can use `uv` to install and load python packages from pip as executable. But for those that don't know, adding this line can be a helpful alternative that doesn't show installation errors, reducing any friction to rust or python developers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,11 @@ Or PiP on any system with Python 3 installed:
 ```bash
 pip3 install cargo-lambda
 ```
+Alternative, install the pip package as an executable using [uv](https://docs.astral.sh/uv/)
+
+```bash
+uv tool install cargo-lambda
+```
 
 See other installation options in [the Cargo Lambda documentation](https://www.cargo-lambda.info/guide/installation.html).
 

### CODE REVIEW

**Comment by calavera** [comment]
Uv is great, but it’s not as standard as pip, so at least, you’d have to explain how to install Uv first. The paragraph below already has a link with additional installation methods, it might be better to pick from that list:

https://www.cargo-lambda.info/guide/installation.html

Installation scripts is probably what people are more used to which can be found here:

```
curl -fsSL https://cargo-lambda.info/install.sh | sh
```

For Windows/Powershell specifically:

```
irm https://cargo-lambda.info/install.ps1 | iex
```

jlizen (approved):
Hi @Bryson14 , the lambda runtime maintainers circled back to this, and we agree that `uv` is a relevant tool to share with customers. I see that you linked to its docs, which should be sufficient discoverability of installation guidance.

jlizen:
Thanks for the contribution!

PR #1005, opened 2025-06-30
fix(ci): cut run-integration-test.yml over to "mlugg/setup-zig@v2", clippy fixes

jlizen:
✍️ *Description of changes:*

#### Run-integ setup-zig action change

We had pinned back our `cargo-lambda` version due to CI failures. Ref: https://github.com/awslabs/aws-lambda-rust-runtime/pull/989

The `cargo-lambda` issue was closed because it was determined that the issue had to do with the underlying zig install. essentially, you need libraries available, not just the binary. Ref: https://github.com/cargo-lambda/cargo-lambda/issues/856#issuecomment-2985326373

In the process of cleaning up, I noticed we were using an archived action. Cutting us over to the modern one.

Unfortunately, this wasn't sufficient to fix the issue, [sample run](https://github.com/awslabs/aws-lambda-rust-runtime/actions/runs/15980135053/job/45072371532), so leaving us pinned for now. I cut a backlog: https://github.com/awslabs/aws-lambda-rust-runtime/issues/1006

#### Clippy fixes
The new clippy turned inline variable enforcement in fmt strings from `pedantic` to `warn`, so our CI started failing. Fixing.

### DIFF
--- a/.github/workflows/run-integration-test.yml
+++ b/.github/workflows/run-integration-test.yml
@@ -18,7 +18,7 @@ jobs:
             repo: cargo-lambda/cargo-lambda
             platform: linux
             arch: x86_64
-            # TODO: unpin once https://github.com/cargo-lambda/cargo-lambda/issues/856 is fixed
+            # TODO: unpin once https://github.com/awslabs/aws-lambda-rust-runtime/issues/1006 is fixed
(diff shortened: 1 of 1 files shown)


bnusunny (approved):
LGTM

PR #1007, opened 2025-07-03
chore(lambda-runtime): slightly optimize graceful shutdown helper by using new tokio::try_join biased; api

jlizen:
✍️ *Description of changes:*
Closing out some minor to-dos. My [tokio PR adding a biased; mode to join/try_join](https://github.com/tokio-rs/tokio/pull/7307) landed. This allows us to skip the bookkeeping of keeping track of which branch of a `join`/`try_join!` statement was polled last, and always just poll in order.

This is a minor optimization for our graceful shutdown handler, since we always want to just poll our graceful shutdown future first before our non-op extension. So we can save a bit of memory on the stack by skipping storing a usize. It also very marginally could impact latency (by micros if not nanos, it's insignificant probably).

Also applying a similar optimization to the internal extension example where we have it the extension wait until after the main handler completes so that it can flush telemetry.

This involved explicitly setting `tokio` to `1.46`. This won't impact our MSRV as our MSRV is set to 1.81 but tokio is 1.70.

### DIFF
--- a/examples/extension-internal-flush/Cargo.toml
+++ b/examples/extension-internal-flush/Cargo.toml
@@ -9,4 +9,4 @@ aws_lambda_events = { path = "../../lambda-events" }
 lambda-extension = { path = "../../lambda-extension" }
 lambda_runtime = { path = "../../lambda-runtime" }
 serde = "1.0.136"
-tokio = { version = "1", features = ["macros", "sync"] }
+tokio = { version = "1.46", features = ["macros", "sync"] }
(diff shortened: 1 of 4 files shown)


jlizen:
Clippy check is failing due to unlined format args changing to warn level.

But I have this fixed in: https://github.com/awslabs/aws-lambda-rust-runtime/pull/1005

Probably fine to ignore in this CR in favor of that fix.

PR #1008, opened 2025-07-10
fix(lambda-events): derive Default on KinesisEvent

sdstolworthy:
📬 *Issue #, if available:*

✍️ *Description of changes:*

Adds a default implementation for KinesisEvent.

When using `cargo lambda init`, the following test is automatically generated:
```rust
#[cfg(test)]
mod tests {
    use super::*;
    use lambda_runtime::{Context, LambdaEvent};

    #[tokio::test]
    async fn test_event_handler() {
        let event = LambdaEvent::new(KinesisEvent::default(), Context::default());
        let response = function_handler(event).await.unwrap();
        assert_eq!((), response);
    }
}
```

This test fails because `Default` is not implemented on `KinesisEvent`.

This change provides a basic implementation of default by adding `Default` to the existing `#[derive(...)]` attribute.

### DIFF
--- a/lambda-events/src/event/kinesis/event.rs
+++ b/lambda-events/src/event/kinesis/event.rs
@@ -4,7 +4,7 @@ use crate::{
 };
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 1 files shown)


jlizen (approved):
Thanks for fixing this up!

I also cut an issue to double check the remaining events, I'll try to get to it when I have cycles if nobody else does first. https://github.com/awslabs/aws-lambda-rust-runtime/issues/1009

Issue #1009, opened 2025-07-10
[lambda-events] Make sure that we have Default implemented for all event types


jlizen:
In https://github.com/awslabs/aws-lambda-rust-runtime/pull/1008, a contributor noticed that our kinesis event was missing its `#[derive(Default)]` and implemented that.

We should double check that no other events are missing Default impl as well. It's irritating to write unit tests without it.

sdstolworthy:
To provide additional context here, `cargo lambda init` allows a user to select the type of event that will invoke their Lambda. All of the events in `lambda_events` are options. After completing the `init` flow, code is autogenerated in `event_handler.rs` that is specific to the event selected.

For example, for the KinesisEvent, the following test is generated in `event_handler.rs`:
```rust
#[cfg(test)]
mod tests {
    use super::*;
    use lambda_runtime::{Context, LambdaEvent};

    #[tokio::test]
    async fn test_event_handler() {
        let event = LambdaEvent::new(KinesisEvent::default(), Context::default());
        let response = function_handler(event).await.unwrap();
        assert_eq!((), response);
    }
}
```

Prior to https://github.com/awslabs/aws-lambda-rust-runtime/pull/1008, this code failed to compile, because `Default` was not implemented on `KinesisEvent`.

A similar issue has been opened previously: https://github.com/awslabs/aws-lambda-rust-runtime/issues/725.

pmd3d:
I'd really like to work on this. I might need a couple weeks in my spare time though. Hope that is ok.

jlizen:
@pmd3d thanks for cutting #1022 !

Leaving this issue open as I see that there are still further events that need handling.

[random example](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-events/src/event/clientvpn/mod.rs#L7)

ghost:
Marking as completed.

PR #1011, opened 2025-07-10
fix(lambda-events): impl Default for all Kinesis Event structs

sdstolworthy:
📬 *Issue #, if available:*
I chatted with @jlizen after she opened https://github.com/awslabs/aws-lambda-rust-runtime/issues/1009.

For testing ergonomics, it makes sense to implement `Default` on all of the structs in Kinesis Events. This PR is a follow up to https://github.com/awslabs/aws-lambda-rust-runtime/pull/1008, and implements `Default` for all the KinesisEvent-related structs.
  
✍️ *Description of changes:*
Implement `Default` on all `KinesisEvent`-related structs.

I also changed the test that I merged in in the last PR. At first glance, this test feels unnecessary because it is not actually making any assertions. However, because `cargo lambda init` generates code that inherently relies on `Default` being implemented on these structs, I think a unit test that attempts to exercise that code path is worthwhile. Theoretically, the test can never fail at runtime, only compile time.

This test will protect against breaking `cargo lambda init`'s autogenerated code in the absence of E2E tests.

### DIFF
--- a/lambda-events/src/encodings/time.rs
+++ b/lambda-events/src/encodings/time.rs
@@ -29,7 +29,7 @@ impl DerefMut for MillisecondTimestamp {
 }
 
 /// Timestamp with second precision.
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
+#[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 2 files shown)


jlizen (approved):
Thanks!

Issue #967, opened 2025-03-12
run_with_streaming_response and automatically streaming buffered responses


keichan34:
I know that right now, buffered responses will return an empty response when running with `run_with_streaming_response` (#805). Would it be possible (would PRs be accepted?) for some code that would intercept buffered responses and transform them to a dummy stream?

I'm currently doing this in my app because I'm sharing some error handling code between the buffered and streaming APIs, but I think it would make sense for something like this to live in the runtime. What do you think?

bnusunny:
I'm not quite understand your use case. Do you mind to share a sample code?

keichan34:
It's basically exactly what's in #805. I've broken up my APIs in to two Lambda functions -- one for buffered and one for streaming, and I want to reuse buffered handlers between the streaming and buffered APIs.

bnusunny:
Got it! It will be useful to able to use response streaming model to process both buffered and streaming functions. PR is welcome.

fangkangmi:
I investigated this to prepare a PR and wanted to share my findings.

## Reproduction attempt (March 2026)

I deployed a minimal Lambda using `run_with_streaming_response` with buffered handlers (`Body::Text`) on a Function URL configured with `RESPONSE_STREAM` invoke mode, using the current `main` branch.

```rust
async fn handler(event: Request) -> Result<Response<Body>, Error> {
    let path = event.uri().path();
    match path {
        "/buffered" => Ok(Response::builder()
            .status(StatusCode::OK)
            .header("content-type", "text/plain")
            .body(Body::Text("Hello from buffered handler!".to_string()))
            .unwrap()),
        _ => Ok(Response::builder()
            .status(StatusCode::OK)
            .body(Body::Text(format!("You requested: {}", path)))
            .unwrap()),
    }
}

#[tokio::main]
async fn main() -> Result<(), Error> {
    run_with_streaming_response(service_fn(handler)).await
}
```

## Result

Both routes returned correct, non-empty responses through the streaming Function URL:

```
$ curl https://<function-url>/buffered
Hello from buffered handler!    # 28 bytes, HTTP 200

$ curl https://<function-url>/hello
You requested: /hello           # 21 bytes, HTTP 200
```

The empty response problem from #805 no longer reproduces. It appears the Lambda platform has resolved the underlying issue since the original report (Feb 2024).

## What I verified in the code

[...]

Issue #805, opened 2024-02-01
Mixing buffered and streaming responses


GabrielBianconi:
I'm trying to build a Lambda with routes for both streaming (Axum SSE) and regular (buffered) handlers.

I've tried using `run_with_streaming_response`, which works as expected for the streaming handlers but sends an empty response for regular handlers.

With `run`, the regular handler works as expected but it also buffers the streaming handler before responding.

Here's a minimal example:

```rs
use axum::{
    response::{
        sse::{Event, Sse},
        IntoResponse,
    },
    routing::get,
    Router,
};
use futures::stream::{self, Stream};
use std::{convert::Infallible, time::Duration};
use tokio_stream::StreamExt as _;

async fn regular_handler() -> impl IntoResponse {
    "Buffered hello world!"
}

async fn sse_handler() -> Sse<impl Stream<Item = Result<Event, Infallible>>> {
    let stream = stream::repeat_with(|| Event::default().data("Streamed hello world!"))
        .map(Ok)
        .throttle(Duration::from_secs(1))
        .take(5);

    Sse::new(stream).keep_alive(
        axum::response::sse::KeepAlive::new()
            .interval(Duration::from_secs(1))
            .text("keep-alive-text"),
    )
}

#[tokio::main]
async fn main() -> Result<(), lambda_http::Error> {
    let app = Router::new()
        .route("/a", get(regular_handler))
        .route("/b", get(sse_handler));

    // lambda_http::run(app).await // --> /a works fine, but also buffers /b before responding

    lambda_http::run_with_streaming_response(app).await // --> /a returns an empty reply from server, /b streams as expected
}
```

Is there a way to set up a Lambda that accepts both types of handlers?

Thank you.

bnusunny:
Lambda response streaming is decided by the invoke mode of function URLs. When you configure a function URL to response streaming, all the invokes are streamed. It is not possible to switch between buffered and streamed responses per invoke.

GabrielBianconi:
Makes sense, thank you.

PR #955, opened 2025-01-03
fix: serialize `AlbTargetGroupRequest::query_string_parameters` value to string

DiscreteTom:
📬 *Issue #, if available:* 

#954

✍️ *Description of changes:* 

Provide a custom serializer `serialize_query_string_parameters` to serialize `AlbTargetGroupRequest::query_string_parameters`.

🔏 *By submitting this pull request*

- [x] I confirm that I've ran `cargo +nightly fmt`.
- [x] I confirm that I've ran `cargo clippy --fix`.
- [x] I confirm that I've made a best effort attempt to update all relevant documentation.
- [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1013, opened 2025-07-11
Expose streaming API

s0l0ist:
📬 *Issue #, if available:*

The SDK supports streaming, but `lambda_http::Adapter` only handles buffered responses from Axum's `Router` (i.e., a `Service`).  The `lambda_http::run_with_streaming_response` is available, but it doesn't allow you to specify a custom runtime which is necessary if you want to support OpenTelemetry.

Related (ish):
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/967
- https://github.com/awslabs/aws-lambda-rust-runtime/issues/805

✍️ *Description of changes:*

This PR exposes:
- `StreamAdapter`: Converts a tower `Service` into an AWS Lambda streaming response.

I also have added two new examples which showcase how to use Axum with `run_with_streaming_response` and with `StreamAdapter` + OTeL

This is how you can use it with a custom runtime supporting OTeL:
```rust
use crate::{error::BoxError, service::build_service};
use axum::Router;
use lambda_runtime::{
    Runtime,
    layers::{OpenTelemetryFaasTrigger, OpenTelemetryLayer},
};
use opentelemetry_sdk::trace as sdktrace;

/// The main function to run the AWS Lambda
pub async fn async_main(tracer_provider: sdktrace::SdkTracerProvider) -> Result<(), BoxError> {
    let (app, _): (Router, _) = build_service().await;

    // For a buffered response:
    // let handler = lambda_http::Adapter::from(app);
    // For a streamed response:
    let handler = lambda_http::StreamAdapter::from(app);

    let runtime = Runtime::new(handler).layer(
        // Create a tracing span for each Lambda invocation
        OpenTelemetryLayer::new(|| {
            if let Err(err) = tracer_provider.force_flush() {
                eprintln!("Error flushing traces: {:#?}", err);
            }
        })
        .with_trigger(OpenTelemetryFaasTrigger::Http),
    );
    runtime.run().await
}
```


jlizen:
Hi @s0l0ist , thanks for cutting this PR. This change seems very reasonable to me.

I do have concerns about leaking internals. This is only naming return types that are already public, but we might want to tweak their composition or otherwise shift bounds in ways that break callers. I think it would be best to type erase the returned service stack.

The easiest way to do that would be via a [tower::util::BoxService](https://docs.rs/tower/latest/tower/util/struct.BoxService.html) / [BoxLayer](https://tower-rs.github.io/tower/tower/util/struct.BoxLayer.html) (or the sync/clone-bounded variants, though I don't think we need them here - but we could certainly add alternate APIs that do include the sync/clone bounds if it would be useful for how the returned struct is used by a runtime). 

That adds a small amount of performance overhead due to an extra allocation and layer of dynamic dispatch. But, I think the ergonomics would be much better compared to a more complex, composable builder-style API using generics and sealed inner layer types. Note that we would probably want some sort of `into_streaming_streaming_response_inner()` that is what `into_streaming_response()` is currently, so that our `run_with_streaming_response` can skip the minor type erasure overhead.

How does that sound to you?

It would also be great to get a small example showing usage of this with a non-tokio runtime, if you'd be open to it! That would both let us validate the API, and make it more discoverable for users. I'd probably be ok with this landing without that, though, if you don't have cycles.

lambda-http/src/streaming.rs
```diff
@@ -54,8 +70,25 @@ where
                 metadata_prelude,
                 stream: BodyStream { body },
             }
-        });
+        })
+}
 
+/// Starts the Lambda Rust runtime and stream response back [Configure Lambda
+/// Streaming
```
jlizen: This formatting could use cleaning up also, some extra newlines.

s0l0ist:
@jlizen, I think that's the right direction. Currently busy right now, but can revisit this PR in a couple of weeks from now.

s0l0ist:
@jlizen - does this API work? I'll add an example or two if I have time, but wanted to make sure this was aligned beforehand.

lambda-http/src/streaming.rs
```diff
@@ -2,21 +2,45 @@ use crate::{http::header::SET_COOKIE, request::LambdaRequest, tower::ServiceBuil
...
-/// This takes care of transforming the LambdaEvent into a [`Request`] and
-/// accepts [`http::Response<http_body::Body>`] as response.
-pub async fn run_with_streaming_response<'a, S, B, E>(handler: S) -> Result<(), Error>
+/// Runs the Lambda runtime with a handler that returns **streaming** HTTP
+/// responses.
+pub fn into_streaming_response<'a, S, B, E>(
+    handler: S,
+) -> BoxService<LambdaEvent<LambdaRequest>, StreamResponse<BodyStream<B>>, E>
+where
+    S: Service<Request, Response = Response<B>, Error = E> + Send + 'static,
```
jlizen: That's a bit awkward that we have the new `Send + 'static` bounds on the service since we don't really need them... I see that is required for [boxed](https://docs.rs/tower/latest/tower/trait.ServiceExt.html#method.boxed) and `BoxService`. That is unfortunate.

But, the machinery to spin our own is [quite complicated](https://docs.rs/tower/latest/src/tower/util/boxed/sync.rs.), i'm not sure it's worth it unless we have use cases for it.

I guess one thing I would suggest would be, we could look at newtyping the `BoxService` that this responds with so that we have the option to swap it out with something with less restrictive bounds down the road?

Today it would be a breaking change to rip out `BoxService` for a handrolled impl that didn't require a `Send + 'static` Service.

jlizen: Ie something like:
```
pub struct TypeErasedService<T, U, E> {
    inner: BoxService<T, U, E>
}

impl <T, U, E> Service<T> for TypeErasedService<T, U, E> {
     type Response = U;

    type Error = E;

    type Future = BoxFuture<U, E>;

    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), E>> {
        self.inner.get_mut().poll_ready(cx)
    }


    fn call(&mut self, request: T) -> BoxFuture<U, E> {
        self.inner.get_mut().call(request)
    }
}
impl<T, U, E> fmt::Debug for TypeErasedService<T, U, E> {
    fn fmt(&self, fmt: &mut fmt::Formatter) -> fmt::Result {
        fmt.debug_struct("TypeErasedService").finish()
    }
}
```

That way we could today require the `Send + 'static` input bound, which will allow us to construct the inner `BoxService`, but we could later looe that restriction and use a different inner.

jlizen (commented):
General approach looks good, but left a note around newtyping the `BoxService` so we don't lock ourselves into extra `Service: Send + 'static` bounds for all time. (For now is fine though, just want to leave the door open to removing them).

I'm seeing that we don't have ANY tests currently of the `run_with_streaming_response` path... that is unfortunate... An example would be really great if you have the time (which would double as a test).

lambda-http/src/streaming.rs
s0l0ist: Agreed about `'static` and I like the compromise. Thanks for the suggestion. I'll iterate and circle back.

s0l0ist:
Hi @jlizen, I think I may have a better approach.

I noticed that [Adapter](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-http/src/lib.rs#L147), handles buffered responses by wrapping a `Service<Request>` and producing a `LambdaResponse` from anything that implements `IntoResponse`. Taking inspiration from that, I built a `StreamAdapter` that wraps a `Service<Request, Response = Response<B>>` and produces a `StreamResponse<BodyStream<B>>`.

This lets us remove the `BoxService` entirely while keeping the same ergonomics. The trade-off is a heap-allocated future per request (same as `Adapter`), but we avoid exposing `BoxService` in the public API.

With this change, `run_with_streaming_response` just mirrors the non-streaming [run](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-http/src/lib.rs#L194):

```rust
pub async fn run_with_streaming_response<'a, S, B, E>(handler: S) -> Result<(), Error>
where
    S: Service<Request, Response = Response<B>, Error = E>,
    S::Future: Send + 'a,
    B: Body + Unpin + Send + 'static,
    B::Data: Into<Bytes> + Send,
    B::Error: Into<Error> + Send + Debug,
{
    lambda_runtime::run(StreamAdapter::from(handler)).await
}
```

And I can selectively choose which one I want like this:

```rust
/// The main function to run the AWS Lambda
pub async fn async_main(tracer_provider: sdktrace::SdkTracerProvider) -> Result<(), BoxError> {
    let (app, _) = build_service().await;

    // For a buffered response:
    // let handler = lambda_http::Adapter::from(app);
    // For a streamed response:
    let handler = lambda_http::StreamAdapter::from(app);

    let runtime = Runtime::new(handler).layer(
        // Create a tracing span for each Lambda invocation
        OpenTelemetryLayer::new(|| {
            if let Err(err) = tracer_provider.force_flush() {
                eprintln!("Error flushing traces: {err:#?}");
            }
        })
        .with_trigger(OpenTelemetryFaasTrigger::Http),
    );
    runtime.run().await
}
```

lambda-http/src/streaming.rs
```diff
@@ -1,22 +1,27 @@
...
-/// Starts the Lambda Rust runtime and stream response back [Configure Lambda
-/// Streaming Response](https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html).
-///
-/// This takes care of transforming the LambdaEvent into a [`Request`] and
-/// accepts [`http::Response<http_body::Body>`] as response.
-pub async fn run_with_streaming_response<'a, S, B, E>(handler: S) -> Result<(), Error>
+/// An adapter that lifts a standard [`Service<Request>`] into a
+/// [`Service<LambdaEvent<LambdaRequest>>`] which produces streaming Lambda HTTP
+/// responses.
+pub struct StreamAdapter<'a, S, B> {
```
jlizen, line 24: This is definitely a cleaner API, I like it! Good call! Anyway we are heap-allocating as part of `BoxService` so not really giving up anything with this approach.

jlizen (changes requested):
I like the direction you went with this for the new API. But, I would prefer to have somewhat duplicated code if that is what is needed to keep from adding a new allocation to the existing API.

Presumably a decent chunk of it could be pulled out into a pure function to avoid too much duplication.

s0l0ist:
Okay I think I have it:

- Add` into_stream_service(...)`: zero-alloc builder (no boxed future/vtable) that is used by `run_with_streaming_response` so that it continues to have the same exact behavior.
- Keep `StreamAdapter` for ergonomic stacking; it still returns a boxed future for API stability.
- Factor common logic into `into_stream_response(...)` which is used by both `StreamAdapter` and `into_stream_service`

I added a parallel test to what `Adapter` already had - ensuring we can compile-check a `BoxService` built the same way. Once we’re happy with the shape, I’ll follow up more examples. I've already added the `http-axum-streaming` example.

examples/http-axum-streaming/src/main.rs
```diff
@@ -0,0 +1,62 @@
+use axum::{
```
jlizen, line 11: Wonderful addition, thanks for this!

lambda-http/src/streaming.rs
```diff
@@ -85,3 +172,56 @@ where
...
+
+            self.inner.call(event)
+        }
+    }
+
+    /// This tests that `StreamAdapter` can be used in a `tower::Service` where
+    /// the user may require additional middleware between `lambda_runtime::run`
+    /// and where the `LambdaEvent` is converted into a `Request`.
+    #[test]
+    fn stream_adapter_is_boxable() {
```
jlizen, line 207: Minor: Since this isn't making any assertions anyway, and is mostly testing that it compiles, would this perhaps be better suited as an example? Seems like a nice simple example of setting up custom middleware? Also would help capture any weirdness with private vs public API.

Up to you, I'm happy to have it here too.

s0l0ist: Yeah can do! I think I was mostly just trying to mimic the same test for the Adapter. I can move this out into the example which would effectively test this.

s0l0ist:
Yes I see the CI checks failing - they're unrelated to this PR, but happy to clean them up.

jlizen:
Ah, ok, I didn't actually look at them - probably we got hit by some new clippy/rustfmt lints across toolchain versions. If you have time to clean them up, feel free / much appreciated (separate commit please).

It sounded like you did plan to hoist that test into an example, so I can sit tight on merging since it sounds like a new commit is coming. But just ping me if you want to ship it without another round of changes.

s0l0ist:
@jlizen , fixed CI [here](https://github.com/awslabs/aws-lambda-rust-runtime/pull/1030).

And yes, I plan on moving the test into the new example and that should wrap this PR up.

s0l0ist:
@jlizen - Okay! I think we're ready :) 

Added two examples, one that showcases `run_with_streaming_response` and another that demonstrates configuring a custom runtime with OTeL using `StreamAdapter`. Both examples type-erase the service into a `BoxService` before converting it into a streaming service. I didn’t have a good example of type-erasing after conversion, so I kept the compile test since it directly validates `Tower::Service`.

I've tested manually e2e to verify everything works.

jlizen:
Sorry to miss this today @s0l0ist - near the top of my list for tomorrow. Thanks for the CI fix.

examples/http-axum-streaming/src/main.rs
```diff
@@ -0,0 +1,100 @@
...
+    });
+
+    Ok(Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, "text/plain; charset=utf-8")
+        .header(CACHE_CONTROL, "no-cache")
+        .body(body)?)
+}
+
+fn create_svc() -> BoxService<Request<lambda_http::Body>, Response<Body>, Infallible> {
```
jlizen: Same concern about the `BoxService`, it's vestigial.

examples/http-axum-streaming-otel/src/main.rs
```diff
@@ -0,0 +1,145 @@
...
+    );
+
+    Ok(Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, "text/plain; charset=utf-8")
+        .header(CACHE_CONTROL, "no-cache")
+        .body(body)?)
+}
+
+fn create_svc() -> BoxService<Request<lambda_http::Body>, Response<Body>, Infallible> {
```
jlizen: This use of `BoxService` seems strange. Anyway `Router` is backed by a `BoxService` under the hood. It internally uses that to flatten down into a nameable type even with arbitrary layers.

It could be trivially rewritten to the following - I included use of the layer to demonstrate the type erasure that `Router` is doing, but we could leave it out of the real example:
```
fn create_svc() -> Router<Request<lambda_http::Body>> {
    if std::env::var("USE_NUMBERS").as_deref() == Ok("1") {
        Router::new()
            .route("/", get(stream_numbers))
            .layer(tower_http::timeout::TimeoutLayer::new(Duration::from_secs(60)))
    } else {
        Router::new().route("/", get(stream_words))
    }
}
```

I'm not sure including `BoxService` buys us very much in the example... It just adds overhead mostly and confuses readers.

Can we just ditch it in favor of the simpler formulation?

lambda-http/src/streaming.rs
jlizen, line 207: Thanks for leaving this in - the new examples don't probe that StreamAdapter itself is dyn safe.

lambda-http/src/streaming.rs
```diff
@@ -85,3 +172,56 @@ where
...
+        fn call(&mut self, event: LambdaEvent<LambdaRequest>) -> Self::Future {
+            // Log the request
+            println!("Lambda event: {event:#?}");
+
+            self.inner.call(event)
+        }
+    }
+
+    /// This tests that `StreamAdapter` can be used in a `tower::Service` where
+    /// the user may require additional middleware between `lambda_runtime::run`
```
jlizen: Nit: Dyn safety isn't REALLY necessary for middleware, it's more just a convenience for composing complicated stacks that you might want to decompose across functions and have conditional composition, but still name explicitly. I know this is existing verbiage from the other example, I'm fine to leave it, truly is just a nit ^^

jlizen:
Sorry for the back and forth on examples @s0l0ist - I'm good to merge with that one requested change (remove the vestigial `BoxService`). Given that users will probably copy-paste these examples, it would be better to show the right general practice.

examples/http-axum-streaming-otel/src/main.rs
s0l0ist: Yeah I was struggling a bit to do something clever. It's trying to demo too many things in any case that isn't necessary. Agreed on just building a simple router. I'll refactor.

s0l0ist:
> Sorry for the back and forth on examples @s0l0ist - I'm good to merge with that one requested change (remove the vestigial `BoxService`). Given that users will probably copy-paste these examples, it would be better to show the right general practice.

No need to apologize! I'm happy to iterate and collaborate on the best approach moving forward.

jlizen (approved):

examples/http-axum-streaming-otel/src/main.rs
jlizen: I do like the idea of showing how to compose interesting tower stacks, for what it's worth... for instance an example that shows how to have a middleware stack that can dump the raw lambda event before conversion into a `Request` might be nice.

jlizen:
Thanks for this contribution @s0l0ist ! Great stuff! I'll work on getting a release out though might not land until early next week, I would like to get https://github.com/awslabs/aws-lambda-rust-runtime/pull/955 reviewed as well so that they can go out together.

Issue #1016, opened 2025-07-11
[aws_lambda_events] Consider tagging stucts/enums with `#[non_exhaustive]`


jlizen:
`aws_lambda_events` is almost entirely composed of structs and enums with public inner fields. This creates a constant semver hazards as we can't add new fields without breaking semver. Meanwhile, we are mapping external schemas that do frequently add fields.

For now, we just bump semver whenever we add events, but that is pretty painful for consumers to always bump, and otherwise they might be silently ignoring payload fields that `serde_json` doesn't recognize. It would be nicer for them to be able to eg dump the full event input via their debug logs and always see the latest fields, even without handling them explicitly in their application code. 

For enums, this seems like a no-brainer. For structs, it's a bit awkward since we'll probably want to implement `Default` so that unit testing isn't a pain. But, most structs are comprised of `Option<T>`, `Vec<T>`, and `HashMap<T, T>` fields anyway, which makes `#[derive(Default)]` pretty easy. String fields probably are fine to just default to `""`. 

Anyway, it's a lot of gruntwork, but probably a good idea to do before an eventual `1.0` release at least? Might be good LLM fodder ^^

jlizen:
Ah, forgot that I had just opened https://github.com/awslabs/aws-lambda-rust-runtime/issues/1009  about adding `Default` impl to all fields, so this would build on top of that quite nicely.

jlizen:
there are also types in `lambda_http` that need this

ghost:
I can take care of this ^_^

ghost:
Marking as completed with latest PR.

PR #1018, opened 2025-07-16
add catchall feature to resolve #1017

### LINKED ISSUE #1017: [aws_lambda_events] Add 'catch-all' field for unmodeled fields, probably behind feature flag
`aws_lambda_events` is essentially a set of shadow schemas that are hand-maintained to match the event request/response types used by other services. This is very useful from a Rust perspective due to all the nice compile-time guarantees and performance optimizations we get using `serde`. However, it is very much a moving target, where schemas are always adding new fields, not to mention frequently including undocumented fields.

It would be nice to have a usage mode for 'I want the bleeding edge, including unmodeled fields'. That way, consumers could both interact with undocumented fields that we will probably never model, as well as the latest and greatest new fields even before we release a new version.

We could do that today via a 'catch-all' serde-modeled field that just is represented as a `Map<String, Value>`, as demonstrated here: https://github.com/serde-rs/serde/issues/941#issuecomment-375486935

```
#[derive(Serialize, Deserialize)]
struct S {
    a: u32,
    b: String,
    #[serde(flatten)]
    other: Map<String, Value>,
}
```

And then the caller can interact with contained fields at runtime, or just dump it to logs.

I think probably it belongs under a something like `feature = "catch-all-fields"` since this will involve extra allocations and other overhead?

It would be a fair bit of gruntwork to implement everywhere, but probably a LLM could do it pretty easily.

cdisselkoen:
📬 *Issue #, if available:*

#1017

✍️ *Description of changes:*

Adds a catchall `other` field to most structs in `aws-lambda-events` that implement `Deserialize`, gated behind a new feature `catch-all-fields`.  The field is documented everywhere it is added.

Adds one test for the new functionality, testing it in both a parent and child struct.

I was assisted by Amazon Q with this PR, so it's possible that an `other` field was added in places it doesn't make sense, or missed in places where it should have been added for consistency.  Nonetheless, I did my best to sanity-check, ensure that formatting was consistent, etc.  I also resolved a few test failures resulting from having a struct with a catchall `serde(flatten)`-ed into another struct with a catchall.  Hopefully there are no instances of this pattern that I missed.

### DIFF
--- a/lambda-events/src/event/activemq/mod.rs
+++ b/lambda-events/src/event/activemq/mod.rs
@@ -17,6 +17,7 @@ pub struct ActiveMqEvent {
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "catch-all-fields")))]
     #[serde(flatten)]
     pub other: serde_json::Map<String, Value>,
 }
@@ -53,6 +54,7 @@ pub struct ActiveMqMessage {
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "catch-all-fields")))]
     #[serde(flatten)]
     pub other: serde_json::Map<String, Value>,
 }
@@ -66,6 +68,7 @@ pub struct ActiveMqDestination {
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "catch-all-fields")))]
     #[serde(flatten)]
     pub other: serde_json::Map<String, Value>,
 }
--- a/lambda-events/src/event/activemq/mod.rs
+++ b/lambda-events/src/event/activemq/mod.rs
@@ -13,7 +13,7 @@ pub struct ActiveMqEvent {
     #[serde(default)]
     pub event_source_arn: Option<String>,
     pub messages: Vec<ActiveMqMessage>,
-    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Catchall to catch any additional fields that were present but not explicitly defined by this struct.
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
@@ -50,7 +50,7 @@ pub struct ActiveMqMessage {
     #[serde(deserialize_with = "deserialize_lambda_map")]
     #[serde(default)]
     pub properties: HashMap<String, String>,
-    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Catchall to catch any additional fields that were present but not explicitly defined by this struct.
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
@@ -64,7 +64,7 @@ pub struct ActiveMqMessage {
 pub struct ActiveMqDestination {
     #[serde(default)]
     pub physical_name: Option<String>,
-    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Catchall to catch any additional fields that were present but not explicitly defined by this struct.
     /// Enabled with Cargo feature `catch-all-fields`.
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
--- a/lambda-events/src/event/activemq/mod.rs
+++ b/lambda-events/src/event/activemq/mod.rs
@@ -18,7 +18,7 @@ pub struct ActiveMqEvent {
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
     #[serde(flatten)]
-    pub other: HashMap<String, Value>,
+    pub other: serde_json::Map<String, Value>,
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
@@ -54,7 +54,7 @@ pub struct ActiveMqMessage {
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
     #[serde(flatten)]
-    pub other: HashMap<String, Value>,
+    pub other: serde_json::Map<String, Value>,
 }
 
 #[derive(Debug, Clone, Eq, PartialEq, Deserialize, Serialize)]
@@ -67,7 +67,7 @@ pub struct ActiveMqDestination {
     /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
     #[cfg(feature = "catch-all-fields")]
     #[serde(flatten)]
-    pub other: HashMap<String, Value>,
+    pub other: serde_json::Map<String, Value>,
 }
 
 #[cfg(test)]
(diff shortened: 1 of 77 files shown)


lambda-events/src/event/activemq/mod.rs
```diff
@@ -11,6 +13,12 @@ pub struct ActiveMqEvent {
     #[serde(default)]
     pub event_source_arn: Option<String>,
     pub messages: Vec<ActiveMqMessage>,
+    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Enabled with Cargo feature `catch-all-fields`.
+    /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
+    #[cfg(feature = "catch-all-fields")]
+    #[serde(flatten)]
+    pub other: HashMap<String, Value>,
```
jlizen: Is there a reason to deserialize to a `HashMap` rather than `serde_json::Map` if anyway we are exposing the `serde_json::Value` type? That would generally be a lighterweight API since it is more restrictive in keys, and also it allows callers to inject json-relevant feature flags such as `preserve_order`?

lambda-events/src/event/activemq/mod.rs
```diff
@@ -11,6 +13,12 @@ pub struct ActiveMqEvent {
     #[serde(default)]
     pub event_source_arn: Option<String>,
     pub messages: Vec<ActiveMqMessage>,
+    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Enabled with Cargo feature `catch-all-fields`.
+    /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
+    #[cfg(feature = "catch-all-fields")]
```
jlizen, line 19: These all also need the cfg directive to make the feature flag show up in docs.rs:
```
#[cfg_attr(docsrs, doc(cfg(feature = "catch-all-fields")))]
```

lambda-http/src/response.rs
```diff
@@ -77,6 +77,9 @@ impl LambdaResponse {
                 // "multi_value_headers" fields together resulting in duplicate response headers.
                 headers: HeaderMap::new(),
                 multi_value_headers: headers,
+                // Today, this implementation doesn't provide any additional fields
+                #[cfg(feature = "catch-all-fields")]
```
jlizen, line 81: These need the comment about usage/feature gating too. I also probably would remove the `Today...` line unless we plan to go back in and update it if we notice other fields being added... the nature of this feature is, it will sweep up fields that the library authors aren't aware of.

lambda-integration-tests/src/helloworld.rs
```diff
@@ -22,6 +22,8 @@ async fn func(_event: LambdaEvent<ApiGatewayProxyRequest>) -> Result<ApiGatewayP
         is_base64_encoded: false,
         body: Some("Hello world!".into()),
         headers,
+        #[cfg(feature = "catch-all-fields")]
+        other: Default::default(),
```
jlizen, line 26: This doesn't make a whole lot of sense in our integration test unless you want to add a CI workflow that runs it with the feature flag enabled. In which case, feel free.

lambda-events/src/event/activemq/mod.rs
```diff
@@ -11,6 +13,12 @@ pub struct ActiveMqEvent {
     #[serde(default)]
     pub event_source_arn: Option<String>,
     pub messages: Vec<ActiveMqMessage>,
+    /// Catchall to catch any additional fields that were present but not expected by this struct.
```
jlizen: Maybe `but not expected by this struct` -> `but not explicitly defined by this struct`?

jlizen (changes requested):
Core impl seems reasonable to me, thanks for cutting this.

We definitely will want to add the docsrs attribute so our rustdocs have the nice feature gating displayed.

Also some bikeshedding on type representation (`HashMap` -> `serde_json::Map`).

We probably also will want to bump the relevant crate minor versions to avoid semver breakage, since we don't yet have our structs tagged with `#[non_exhaustive]` ([issue](https://github.com/awslabs/aws-lambda-rust-runtime/issues/1016)) - while this the field is under a new feature, technically somebody using `--all-features` would break, and it seems harmless just to bump.

Finally, it would be really nice to have a simple example  in `./examples` demonstrating usage of this functionality (ie a `Cargo.toml` enabling it, some sort of handler logic that checks whether a given field is present in the 'other' field, dump the full field output to a log line, etc. But I wouldn't block the PR on it, just a nice to have if you have cycles.

lambda-events/src/event/activemq/mod.rs
cdisselkoen: Happy to change to `serde_json::Map`

lambda-http/src/response.rs
jlizen: Disregard the missing doc comment, missed that this was just inline construction, not the struct definition

I would probably still ditch the "today..." but up to you, not a major concern

lambda-events/src/event/activemq/mod.rs
cdisselkoen: Made this change and pushed to this PR

lambda-events/src/event/activemq/mod.rs
cdisselkoen: Added and pushed to this PR

lambda-events/src/event/activemq/mod.rs
```diff
@@ -41,13 +49,25 @@ pub struct ActiveMqMessage {
     #[serde(deserialize_with = "deserialize_lambda_map")]
     #[serde(default)]
     pub properties: HashMap<String, String>,
+    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Enabled with Cargo feature `catch-all-fields`.
+    /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
+    #[cfg(feature = "catch-all-fields")]
+    #[serde(flatten)]
+    pub other: serde_json::Map<String, Value>,
```
jlizen, line 59: Another wrinkle (sorry) - would it be worth making these `Option<Map<String, Value>>` to save the allocation for the vast majority of cases where there are no extra fields? Ergonomics are slightly worse for people referencing the fields, but very minor.

lambda-events/src/event/activemq/mod.rs
```diff
@@ -41,13 +49,25 @@ pub struct ActiveMqMessage {
...
 #[serde(rename_all = "camelCase")]
 pub struct ActiveMqDestination {
     #[serde(default)]
     pub physical_name: Option<String>,
+    /// Catchall to catch any additional fields that were present but not expected by this struct.
+    /// Enabled with Cargo feature `catch-all-fields`.
+    /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
+    #[cfg(feature = "catch-all-fields")]
+    #[serde(flatten)]
+    pub other: serde_json::Map<String, Value>,
```
jlizen, line 73: Just sharing a thought - I was wondering this is worth making `Option<Map<String, Value>>`. But, given that anyway the size of the field would be one bit + the largest variant, this doesn't really get us very much. Just mostly is a question of ergonomics for accessing it. And probably skipping unpacking the option is preferable for consumers, they can already use `.is_empty()`.

I guess the map could be boxed, but it already is small since it is backed by allocated members.

lambda-integration-tests/src/helloworld.rs
cdisselkoen: Would you suggest removing the `catch-all-fields` Cargo feature from `lambda-integration-tests`?  I tried this but then `cargo test --all-features` fails because it expects us to initialize `other` here.  Or do we not care about `cargo test --all-features`?

jlizen: Ah, I missed that we were invoking with `--all-features`. In that case, it's fine to leave it.

jlizen:
Thanks for those initial changes! Only blocking thing left would be that we need to bump the minor version in `Cargo.toml` (and then update `lambda-http`/`lambda-integration-tests` path dependency to use the newer version).

Up to you about messing with comments / adding an example, just let me know if you do plan to do it and I'll hold off on merging once the semver thing lands.

lambda-events/src/event/activemq/mod.rs
cdisselkoen: Find-and-replaced, and pushed to this PR

lambda-events/src/event/activemq/mod.rs
cdisselkoen: My thought is consumers would prefer an empty map to `None` -- and with the `Option`, it raises the question of whether `Some(empty map)` is a legal state or what it means if it is legal.

cdisselkoen:
Made the suggested change on comments; if you're willing to merge without an example, I would be happy to accept that.

jlizen (approved):
Thanks for the quick turnaround on feedback! 

Here's the issue to track a new release: https://github.com/awslabs/aws-lambda-rust-runtime/issues/1019

bishnu-2003:
ok

PR #1022, opened 2025-07-19
fix(lambda-events): derive Default on various Events

pmd3d:
Issue #1009

Add derive default to various event types. CloudFormationCustomResourceRequest has each enum with a type parameter so manually wrote a default trait implementation that did not need a type parameter.

- [skipped this because it changed code outside scope of the issue] I confirm that I've ran `cargo clippy --fix`.

### DIFF
--- a/lambda-events/src/encodings/time.rs
+++ b/lambda-events/src/encodings/time.rs
@@ -7,7 +7,7 @@ use serde::{
 use std::ops::{Deref, DerefMut};
 
 /// Timestamp with millisecond precision.
-#[derive(Clone, Debug, Deserialize, Eq, PartialEq, Serialize)]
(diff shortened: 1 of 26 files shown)


jlizen (approved):
Thanks for the contribution! This seems reasonable to me.

These default impls are bit opinionated with the defaults for the enum variants, but, in practice they are only really useful for tests or allowing us to add (#[non-exhaustive] annotations](https://github.com/awslabs/aws-lambda-rust-runtime/issues/1016) to use with future (Optional) fields. It's not like all of these String fields defaulting to `""` would be functional anyway. I don't see a great alternative to this approach that would still allow us to get out of the pit of semver hazards as fields are gradually added.

I would be more concerned around default implementations on response structs since consumers would actually be constructing those directly in their binaries. And, that's not part of this PR.

PR #1034, opened 2025-08-26
chore: prepare lambda_http@0.17.0, lambda_events@0.18.0, lambda_runtime_api_client@0.12.4, lambda_runtime@0.14.4

jlizen:
📬 *Issue #, if available:*
n/a

✍️ *Description of changes:*
Preparing releases for new changes to various crates.

### Breaking changes
I was considering https://github.com/awslabs/aws-lambda-rust-runtime/pull/955 as a semver breaking change since it is changing the serialized output type of a field. It seems like it would mostly break tests since customers are not really serializing request structs in the wild, rather they are deserializing. But test breakage is still breakage.

And since that is a breaking change, `lambda_http` would be breaking by bumping its dependency.

Glad to revise this if other maintainers feel differently.

### Not bumping patch versions on path dependencies
Meanwhile, I intentionally did not bump path dependency versions for non-breaking (patch version) changes. Since, all that really accomplishes is break somebody that for some odd reason wants to pin back to an older version and/or is using an older mirror of the crates.io index or something.

Anyway cargo resolution will use the newest version for any consumers that aren't pinning it. So, I don't see much benefit.

Generally you only see larger repos like tokio bumping patch versions of patch dependencies when they want to explicitly use a new API.

### DIFF
--- a/lambda-events/Cargo.toml
+++ b/lambda-events/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aws_lambda_events"
-version = "0.17.0"
+version = "0.18.0"
 rust-version = "1.81.0"
 description = "AWS Lambda event definitions"
 authors = [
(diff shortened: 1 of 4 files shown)


bnusunny (approved):
LGTM

Issue #1028, opened 2025-08-18
[aws_lambda_events] Add events for VPC Lattice


andrewpatto:
A relatively new service I know - but one that matches well with existing API gateway support (it will not be much extra to add support with `lambda-http` - I will make a separate ticket for that).

First step I guess would be to add some of its event structure types.

https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html

jlizen:
Event types are landed and will go out in the next release, looks like next step will be the `lambda-http` side? Do you prefer to close this ticket out in favor of #1029 @andrewpatto ?

andrewpatto:
Yes no problems. We have done a version of the lambda-http PR - just waiting for the dust to settle on the event types first before pushing. So we'll put that in this afternoon.

PR #1036, opened 2025-08-29
VPC Lattice event structures

andrewpatto:
📬 *Issue #, if available:*

#1028 

✍️ *Description of changes:*

Added event structures for the VPC lattice service - including tests and sample request/responses.


GradExMachina:
If you don't mind, I will take over this PR to make adjustments based on the changes we've done over the last few months (mainly non-exhaustive).

lambda-events/src/event/vpc_lattice/common.rs
```diff
@@ -0,0 +1,62 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct VpcLatticeResponse {
+    // https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html#respond-to-service
+    /// Whether the body is base64 encoded
+    #[serde(default)]
+    pub is_base64_encoded: bool,
+
+    /// The HTTP status code for the request
+    // i64 for consistency with other event types (e.g. AlbTargetGroupResponse, ApiGatewayProxyResponse)
+    pub status_code: i64,
```
jlizen: Minor: oof it's a bit odd to use 64 bits for a http status code, I would have gone with u16. I guess it is a breaking change for the others, but maybe new events should use more efficient sizing?

lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -0,0 +1,240 @@
...
+/// `VpcLatticeRequestV2` contains data coming from VPC Lattice service (V2 format)
+/// see: <https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html#receive-event-from-service>
+/// for field definitions.
+#[non_exhaustive]
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VpcLatticeRequestV2 {
+    /// The version of the event structure (always "2.0" for V2)
+    #[serde(default)]
+    pub version: Option<String>,
```
jlizen: Is this actually optional? I would have expected an error, or an implicit default if unset. (Though implicit default would be annoying to change when v3 comes out)

lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -0,0 +1,240 @@
...
+
+    /// The url path for the request
+    #[serde(default)]
+    pub path: Option<String>,
```
jlizen, line 29: Same question, is this actually optional?

lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -0,0 +1,240 @@
...
+    #[serde(default)]
+    pub identity: Option<VpcLatticeRequestV2Identity>,
+
+    /// AWS region where the request is processed
+    #[serde(default)]
+    pub region: Option<String>,
+
+    /// Time of the request in microseconds since epoch
+    #[serde(default)]
+    pub time_epoch: Option<String>,
```
jlizen: Is there a reason we wouldn't want to set an implicit default if this is unset? I guess it would be misleading to calculate it from the lambda's perspective?

andrewpatto:
> If you don't mind, I will take over this PR to make adjustments based on the changes we've done over the last few months (mainly non-exhaustive).

No problems. The original PR was a straight conversion of the schemas from the docs and mirroring other conventions from this library. Happy for actual experts to take over!

lambda-events/src/custom_serde/headers.rs
```diff
@@ -124,13 +151,51 @@ impl<'de> Visitor<'de> for HeaderMapVisitor {
...
+    de.deserialize_option(HeaderMapVisitor {
+        is_human_readable,
+        split_comma_separated: false,
+    })
+}
+
+/// Deserialize headers, splitting comma-separated values into multiple header values.
+/// Used by VPC Lattice V1 which sends multi-value headers as "value1, value2".
+#[cfg(feature = "vpc_lattice")]
+pub(crate) fn deserialize_comma_separated_headers<'de, D>(de: D) -> Result<HeaderMap, D::Error>
```
jlizen, line 190: Oof, thanks for slogging through this one.

lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -0,0 +1,247 @@
...
+/// see: <https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html#receive-event-from-service>
+/// for field definitions.
+#[non_exhaustive]
+#[cfg_attr(feature = "builders", derive(Builder))]
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct VpcLatticeRequestV2 {
+    /// The version of the event structure (always "2.0" for V2)
+    #[serde(default = "default_version")]
+    pub version: String,
```
jlizen, line 23: Minor: if it is actually always `2.0`, should we consider unconditionally deserializing to a `&'static str`?

I guess, it could easily become `2.1` eventually, at which point we are stuck managing more complicated deserialization logic, so better to take the minor perf hit of a string?

Probably the ideal would be something like a `Cow<'static, str>` so that we could initially mark it with `#[serde(skip)]` and have custom deserialization, but we could roll forward with owned in the future.

Anyway I think it's probably ok to optimize for simplicity, just wondering out loud.

lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -0,0 +1,247 @@
...
+    /// Identity information for the request.
+    /// Present only if authentication is configured.
+    #[serde(default)]
+    pub identity: Option<VpcLatticeRequestV2Identity>,
+
+    /// AWS region where the request is processed
+    pub region: String,
+
+    /// Time of the request in microseconds since epoch
+    pub time_epoch: String,
```
jlizen, line 93: nit: I might personally prefer `time_epoch_micros` with a `#[serde(rename = "timeEpoch")]`, but I can't tell if that's more confusing to deviate from upstream. Just my assumption when I see epoch time, is not micros ^^

lambda-events/src/fixtures/example-vpc-lattice-v1-request.json
```diff
@@ -0,0 +1,19 @@
+{
+  "raw_path": "/api/product",
+  "method": "POST",
+  "headers": {
+    "accept": "*/*",
+    "user-agent": "curl/7.68.0",
+    "x-forwarded-for": "10.0.2.100",
+    "authorization": "Bearer abc123def456",
+    "multi": "abcd, DEF"
```
jlizen, line 9: thanks for adding this one

lambda-events/src/fixtures/example-vpc-lattice-v2-request.json
```diff
@@ -0,0 +1,30 @@
+{
+  "version": "2.0",
+  "path": "/health",
+  "method": "GET",
+  "headers": {
+    "accept": ["*/*"],
+    "user-agent": ["curl/7.68.0"],
+    "x-forwarded-for": ["10.0.2.100"],
+    "multi": ["x", "y"]
```
jlizen, line 9: and thanks for this one

jlizen (approved):
Thanks for the cleanup! Some nits that are all probably bad ideas anyway, I'm fine with what you have.

jlizen:
@maxday / @mpindaru would it be possible to get a second pair of eyes from somebody on your team for [correctness relative to spec](https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html)?

jlizen:
I did my own double check of the schema. Specifically, I diffed the fixtures against the [upstream docs](https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html).

The only issues found were:
1. We have a vestigial `statusDescription` in the fixture, which is dropped from the response struct (and isn't in AWS docs). I'll remove it.

2. We have `method` and `path` set to mandatory, but the AWS docs imply they should be Option<T>:
> Present only if the protocol is HTTP, HTTPS, or gRPC

Meanwhile VPC Lattice's limitations section states:
>  The protocol must be HTTP or HTTPS

So, today it's true that they are effectively present. But it's not clear that we are guaranteed a V3 in case other protocols might be added under V2, in which case we have an issue.

Which... is unfortunate, since it would force everybody to unpack the Option for a hypothetical future state... but I think probably we should follow the letter of the schema here, and keep it optional.

jlizen:
Semver failure is because we need a patch version related to a separate deprecation PR. Otherwise lgtm.

PR #1054, opened 2025-11-14
Update documentation links from awslabs/aws-lambda-rust-runtime to aws/aws-lambda-rust-runtime

ghost:
Update docs to point to aws/aws-lambda-rust-runtime instead of awslabs/aws-lambda-rust-runtime. This updates links and textual references in documentation files only.


jlizen (approved):
The rustdoc check is failing with errors, which are probably worth looking at, but appear unrelated. Might be worth a separate issue to investigate?

Issue #1035, opened 2025-08-27
[CI] Integrated with cargo trusted publishing, release-plz


jlizen:
Crates.io now supports trusted publishing: https://crates.io/docs/trusted-publishing

We should use it as it no longer involves generating api tokens that are individually owned. Better to deploy via CI using repo trust relationship. It also gives our users confidence that what is in crates.io is what is in github.

We probably should use release-plz rather than a naked cargo publish like in the crates.io doc:
https://github.com/release-plz/release-plz

Sample usage:
- https://github.com/async-profiler/rust-agent/blob/main/.github/workflows/release.yml

bnusunny:
Agreed

jlizen:
More `release-plz` usage examples:
- https://github.com/awslabs/metrique/blob/main/release-plz.toml
- https://github.com/awslabs/metrique/blob/main/.github/workflows/release.yml

GradExMachina:
Closing as completed via #1083

PR #1061, opened 2025-11-18
Use release-plz for automated releases

ghost:
📬 *Issue #, if available:*

✍️ *Description of changes:*

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,4 +1,4 @@
-name: Release-plz
+name: Publish release
 
 on:
   push:
@@ -19,17 +19,19 @@ jobs:
         uses: actions/checkout@v5
         with:
           fetch-depth: 0
-          persist-credentials: true
-          token: ${{ secrets.RELEASE_PLZ_TOKEN }}
       - &install-rust
         name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
+      - name: Authenticate to crates.io
+        uses: rust-lang/crates-io-auth-action@v1
+        id: auth
       - name: Run release-plz
         uses: release-plz/action@v0.5
         with:
           command: release
         env:
           GITHUB_TOKEN: ${{ secrets.RELEASE_PLZ_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
 
   release-plz-pr:
     name: Release-plz PR


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,52 @@
...
+      contents: write
+      id-token: write
+    steps:
+      - &checkout
+        name: Checkout repository
+        uses: actions/checkout@v5
+        with:
+          fetch-depth: 0
+          persist-credentials: true
+          token: ${{ secrets.RELEASE_PLZ_TOKEN }}
```
jlizen: you shouldn't need the token on this step?

.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,52 @@
...
+      - &install-rust
+        name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Run release-plz
```
jlizen, line 28: I highly recommend integrating with trusted publishing as part of this PR, which lets you do:
```
      - name: Authenticate to crates.io
        uses: rust-lang/crates-io-auth-action@v1
        id: auth
      - name: Run release-plz
        uses: release-plz/action@v0.5.102
        with:
          command: release
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
          CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
```

https://github.com/awslabs/aws-lambda-rust-runtime/issues/1035

release-plz.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+semver_check = true
+git_release_enable = false
+
+[[package]]
```
jlizen, line 5: I'm not sure if this will work, I think we would need positive statements for our packages if our workspace has it set to false?

Alternative example: 
```
[[package]]
name = "metrique"
changelog_include = [
    "metrique-core",
    "metrique-macro",
    "metrique-metricsrs",
    "metrique-service-metrics",
    "metrique-timesource",
    "metrique-writer",
    "metrique-writer-core",
    "metrique-writer-format-emf",
    "metrique-writer-macro"
]
changelog_path = "./CHANGELOG.md"
git_release_enable = true
```

But worst case we can always give it a shot and fix if it doesn't work, low stakes.

release-plz.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+semver_check = true
```
jlizen, line 2: This is probably redundant since we already have cargo-semver-check CI, but harmless

.github/workflows/release-plz.yml
ghost: Autogenerated by release-plz 🤔

release-plz.toml
ghost: Intended to have it explicit 👍🏻

.github/workflows/release-plz.yml
ghost: It seems that once you enable using a personal access token it sets it automatically without even though it's not needed.

Issue #1065, opened 2025-12-01
Support Lambda Managed Instances


bnusunny:
AWS just launched Lambda Managed Instances [What's New](https://aws.amazon.com/about-aws/whats-new/2025/11/aws-lambda-managed-instances/). It supports multiple concurrent requests on a single execution environment. We should support it to get the benefit to the Rust Runtime users. 

To support multi-concurrency, we can spawn a new async task for each invoke. Both `lambda-runtime` and `lambda-http` crates need new run methods because the tower service instances need implement `Clone` trait. Adding `Clone` trait to the exiting run methods will be a breaking change. 

X-ray trace header propagation should not use the environment variable. We need use a thread-safe way to store the x-tray trace header for each request.

jlizen:
Re:
> we can spawn a new async task for each invoke

(This is assuming this feature is only going to be implemented for tokio, initially).

Just as an implementation suggestion, we probably should make sure that we aren't dispatching work from the tokio `block_on` thread (aka [non-worker thread](https://docs.rs/tokio/latest/tokio/runtime/struct.Runtime.html#non-worker-future)), but rather from a worker thread (ie inside a task).

We don't control whether the caller awaits the lambda_runtime future on a worker thread or not, so it seems prudent to always spawn a task to drive the receive loop.

This is a performance optimization to avoid ping ponging the socket between the I/O driver on a worker thread, and the non-worker thread.

jlizen:
Another thing to think about would be, how does this fit into [turnkey graceful shutdown handling](https://docs.rs/lambda_runtime/latest/lambda_runtime/fn.spawn_graceful_shutdown_handler.html). The simple answer is, just fire the graceful shutdown once, globally, but I imagine the nicer thing might be some signaling to allow each accept loop to clean up its own state.

jlizen:
> new async task for each invoke

Correction (discussed offline): new task per event loop based on max concurrency, not per invoke

garysassano:
Thanks for opening this; now I understand why my containerized function that uses the Lambda Rust Runtime breaks as soon as I attach a Capacity Provider to it.

I also noticed that AWS recently updated the Lambda docs with a section on [building custom runtimes for Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html#runtimes-custom-managed-instances), which helps clarify how custom runtimes need to be implemented in that setup..

alessandrobologna:
[...]

PR #1067, opened 2025-12-10
feat: support Lambda Managed Instances

alessandrobologna:
📬 Issue #: aws/aws-lambda-rust-runtime#1065

✍️ Description of changes:

Support Lambda Managed Instances / concurrent runtime (refs #1065)

> [!NOTE]  
> _I originally built this to experiment with Lambda Managed Instances in a test workload and thought it might be a useful starting point for upstream support. Parts of the implementation were drafted with the help of an AI assistant and have only been exercised in my own workloads and the existing test suite so far, so please treat this as a starting point for discussion rather than a final design. A super simple demo/benchmark of this implementation is at https://github.com/alessandrobologna/rust-lmi-demo_

This draft PR implements a concurrent runtime mode for Lambda Managed Instances, plus the API changes required in `lambda-runtime` and `lambda-http`. It preserves existing single-invocation behavior for classic Lambda. The implementation is based on the AWS guidance for building runtimes that support Lambda Managed Instances, as described in [Building custom runtimes for Lambda Managed Instances](https://docs.aws.amazon.com/lambda/latest/dg/runtimes-custom.html#runtimes-custom-managed-instances).

**Architecture (high level)**

The runtime chooses between a classic "one invocation at a time" loop and a concurrent mode controlled by `AWS_LAMBDA_MAX_CONCURRENCY`:

- Sequential mode (default): a single `/next` long-poll loop fetches one event at a time and invokes the handler synchronously, preserving existing behavior.
- Concurrent mode (`AWS_LAMBDA_MAX_CONCURRENCY >= 2`): the runtime spawns `AWS_LAMBDA_MAX_CONCURRENCY` worker tasks. Each worker runs its own `/next` long-poll loop and processes invocations sequentially within that task, so overall concurrency is bounded by the number of workers. If a worker crashes, the runtime continues operating with the remaining workers and only terminates if it can't keep at least 1 worker alive.

```mermaid
sequenceDiagram
    participant Lambda as Lambda orchestrator
    participant Runtime as lambda-runtime
    participant Worker as Runtime worker task
    participant API as Runtime API
    participant Handler as User handler

    Lambda->>Runtime: Start process (env AWS_LAMBDA_MAX_CONCURRENCY = N)
    Runtime->>Worker: Spawn N worker tasks
    loop each worker (identical)
        loop each invocation
            Worker->>API: GET /runtime/invocation/next
            API-->>Worker: Invocation event + context headers
            Worker->>Handler: Call handler(event, context)
            Handler-->>Worker: Result or error
            Worker->>API: POST /response or /error
        end
    end
```

**Breaking changes & compatibility**

- Existing entrypoints (`lambda_runtime::run`, `lambda_http::run`, and `lambda_http::run_with_streaming_response`) keep their original signatures and sequential behavior. Handlers that compiled against the current release should continue to compile unchanged.
- New concurrent entrypoints are added:
  - `lambda_runtime::run_concurrent`
  - `lambda_http::run_concurrent`
  - `lambda_http::run_with_streaming_response_concurrent`
  These require handler services to implement `Clone + Send + 'static` (with responses/stream bodies `Send`/`Sync + 'static`) so they can be safely cloned and driven by the concurrent runtime.
- `AWS_LAMBDA_MAX_CONCURRENCY` is read directly by the runtime to decide concurrency and size the HTTP pool, without adding new public fields to `Config` (avoids semver breakage for struct literals).
- In concurrent mode (when the new entrypoints are used and `AWS_LAMBDA_MAX_CONCURRENCY > 1`), the runtime no longer sets `_X_AMZN_TRACE_ID` in the process environment. The per-invocation X-Ray trace ID is available via `Context::xray_trace_id` and tracing spans instead. Sequential mode behavior is unchanged.
- For the concurrent entrypoints, the behavior when `AWS_LAMBDA_MAX_CONCURRENCY` is set changes from "always sequential per environment" to "per-environment concurrency up to that value". Code that continues to call the existing `run` functions will remain strictly sequential even if the env var is set.

In other words: the earlier versions of this branch tightened the bounds on the existing `run` functions, but after maintainer feedback those entrypoints are left as-is and concurrency is opt-in via the new `*_concurrent` APIs.

Below is a concise summary of the changes (unfortunately many) by area.

**Runtime & Config (`lambda-runtime`)**

- Read `AWS_LAMBDA_MAX_CONCURRENCY` directly inside the runtime to decide whether concurrent mode should be enabled and to size the HTTP client pool, without extending `Config`.
- `Runtime::new` now sizes the `lambda_runtime_api_client` HTTP pool from `max_concurrency` so the number of idle connections matches expected concurrency.
- `Runtime::run` remains the original sequential `/next` loop via `run_with_incoming`, preserving existing behavior.
- New `Runtime::run_concurrent` spawns `AWS_LAMBDA_MAX_CONCURRENCY` worker tasks, each running its own `/next` loop. Sequential and concurrent modes share the same invocation processing helper, so there are not two separate event loop implementations to maintain. When `Config::is_concurrent()` is false, it falls back to the same sequential loop as `Runtime::run`.
- Worker failures are isolated: if a worker exits/panics, the runtime logs the failure and continues with fewer workers. The runtime terminates only if it can't keep at least 1 worker alive (i.e., all workers have exited).
- Unexpected clean worker exits are treated as infrastructure errors so the runtime returns an error once all workers are gone.
- Handler errors (user code returning `Err`) are reported to Lambda via `/invocation/{id}/error` and do not terminate the runtime.
- The existing graceful shutdown handler stays aligned with upstream. We do not attempt to preempt/coordinate in-flight invocations on SIGTERM (the expectation is that Lambda only shuts down environments when they are idle; local emulators may behave differently).
- Internal layers (`api_client`, `api_response`, `trace`) now implement/derive `Clone` so they can be composed into a cloneable service stack for the concurrent entrypoints.
- `Context::new` is more robust when `lambda-runtime-client-context` / `lambda-runtime-cognito-identity` headers are present but empty (treated as `None` instead of failing JSON parse).
- Added a regression test (`concurrent_worker_crash_does_not_stop_other_workers`) to verify worker isolation behavior.

**HTTP & streaming (`lambda-http`, `lambda-runtime-api-client`)**

- `lambda_runtime_api_client::Client`:
  - Added `with_pool_size(usize)` on the builder and threads a `pool_size: Option<usize>` into the Hyper client to set `pool_max_idle_per_host`.
  - Still works as before when `pool_size` is not provided.
- `lambda_http::run` and `run_with_streaming_response` keep their existing signatures and sequential behavior, delegating to `lambda_runtime::run`.
- New `lambda_http::run_concurrent` and `lambda_http::run_with_streaming_response_concurrent` wrap the same handler types but require them to be `Clone + Send + 'static` (with response/stream bounds aligned to `lambda_runtime::run_concurrent`) so they can be driven by the concurrent runtime.
- HTTP adapters (`Adapter`, `StreamAdapter`) are now `Clone` when the inner service is `Clone`, and the streaming path uses `BoxCloneService` internally for the concurrent entrypoint so the composed service stack can be cloned.
  - Note: `run_with_streaming_response_concurrent` is exported from `lambda_http`, similarly to `run_with_streaming_response`.

**Tooling & examples**

- `Makefile` and `scripts/test-rie.sh`:
  - Add `RIE_MAX_CONCURRENCY` and a `test-rie-lmi` target that runs RIE with `AWS_LAMBDA_MAX_CONCURRENCY` set, making it easy to exercise managed-instances behavior locally.
- `examples/basic-lambda-concurrent`:
  - New minimal example that calls `lambda_runtime::run_concurrent` so the concurrent code path can be exercised under RIE/LMI.
- `examples/basic-lambda/src/main.rs`:
  - Wraps `lambda_runtime::run(func).await` in an `if let Err(err)` block to log and propagate runtime errors when testing under RIE.

---

**Validation**

On this branch, I ran:

- `cargo +nightly fmt --all -- --check`
- `cargo clippy --workspace --all-features -- -D warnings`
- `cargo +stable test --all-features -p lambda_runtime_api_client -p lambda_runtime -p lambda_http`
- `make test-rie EXAMPLE=basic-lambda`
- `make test-rie-lmi EXAMPLE=basic-lambda-concurrent` (sets `AWS_LAMBDA_MAX_CONCURRENCY` inside the RIE container)

---

If maintainers prefer, this could be split into smaller PRs (e.g., builder/Config prep, handler `Clone` changes, and finally the concurrent runtime), but this branch shows the full "end-to-end" implementation so that it can be tested with Lambda Managed Instances.

---

🔏 By submitting this pull request

- [x] I confirm that I've ran `cargo +nightly fmt --all -- --check`.
- [x] I confirm that I've ran `cargo clippy --workspace --all-features -- -D warnings`.

### DIFF
--- a/lambda-http/src/lib.rs
+++ b/lambda-http/src/lib.rs
@@ -102,7 +102,9 @@ use std::{
 };
 
 mod streaming;
-pub use streaming::{run_with_streaming_response, run_with_streaming_response_concurrent, StreamAdapter};
+#[cfg(feature = "experimental-concurrency")]
+pub use streaming::run_with_streaming_response_concurrent;
+pub use streaming::{run_with_streaming_response, StreamAdapter};
 
 /// Type alias for `http::Request`s with a fixed [`Body`](enum.Body.html) type
 pub type Request = http::Request<Body>;
@@ -195,7 +197,7 @@ where
         let LambdaEvent { payload, context } = req;
         let request_origin = payload.request_origin();
         let mut event: Request = payload.into();
-        update_xray_trace_id_header_from_context(event.headers_mut(), &context);
+        update_xray_trace_id_header(event.headers_mut(), &context);
         let fut = Box::pin(self.service.call(event.with_lambda_context(context)));
 
         TransformResponse::Request(request_origin, fut)
@@ -211,8 +213,9 @@ where
 /// # Managed concurrency
 /// If `AWS_LAMBDA_MAX_CONCURRENCY` is set, this function returns an error because
 /// it does not enable concurrent polling. If your handler can satisfy `Clone`,
-/// prefer [`run_concurrent`], which honors managed concurrency and falls back to
-/// sequential behavior when unset.
+/// prefer [`run_concurrent`] (requires the `experimental-concurrency` feature),
+/// which honors managed concurrency and falls back to sequential behavior when
+/// unset.
 pub async fn run<'a, R, S, E>(handler: S) -> Result<(), Error>
 where
     S: Service<Request, Response = R, Error = E>,
@@ -226,11 +229,15 @@ where
 /// Starts the Lambda Rust runtime in a mode that is compatible with
 /// Lambda Managed Instances (concurrent invocations).
 ///
+/// Requires the `experimental-concurrency` feature.
+///
 /// When `AWS_LAMBDA_MAX_CONCURRENCY` is set to a value greater than 1, this
 /// will spawn `AWS_LAMBDA_MAX_CONCURRENCY` worker tasks, each running its own
 /// `/next` polling loop. When the environment variable is unset or `<= 1`,
 /// it falls back to the same sequential behavior as [`run`], so the same
 /// handler can run on both classic Lambda and Lambda Managed Instances.
+#[cfg(feature = "experimental-concurrency")]
+#[cfg_attr(docsrs, doc(cfg(feature = "experimental-concurrency")))]
 pub async fn run_concurrent<R, S, E>(handler: S) -> Result<(), Error>
 where
     S: Service<Request, Response = R, Error = E> + Clone + Send + 'static,
@@ -241,8 +248,8 @@ where
     lambda_runtime::run_concurrent(Adapter::from(handler)).await
 }
 
-// Replaces update_xray_trace_id_header (env var), now set from Context
-fn update_xray_trace_id_header_from_context(headers: &mut http::HeaderMap, context: &Context) {
+// In concurrent mode we must use the per-request context.
+fn update_xray_trace_id_header(headers: &mut http::HeaderMap, context: &Context) {
     if let Some(trace_id) = context.xray_trace_id.as_deref() {
         if let Ok(header_value) = http::HeaderValue::from_str(trace_id) {
             headers.insert(http::header::HeaderName::from_static("x-amzn-trace-id"), header_value);
--- a/lambda-http/src/streaming.rs
+++ b/lambda-http/src/streaming.rs
@@ -1,6 +1,4 @@
-use crate::{
-    http::header::SET_COOKIE, request::LambdaRequest, update_xray_trace_id_header_from_context, Request, RequestExt,
-};
+use crate::{http::header::SET_COOKIE, request::LambdaRequest, update_xray_trace_id_header, Request, RequestExt};
 use bytes::Bytes;
 use core::{
     fmt::Debug,
@@ -76,7 +74,7 @@ where
     fn call(&mut self, req: LambdaEvent<LambdaRequest>) -> Self::Future {
         let LambdaEvent { payload, context } = req;
         let mut event: Request = payload.into();
-        update_xray_trace_id_header_from_context(event.headers_mut(), &context);
+        update_xray_trace_id_header(event.headers_mut(), &context);
         Box::pin(
             self.service
                 .call(event.with_lambda_context(context))
@@ -116,8 +114,10 @@ where
 
 /// Builds a streaming-aware Tower service from a `Service<Request>` that can be
 /// cloned and sent across tasks. This is used by the concurrent HTTP entrypoint.
+#[cfg(feature = "experimental-concurrency")]
 type EventToRequest = fn(LambdaEvent<LambdaRequest>) -> Request;
 
+#[cfg(feature = "experimental-concurrency")]
 #[allow(clippy::type_complexity)]
 fn into_stream_service_cloneable<S, B, E>(
     handler: S,
@@ -166,7 +166,7 @@ where
 fn event_to_request(req: LambdaEvent<LambdaRequest>) -> Request {
     let LambdaEvent { payload, context } = req;
     let mut event: Request = payload.into();
-    update_xray_trace_id_header_from_context(event.headers_mut(), &context);
+    update_xray_trace_id_header(event.headers_mut(), &context);
     event.with_lambda_context(context)
 }
 
@@ -178,7 +178,7 @@ fn event_to_request(req: LambdaEvent<LambdaRequest>) -> Request {
 /// # Managed concurrency
 /// If `AWS_LAMBDA_MAX_CONCURRENCY` is set, this function returns an error because
 /// it does not enable concurrent polling. Use [`run_with_streaming_response_concurrent`]
-/// instead.
+/// (requires the `experimental-concurrency` feature) instead.
 ///
 /// [AWS docs for response streaming]:
 ///     https://docs.aws.amazon.com/lambda/latest/dg/configuration-response-streaming.html
@@ -197,9 +197,13 @@ where
 /// Runs the Lambda runtime with a handler that returns **streaming** HTTP
 /// responses, in a mode that is compatible with Lambda Managed Instances.
 ///
+/// Requires the `experimental-concurrency` feature.
+///
 /// This uses a cloneable, boxed service internally so it can be driven by the
 /// concurrent runtime. When `AWS_LAMBDA_MAX_CONCURRENCY` is not set or `<= 1`,
 /// it falls back to the same sequential behavior as [`run_with_streaming_response`].
+#[cfg(feature = "experimental-concurrency")]
+#[cfg_attr(docsrs, doc(cfg(feature = "experimental-concurrency")))]
 pub async fn run_with_streaming_response_concurrent<S, B, E>(handler: S) -> Result<(), Error>
 where
     S: Service<Request, Response = Response<B>, Error = E> + Clone + Send + 'static,
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -97,8 +97,9 @@ where
 /// # Managed concurrency
 /// If `AWS_LAMBDA_MAX_CONCURRENCY` is set, this function returns an error because
 /// it does not enable concurrent polling. If your handler can satisfy `Clone`,
-/// prefer [`run_concurrent`], which honors managed concurrency and falls back to
-/// sequential behavior when unset.
+/// prefer [`run_concurrent`] (requires the `experimental-concurrency` feature),
+/// which honors managed concurrency and falls back to sequential behavior when
+/// unset.
 ///
 /// # Example
 /// ```no_run
@@ -135,6 +136,8 @@ where
 /// Starts the Lambda Rust runtime in a mode that is compatible with
 /// Lambda Managed Instances (concurrent invocations).
 ///
+/// Requires the `experimental-concurrency` feature.
+///
 /// When `AWS_LAMBDA_MAX_CONCURRENCY` is set to a value greater than 1, this
 /// will spawn `AWS_LAMBDA_MAX_CONCURRENCY` worker tasks, each running its own
 /// `/next` polling loop. When the environment variable is unset or `<= 1`, it
@@ -160,6 +163,8 @@ where
 ///     Ok(event.payload)
 /// }
 /// ```
+#[cfg(feature = "experimental-concurrency")]
+#[cfg_attr(docsrs, doc(cfg(feature = "experimental-concurrency")))]
 pub async fn run_concurrent<A, F, R, B, S, D, E>(handler: F) -> Result<(), Error>
 where
     F: Service<LambdaEvent<A>, Response = R> + Clone + Send + 'static,
--- a/lambda-runtime/src/runtime.rs
+++ b/lambda-runtime/src/runtime.rs
@@ -4,14 +4,19 @@ use crate::{
     types::{invoke_request_id, IntoFunctionResponse, LambdaEvent},
     Config, Context, Diagnostic,
 };
+#[cfg(feature = "experimental-concurrency")]
 use futures::stream::FuturesUnordered;
 use http_body_util::BodyExt;
 use lambda_runtime_api_client::{BoxError, Client as ApiClient};
 use serde::{Deserialize, Serialize};
-use std::{env, fmt, fmt::Debug, future::Future, io, sync::Arc};
+#[cfg(feature = "experimental-concurrency")]
+use std::fmt;
+use std::{env, fmt::Debug, future::Future, io, sync::Arc};
 use tokio_stream::{Stream, StreamExt};
 use tower::{Layer, Service, ServiceExt};
-use tracing::{debug, error, info_span, trace, warn, Instrument};
+use tracing::trace;
+#[cfg(feature = "experimental-concurrency")]
+use tracing::{debug, error, info_span, warn, Instrument};
 
 /* ----------------------------------------- INVOCATION ---------------------------------------- */
 
@@ -149,6 +154,7 @@ impl<S> Runtime<S> {
     }
 }
 
+#[cfg(feature = "experimental-concurrency")]
 impl<S> Runtime<S>
 where
     S: Service<LambdaInvocation, Response = (), Error = BoxError> + Clone + Send + 'static,
@@ -159,6 +165,7 @@ where
     /// If `AWS_LAMBDA_MAX_CONCURRENCY` is not set or is `<= 1`, this falls back to the
     /// sequential `run_with_incoming` loop so that the same handler can run on both
     /// classic Lambda and Lambda Managed Instances.
+    #[cfg_attr(docsrs, doc(cfg(feature = "experimental-concurrency")))]
     pub async fn run_concurrent(self) -> Result<(), BoxError> {
         if self.concurrency_limit > 1 {
             trace!("Concurrent mode: _X_AMZN_TRACE_ID is not set; use context.xray_trace_id");
@@ -252,17 +259,20 @@ where
     }
 }
 
+#[cfg(feature = "experimental-concurrency")]
 #[derive(Debug)]
 enum WorkerError {
     CleanExit(tokio::task::Id),
     Failure(tokio::task::Id, BoxError),
 }
 
+#[cfg(feature = "experimental-concurrency")]
 #[derive(Debug)]
 struct ConcurrentWorkerErrors {
     errors: Vec<WorkerError>,
 }
 
+#[cfg(feature = "experimental-concurrency")]
 #[derive(Serialize)]
 struct ConcurrentWorkerErrorsPayload<'a> {
     message: &'a str,
@@ -272,12 +282,14 @@ struct ConcurrentWorkerErrorsPayload<'a> {
     failures: Vec<WorkerFailurePayload>,
 }
 
+#[cfg(feature = "experimental-concurrency")]
 #[derive(Serialize)]
 struct WorkerFailurePayload {
     id: String,
     err: String,
 }
 
+#[cfg(feature = "experimental-concurrency")]
 impl fmt::Display for ConcurrentWorkerErrors {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         let mut clean = Vec::new();
@@ -314,6 +326,7 @@ impl fmt::Display for ConcurrentWorkerErrors {
     }
 }
 
+#[cfg(feature = "experimental-concurrency")]
 impl std::error::Error for ConcurrentWorkerErrors {}
 
 impl<S> Runtime<S>
@@ -323,11 +336,12 @@ where
     /// Start the runtime and begin polling for events on the Lambda Runtime API.
     ///
     /// If `AWS_LAMBDA_MAX_CONCURRENCY` is set, this returns an error because it does not enable
-    /// concurrent polling. Use [`Runtime::run_concurrent`] instead.
+    /// concurrent polling. Enable the `experimental-concurrency` feature and use
+    /// [`Runtime::run_concurrent`] instead.
     pub async fn run(self) -> Result<(), BoxError> {
         if let Some(raw) = concurrency_env_value() {
             return Err(Box::new(io::Error::other(format!(
-                "AWS_LAMBDA_MAX_CONCURRENCY is set to '{raw}', but Runtime::run does not support concurrent polling; use Runtime::run_concurrent instead"
+                "AWS_LAMBDA_MAX_CONCURRENCY is set to '{raw}', but Runtime::run does not support concurrent polling; enable the experimental-concurrency feature and use Runtime::run_concurrent instead"
             ))));
         }
         let incoming = incoming(&self.client);
@@ -398,6 +412,7 @@ fn incoming(
 }
 
 /// Creates a future that polls the `/next` endpoint.
+#[cfg(feature = "experimental-concurrency")]
 async fn next_event_future(client: &ApiClient) -> Result<http::Response<hyper::body::Incoming>, BoxError> {
     let req = NextEventRequest.into_req()?;
     client.call(req).await
@@ -414,6 +429,7 @@ fn concurrency_env_value() -> Option<String> {
     env::var("AWS_LAMBDA_MAX_CONCURRENCY").ok()
 }
 
+#[cfg(feature = "experimental-concurrency")]
 async fn concurrent_worker_loop<S>(mut service: S, config: Arc<Config>, client: Arc<ApiClient>) -> Result<(), BoxError>
 where
     S: Service<LambdaInvocation, Response = (), Error = BoxError>,
@@ -744,6 +760,7 @@ mod endpoint_tests {
         .await
     }
 
+    #[cfg(feature = "experimental-concurrency")]
     #[tokio::test]
     async fn concurrent_worker_crash_does_not_stop_other_workers() -> Result<(), Error> {
         let next_calls = Arc::new(AtomicUsize::new(0));
--- a/examples/basic-lambda-concurrent/Cargo.toml
+++ b/examples/basic-lambda-concurrent/Cargo.toml
@@ -4,6 +4,6 @@ version = "0.1.0"
 edition = "2021"
 
 [dependencies]
-lambda_runtime = { path = "../../lambda-runtime" }
+lambda_runtime = { path = "../../lambda-runtime", features = ["experimental-concurrency"] }
 serde = "1.0.219"
 tokio = { version = "1", features = ["macros"] }
--- a/lambda-http/Cargo.toml
+++ b/lambda-http/Cargo.toml
@@ -29,6 +29,7 @@ opentelemetry = ["lambda_runtime/opentelemetry"] # enables access to the OpenTel
 anyhow = ["lambda_runtime/anyhow"] # enables From<T> for Diagnostic for anyhow error types, see README.md for more info
 eyre = ["lambda_runtime/eyre"] # enables From<T> for Diagnostic for eyre error types, see README.md for more info
 miette = ["lambda_runtime/miette"] # enables From<T> for Diagnostic for miette error types, see README.md for more info
+experimental-concurrency = ["lambda_runtime/experimental-concurrency"]
 
 [dependencies]
 bytes = { workspace = true }
--- a/lambda-runtime/Cargo.toml
+++ b/lambda-runtime/Cargo.toml
@@ -25,6 +25,7 @@ miette = ["dep:miette"] # enables From<T> for Diagnostic for miette error types,
 # as well as default features
 # https://github.com/aws/aws-lambda-rust-runtime/issues/984
 graceful-shutdown = ["tokio/rt", "tokio/signal", "dep:lambda-extension"]
+experimental-concurrency = []
 
 [dependencies]
 anyhow = { version = "1.0.86", optional = true }


lambda-runtime/src/runtime.rs
```diff
@@ -137,6 +153,92 @@ impl<S> Runtime<S> {
...
+        // API client failures, runtime panics, etc.
+        let mut first_error: Option<BoxError> = None;
+        let mut remaining_workers = limit;
+        while let Some(result) = futures::StreamExt::next(&mut workers).await {
+            remaining_workers = remaining_workers.saturating_sub(1);
+            match result {
+                Ok(Ok(())) => {
+                    // `concurrent_worker_loop` runs indefinitely, so an Ok return indicates
+                    // an unexpected worker exit; we still decrement because the task is gone.
+                    warn!(remaining_workers, "Concurrent worker exited unexpectedly without error");
```
bnusunny: nit: the warning message could be more precise: 

"Concurrent worker exited cleanly (unexpected - loop should run forever)"

alessandrobologna: changed it as suggested

alessandrobologna:
Thanks @bnusunny for the approval!
Currently CI is failing on the semver check because `Config` has a new public field. I have a small follow‑up that keeps `Config` unchanged by moving the concurrency limit into `Runtime` (derived from `AWS_LAMBDA_MAX_CONCURRENCY`), which avoids a major version bump.

Patch: https://github.com/alessandrobologna/aws-lambda-rust-runtime/commit/a8af68a464f0322d212f7e4d74b062c115e2366f

If you’re OK with that approach, I can push it directly to this PR branch so the checks go green.

bnusunny:
Yes, please.

alessandrobologna:
Ok, I've pushed the update to keep concurrency settings internal and to avoid the semver bump (so no Config public field needed).  I also rebased the branch on current main.

Makefile
```diff
@@ -111,4 +112,8 @@ fmt:
 	cargo +nightly fmt --all
 
 test-rie:
-	./scripts/test-rie.sh $(EXAMPLE)
\ No newline at end of file
+	./scripts/test-rie.sh $(EXAMPLE)
+
+# Run RIE in Lambda Managed Instance (LMI) mode with concurrent polling.
+test-rie-lmi:
```
jlizen, line 118: Minor: is this makefile target + configurable variable worth mentioning in the README?

lambda-http/src/lib.rs
```diff
@@ -203,6 +217,33 @@ where
...
 
+/// Starts the Lambda Rust runtime in a mode that is compatible with
+/// Lambda Managed Instances (concurrent invocations).
+///
+/// When `AWS_LAMBDA_MAX_CONCURRENCY` is set to a value greater than 1, this
+/// will spawn `AWS_LAMBDA_MAX_CONCURRENCY` worker tasks, each running its own
+/// `/next` polling loop. When the environment variable is unset or `<= 1`,
+/// it falls back to the same sequential behavior as [`run`], so the same
+/// handler can run on both classic Lambda and Lambda Managed Instances.
+pub async fn run_concurrent<R, S, E>(handler: S) -> Result<(), Error>
```
jlizen, line 241: (Question, not sure): If the fallback case if `AWS_LAMBDA_MAX_CONCURRENCY` if unset is just `lambda_runtime::run()`, should this be our new suggested entrypoint, with `fn run()` marked as deprecated? I guess the main issue would be the `Clone` bounds are more restrictive, so maybe deprecation is a bit aggressive, but we could add that in the `run()` doc comments?

The regular `run()` should definitely also add a warning that it WON'T respect AWS_LAMBDA_MAX_CONCURRENCY.

Context: My suspicion is, most consumers don't mind the `Clone` bound, and the 'principal of least surprise' would be for it to respect the ENV. So we probably should be pointing people to prefer this method where possible.

I would suggest at minimum the following:
- probably this should be named `run_maybe_concurrent`? Or maybe `run_extended` to avoid a combinatorial explosion of new entrypoint names as featureset grows that needs `Clone`?
- We should have a guard clause in `lambda_runtime::run()` that throws either either crashes the lambda runtime (preferred), or emits error logs (more conservative), if `AWS_LAMBDA_MAX_CONCURRENCY` is set, since that is clearly a misconfiguration that would prove confusing to debug.

lambda-http/src/lib.rs
```diff
@@ -203,6 +217,33 @@ where
...
+    R: IntoResponse + Send + Sync + 'static,
+    E: std::fmt::Debug + Into<Diagnostic> + Send + 'static,
+{
+    lambda_runtime::run_concurrent(Adapter::from(handler)).await
+}
+
+// Replaces update_xray_trace_id_header (env var), now set from Context
+fn update_xray_trace_id_header_from_context(headers: &mut http::HeaderMap, context: &Context) {
+    if let Some(trace_id) = context.xray_trace_id.as_deref() {
+        if let Ok(header_value) = http::HeaderValue::from_str(trace_id) {
```
jlizen, line 254: Minor: What are the cases where `http::HeaderValue::from_str()` would fail? Should we have a warn or error-level log line if it does?

(I guess this is an existing behavior and this is more of a lift and shift?)

lambda-http/src/streaming.rs
```diff
@@ -93,14 +97,33 @@ where
...
-        })
+        .map_request(event_to_request as fn(LambdaEvent<LambdaRequest>) -> Request)
         .service(handler)
         .map_response(into_stream_response)
 }
 
+/// Builds a streaming-aware Tower service from a `Service<Request>` that can be
+/// cloned and sent across tasks. This is used by the concurrent HTTP entrypoint.
+#[allow(clippy::type_complexity)]
+fn into_stream_service_boxed<S, B, E>(
```
jlizen: Non-blocking: Do we actually need the extra allocation + indirection of `BoxCloneServiceHere`? Anyway we aren't leaking into the public API any hideous generic output type from this, so we could probably get by with pure generics?

IE, we should be able to use similar bounds with `MapRequest` and `MapResponse` to `into_stream_service`, no? Just add the `Clone` bound and call it `into_stream_service_cloneable`?

Non-blocking since it's a fairly minor perf optimization, and we can change it in the future without semver impact.

examples/basic-lambda/src/main.rs
```diff
@@ -28,7 +28,10 @@ async fn main() -> Result<(), Error> {
     tracing::init_default_subscriber();
 
     let func = service_fn(my_handler);
-    lambda_runtime::run(func).await?;
+    if let Err(err) = lambda_runtime::run(func).await {
```
jlizen, line 31: Thanks for the fixup here

lambda-http/src/streaming.rs
```diff
@@ -1,4 +1,6 @@
-use crate::{http::header::SET_COOKIE, request::LambdaRequest, Request, RequestExt};
+use crate::{
```
jlizen: We probably want a `ClonableStreamAdapter` too right? To have `Clone` bounds to support building custom pre-lambda-event-conversion tower stacks, on concurrent handlers?

This relates to: https://github.com/aws/aws-lambda-rust-runtime/pull/1013

lambda-http/src/streaming.rs
```diff
@@ -147,6 +177,24 @@ where
     lambda_runtime::run(into_stream_service(handler)).await
 }
 
+/// Runs the Lambda runtime with a handler that returns **streaming** HTTP
+/// responses, in a mode that is compatible with Lambda Managed Instances.
+///
+/// This uses a cloneable, boxed service internally so it can be driven by the
+/// concurrent runtime. When `AWS_LAMBDA_MAX_CONCURRENCY` is not set or `<= 1`,
+/// it falls back to the same sequential behavior as [`run_with_streaming_response`].
+pub async fn run_with_streaming_response_concurrent<S, B, E>(handler: S) -> Result<(), Error>
```
jlizen, line 207: Same question around naming `run_with_streaming_response_extended` / `run_with_streaming_response_maybe_concurrent` / etc, and nudging people to prefer this to `run_with_streaming_response`  + warning that it doesn't respect the concurrency env

lambda-runtime-api-client/src/lib.rs
```diff
@@ -59,11 +60,16 @@ impl Client {
...
-        let client = hyper_util::client::legacy::Client::builder(hyper_util::rt::TokioExecutor::new())
-            .http1_max_buf_size(1024 * 1024)
-            .build(connector);
+    /// Create a new client with a given base URI, HTTP connector, and optional pool size hint.
+    fn with(base: Uri, connector: HttpConnector, pool_size: Option<usize>) -> Self {
+        let mut builder = hyper_util::client::legacy::Client::builder(hyper_util::rt::TokioExecutor::new());
+        builder.http1_max_buf_size(1024 * 1024);
+
+        if let Some(size) = pool_size {
+            builder.pool_max_idle_per_host(size);
```
jlizen: nice addition here

lambda-runtime/src/lib.rs
```diff
@@ -126,6 +126,50 @@ where
...
+///     let func = service_fn(func);
+///     lambda_runtime::run_concurrent(func).await?;
+///     Ok(())
+/// }
+///
+/// async fn func(event: LambdaEvent<Value>) -> Result<Value, Error> {
+///     Ok(event.payload)
+/// }
+/// ```
+pub async fn run_concurrent<A, F, R, B, S, D, E>(handler: F) -> Result<(), Error>
```
jlizen, line 168: Same feedback on `run_extended` / `run_maybe_concurrent`, and also nudging regular `run` to use this + warning that it doesn't respect the concurrency env

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+where
+    S: Service<LambdaInvocation, Response = (), Error = BoxError> + Clone + Send + 'static,
+    S::Future: Send,
+{
+    /// Start the runtime in concurrent mode when configured for Lambda managed-concurrency.
+    ///
+    /// If `AWS_LAMBDA_MAX_CONCURRENCY` is not set or is `<= 1`, this falls back to the
+    /// sequential `run_with_incoming` loop so that the same handler can run on both
+    /// classic Lambda and Lambda Managed Instances.
+    pub async fn run_concurrent(self) -> Result<(), BoxError> {
```
jlizen, line 169: Same feedback as these as the other `*_concurrent`

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+                    if first_error.is_none() {
+                        first_error = Some(err);
+                    }
+                }
+            }
+        }
+
+        match first_error {
+            Some(err) => Err(err),
+            None => Ok(()),
```
jlizen: (Couldn't leave feedback on `fn run()`):

We should probably return an error or log an error message or something if `run()` is invoked with concurrency set, no? Currently it silently ignores it which is fairly confusing.

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+    /// Start the runtime in concurrent mode when configured for Lambda managed-concurrency.
+    ///
+    /// If `AWS_LAMBDA_MAX_CONCURRENCY` is not set or is `<= 1`, this falls back to the
+    /// sequential `run_with_incoming` loop so that the same handler can run on both
+    /// classic Lambda and Lambda Managed Instances.
+    pub async fn run_concurrent(self) -> Result<(), BoxError> {
+        if self.concurrency_limit > 1 {
+            Self::run_concurrent_inner(self.service, self.config, self.client, self.concurrency_limit).await
+        } else {
+            let incoming = incoming(&self.client);
```
jlizen, line 177: Would this merit a trace or debug-level log? Since if somebody is using `run_concurrent`, there is a decent chance they meant to set the ENV and could use help debugging

examples/basic-lambda-concurrent/src/main.rs
```diff
@@ -0,0 +1,74 @@
...
+}
+
+#[tokio::main]
+async fn main() -> Result<(), Error> {
+    // required to enable CloudWatch error logging by the runtime
+    tracing::init_default_subscriber();
+
+    let func = service_fn(my_handler);
+    if let Err(err) = lambda_runtime::run_concurrent(func).await {
+        eprintln!("run error: {:?}", err);
```
jlizen: nit: a bit odd to both init tracing, but then emit an error with `eprintln!`

lambda-runtime-api-client/src/lib.rs
```diff
@@ -111,6 +119,14 @@ impl ClientBuilder {
         Self { uri: Some(uri), ..self }
     }
 
+    /// Provide a pool size hint for the underlying Hyper client.
+    pub fn with_pool_size(self, pool_size: usize) -> Self {
```
jlizen, line 127: Nit: Is it worth mentioning that if using concurrency, this should be at least equal to the maximum concurrency?

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+    async fn run_concurrent_inner(
+        service: S,
+        config: Arc<Config>,
+        client: Arc<ApiClient>,
+        concurrency_limit: u32,
+    ) -> Result<(), BoxError> {
+        let limit = concurrency_limit as usize;
+
+        let mut workers = FuturesUnordered::new();
+        for _ in 1..limit {
```
jlizen, line 203: Nit: Might be nice to have a comment here like `// save a few atomic increments by pushing the last loop with owned values

(Confused me first read until I read onwards to line 197)

Related: Maybe we could have a unit test probing we are properly managing concurrency? Doesn't seem too difficult when module-local since we can just assert against the refcount of eg `Arc<ApiClient>`?

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+    /// Concurrent processing using N independent long-poll loops (for Lambda managed-concurrency).
+    async fn run_concurrent_inner(
+        service: S,
+        config: Arc<Config>,
+        client: Arc<ApiClient>,
+        concurrency_limit: u32,
+    ) -> Result<(), BoxError> {
+        let limit = concurrency_limit as usize;
+
+        let mut workers = FuturesUnordered::new();
```
jlizen: FuturesUnordered seem a bit heavyweight here compared to [join_all](https://docs.rs/futures/latest/futures/future/fn.join_all.html), since we have a fixed (fairly small) set of futures, and they should be rarely woken to the point where we don't need the bookkeeping to only poll specific futures when one wakes?

Should be a bit smaller of a memory footprint that way.

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+        // API client failures, runtime panics, etc.
+        let mut first_error: Option<BoxError> = None;
+        let mut remaining_workers = limit;
+        while let Some(result) = futures::StreamExt::next(&mut workers).await {
+            remaining_workers = remaining_workers.saturating_sub(1);
+            match result {
+                Ok(Ok(())) => {
+                    // `concurrent_worker_loop` runs indefinitely, so an Ok return indicates
+                    // an unexpected worker exit; we still decrement because the task is gone.
+                    warn!(
```
jlizen: Minor: If this is indicates a system issue, why not error log level?

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+                        remaining_workers,
+                        "Concurrent worker exited cleanly (unexpected - loop should run forever)"
+                    );
+                    if first_error.is_none() {
+                        first_error = Some(Box::new(io::Error::other(
```
jlizen: This is extra confusing since we only get an eventual error related to the clean exit if there were no prior individual workers with errors. Probably we want to always surface this case?

Also, this seems inaccurate:
> all concurrent workers exited cleanly (unexpected - loop should run forever

Since we only know that ONE worker exited, here?

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+        // Track the first infrastructure error. A single worker failing should
+        // not terminate the whole runtime (LMI keeps running with the remaining
+        // healthy workers). We only return an error once there are no workers
+        // left (i.e., we cannot keep at least 1 worker alive).
+        //
+        // Note: Handler errors (Err returned from user code) do NOT trigger this.
+        // They are reported to Lambda via /invocation/{id}/error and the worker
+        // continues. This only captures unrecoverable runtime failures like
+        // API client failures, runtime panics, etc.
+        let mut first_error: Option<BoxError> = None;
```
jlizen: This `first_error` piece is fairly confusing to me. Why would we not want to surface all errors? Ie, collect into `Vec<BoxError>` that we implement `std::error::Error` and `Display` for?

That would have the side benefit of letting us both emit an error if all workers exit cleanly, but still preserve any earlier single-worker errors.

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+        let mut remaining_workers = limit;
+        while let Some(result) = futures::StreamExt::next(&mut workers).await {
```
jlizen, line 219: Is there a way to retrieve the [task Id](https://docs.rs/tokio/latest/tokio/task/struct.Id.html) to enrich these error messages with? I guess we've already finished the `JoinHandle` by the time we have a result. Maybe the Return could be `Result<(Id, (Id, BoxError)>`? And then `JoinError` has an `id` getter in case of a task panic?

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+                Err(join_err) => {
+                    let err: BoxError = Box::new(join_err);
+                    error!(error = %err, remaining_workers, "Concurrent worker panicked");
+                    if first_error.is_none() {
+                        first_error = Some(err);
+                    }
+                }
+            }
+        }
+
```
jlizen, line 254: It seems like THIS is the place where we want to capture whether ALL workers exited cleanly (though that would be a bit tricky to capture in current impl).

The nicer way, might be to model the collected errors as:
```
enum WorkerError {
    CleanExit(Id),
    Failure((Id, BoxError))
}
```

And that way we could flatten down to into either a BoxError representing "all workers exited cleanly" versus "these workers exited cleanly, and these ones had these errors".

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+        match first_error {
```
jlizen: In the current impl, `None` should essentially never be reachable, right? Since even if we have no task-specific errors, we instead create an error out of a clean exit?

If the desired API is to return `Ok(())` on all workers cleanly exiting, then we could use the enum mentioned above to gather that info, and emit an error log line if all exited cleanly, but then drop the error.

lambda-runtime/src/runtime.rs
```diff
@@ -133,6 +153,95 @@ impl<S> Runtime<S> {
...
+        service: S,
+        config: Arc<Config>,
+        client: Arc<ApiClient>,
+        concurrency_limit: u32,
+    ) -> Result<(), BoxError> {
+        let limit = concurrency_limit as usize;
+
+        let mut workers = FuturesUnordered::new();
+        for _ in 1..limit {
+            workers.push(tokio::spawn(concurrent_worker_loop(
```
jlizen: Would it be worth instrumenting this future with a span containing the taskId? Seems pretty useful for debug

lambda-runtime/src/runtime.rs
```diff
@@ -233,6 +319,80 @@ fn incoming(
...
+        return Ok(());
+    }
+
+    // Build the invocation such that it can be sent to the service right away
+    // when it is ready
+    let body = incoming.collect().await?.to_bytes();
+    let context = Context::new(invoke_request_id(&parts.headers)?, config.clone(), &parts.headers)?;
+    let invocation = LambdaInvocation { parts, body, context };
+
+    if set_amzn_trace_env {
```
jlizen, line 480: Nit: I don't love the extra branching here since this is pure overhead for the non-concurrent case. But, I guess branch prediction will probably make this pointless to optimize. Don't think it is worth a bunch of code duplication for.

lambda-runtime/src/runtime.rs
```diff
@@ -233,6 +319,80 @@ fn incoming(
...
+        // Setup Amazon's default tracing data
+        amzn_trace_env(&invocation.context);
+    } else {
+        // Inform users that X-Ray is available via context, not env var, in concurrent mode.
+        XRAY_LOGGED.get_or_init(|| {
```
jlizen: The OnceLock + get_or_init every invocation seems like unnecessary overhead / complexity. Why can't we just display this message once, at the start of `run_concurrent`?

lambda-runtime/src/runtime.rs
jlizen: (No action needed here, unless you feel particularly inspired, fine to resolve this)

Makefile
alessandrobologna: Good point, thank you. I added a short note in the README under RIE testing showing `make test-rie-lmi EXAMPLE=basic-lambda-concurrent`

lambda-http/src/lib.rs
alessandrobologna: Thanks! I added a doc warning on `run()` and a fail‑fast guard when `AWS_LAMBDA_MAX_CONCURRENCY` is set. I kept the `*_concurrent` naming for now to limit "code churn", and also because `_extended` seems to me a little vague. But I am open to change it if there's consensus on renaming all of them to what you suggested.
I have also added a note to the `run()` docstring to make it more clean that `run_concurrent` is preferred if the handler can satisfy `Clone`

lambda-http/src/lib.rs
alessandrobologna: Correct, this was the existing behavior. It could fail if the `xray_trace_id` contains invalid header characters, but it should always be valid so I didn't add logging to avoid noise. We could add a debug level log though. Let me know.

lambda-http/src/streaming.rs
alessandrobologna: Very true! :) I dropped `BoxCloneService` and switched the concurrent streaming path to a cloneable `MapRequest/MapResponse` pipeline (similar to `into_stream_service`) so now it returns a generic cloneable service without the extra allocation/indirection. Thank you for pointing that out.

lambda-http/src/streaming.rs
alessandrobologna: Yes, that makes sense. I added `Clone` for `StreamAdapter` when the inner service is Clone, so custom stacks can still work with concurrent handlers.

lambda-http/src/streaming.rs
alessandrobologna: Same as in the non-streaming case, keeping the `_concurrent` naming for now, but happy to rename it. I also added a doc warning and the runtime now fails fast if `AWS_LAMBDA_MAX_CONCURRENCY` is set and the non‑concurrent entrypoint is used.

lambda-runtime/src/lib.rs
alessandrobologna: Agree, happy to rename if there's consensus to change it to `run_extended` or `run_maybe_concurrent`.

lambda-runtime/src/runtime.rs
alessandrobologna: Yup.

lambda-runtime/src/runtime.rs
alessandrobologna: Yes, as mentioned above, I added a fail‑fast guard in `Runtime::run()` when `AWS_LAMBDA_MAX_CONCURRENCY` is set, and documented it on the public `run()` APIs so it’s not silently ignored anymore.

lambda-runtime/src/runtime.rs
alessandrobologna: I have added `trace!("Concurrent mode: _X_AMZN_TRACE_ID is not set; use context.xray_trace_id")`. I can bump it to debug if you prefer.

examples/basic-lambda-concurrent/src/main.rs
alessandrobologna: Fixed it.

lambda-runtime-api-client/src/lib.rs
alessandrobologna: Added a note to `with_pool_size` to that extent.

lambda-runtime/src/runtime.rs
alessandrobologna: I added the clarifying comment. I didn't add a refcount test since I am concerned that it could be pretty brittle (ties us to current cloning/Arc behavior), and we already have a functional concurrency test in place. If you feel strongly, I can add a simple clone‑count test, but I'd prefer to keep it behavior‑focused.

lambda-runtime/src/runtime.rs
alessandrobologna: Good point, `join_all` would be a bit lighter on memory, especially with a fixed small set of workers. I have kept `FuturesUnordered` at least for now so that we can observe/log worker exits as they happen (as far as I know,  `join_all` only returns when _all workers_ are done) 
Given that N is small, the extra overhead should be minimal, but if you'd rather optimize for footprint and only log at the end, I can switch to `join_all`

lambda-runtime/src/runtime.rs
alessandrobologna: Changed to `error!()`

lambda-runtime/src/runtime.rs
alessandrobologna: Agreed. I changed this so clean exits are always recorded (even if other workers already errored), and I fixed the wording to be per‑worker rather than "all workers’". Now the aggregate error includes clean exits alongside any other worker errors.

lambda-runtime/src/runtime.rs
alessandrobologna: Also agreed. I switched this to collect all worker errors into a `Vec<WorkerError>` and return a `ConcurrentWorkerErrors` aggregate (Display/Error). Clean exits are recorded too, so they’re surfaced alongside any other failures

lambda-runtime/src/runtime.rs
alessandrobologna: Great idea. I now capture `JoinHandle::id()` at spawn time and propagate it with the join result, so logs and the aggregate error include the task id. I also add a per‑worker `info_span` with the task id that instruments each invocation

lambda-runtime/src/runtime.rs
alessandrobologna: Yes, thank you for the suggestion. I modeled this as an explicit enum (`WorkerError::CleanExit(Id)` / `WorkerError::Failure(Id, BoxError))` as per your suggestion, and aggregate those into a single error. The formatter now distinguishes ‘all clean’ vs ‘clean + failures’

lambda-runtime/src/runtime.rs
alessandrobologna: Yes, I am now using the `WorkerError` and aggregate everything. 

For the other point, I currently treat all workers exiting (clean or not) as an error because then the runtime has no workers left and can't serve any more invocations. Returning `Ok(())` here would make the runtime exit quietly and hide the failure. But I can change it if you prefer.

lambda-runtime/src/runtime.rs
alessandrobologna: Yes, I have added a span on the worker loop with the Tokio task id and instrumented the per-invocation processing with it.

lambda-runtime/src/runtime.rs
alessandrobologna: I removed the `XRAY_LOGGED.get_or_init` branch and now only call `amzn_trace_env()` when `set_amzn_trace_env` is true. The concurrent message is logged once at startup, as per your suggestion, so the per‑invocation path is simpler

lambda-runtime/src/runtime.rs
alessandrobologna: Agree.

lambda-http/src/lib.rs
jlizen: `_concurrent` is probably ok... ultimately future features that require `Clone` will likely involve concurrency, so the naming probably makes sense...

lambda-http/src/lib.rs
jlizen: I'm fine to leave this out to avoid scope creep

lambda-runtime/src/runtime.rs
jlizen: Good point, I agree that surfacing the telemetry worker by worker merits the extra footprint for now.

lambda-runtime/src/runtime.rs
jlizen: This new log seems fine at trace, but I more meant a debug log indicating that we are falling back to sequential polling.

lambda-runtime/src/runtime.rs
jlizen: I'm fine with leaving out the test for now. If we did want to do it, I agree that refcount handling is probably too brittle. We could do something like pass a handler that writes the current tokio task id to a mutex-guarded hashset.

But anyway, fine to skip this.

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
         let limit = concurrency_limit as usize;
 
-        let mut workers = FuturesUnordered::new();
+        // Use FuturesUnordered so we can observe worker exits as they happen,
+        // rather than waiting for all workers to finish (join_all).
+        let mut workers: FuturesUnordered<WorkerJoinFuture> = FuturesUnordered::new();
+        let spawn_worker = |service: S, config: Arc<Config>, client: Arc<ApiClient>| -> WorkerJoinFuture {
+            let handle = tokio::spawn(concurrent_worker_loop(service, config, client));
+            let task_id = handle.id();
+            Box::pin(async move { (task_id, handle.await) })
```
jlizen: Why the extra `Box::pin()` here? The `tokio::task::Id` is `Unpin`? Is there away to avoid this extra allocation / indirection?

Can we just get away with using an unnameable type like we did previously?

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
-                client.clone(),
-            )));
+            workers.push(spawn_worker(service.clone(), config.clone(), client.clone()));
         }
-        workers.push(tokio::spawn(concurrent_worker_loop(service, config, client)));
+        workers.push(spawn_worker(service, config, client));
 
-        // Track the first infrastructure error. A single worker failing should
-        // not terminate the whole runtime (LMI keeps running with the remaining
+        // Track worker exits across tasks. A single worker failing should not
```
jlizen, line 208: Thanks for the detailed comment on this passage

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
                     // `concurrent_worker_loop` runs indefinitely, so an Ok return indicates
                     // an unexpected worker exit; we still decrement because the task is gone.
-                    warn!(
+                    let clean_exit_msg = "Concurrent worker exited cleanly (unexpected - loop should run forever)";
+                    error!(
+                        task_id = %task_id,
                         remaining_workers,
-                        "Concurrent worker exited cleanly (unexpected - loop should run forever)"
+                        "{}",
+                        clean_exit_msg
```
jlizen: nit: i would probably just put this string slice inline since we only use it once

lambda-runtime/src/runtime.rs
jlizen: New impl looks great!

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
+            }
+        }
+
+        if failures.is_empty() && !clean.is_empty() {
+            write!(
+                f,
+                "all concurrent workers exited cleanly (unexpected - loop should run forever)"
+            )?;
+            for task_id in clean {
+                write!(f, " [task {task_id}]")?;
```
jlizen: nit: this is a bit odd, i would have expected something like `[task_1, task_2, ..]`

Might be easier with `write!(f, " {}", clean.join(", "))` (and more efficient anyway, fewer write calls, though probably we aren't perf sensitive in this code path)

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
+            }
+            return Ok(());
+        }
+
+        write!(f, "concurrent workers exited unexpectedly")?;
+        if !clean.is_empty() {
+            write!(f, "; clean exits:")?;
+            for task_id in clean {
+                write!(f, " [task {task_id}]")?;
```
jlizen: Same nit

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
+            }
+        }
+        if !failures.is_empty() {
+            write!(f, "; failures:")?;
+            for (task_id, err) in failures {
+                write!(f, " [task {task_id}] {err}")?;
```
jlizen: Same nit.

Also, I wonder, would it be worth something like: `[{"id": "foo", "err": "foo" },..]` to make it easier to query against in log insights or similar? Even if you need to regex match only the `[..]` part of the message, having some signpost that there is an error message is pretty nice.

lambda-runtime/src/runtime.rs
```diff
@@ -186,72 +178,144 @@ where
...
 impl<S> Runtime<S>
 where
     S: Service<LambdaInvocation, Response = (), Error = BoxError>,
 {
     /// Start the runtime and begin polling for events on the Lambda Runtime API.
+    ///
+    /// If `AWS_LAMBDA_MAX_CONCURRENCY` is set, this returns an error because it does not enable
+    /// concurrent polling. Use [`Runtime::run_concurrent`] instead.
     pub async fn run(self) -> Result<(), BoxError> {
+        if let Some(raw) = concurrency_env_value() {
```
jlizen, line 342: So perf impact here for non-concurrent path is, exactly one environment variable check? That seems fine, good tradeoff for footgun avoidance.

lambda-runtime/src/runtime.rs
```diff
@@ -332,21 +396,29 @@ fn max_concurrency_from_env() -> Option<u32> {
...
+    env::var("AWS_LAMBDA_MAX_CONCURRENCY").ok()
+}
+
 async fn concurrent_worker_loop<S>(mut service: S, config: Arc<Config>, client: Arc<ApiClient>) -> Result<(), BoxError>
 where
     S: Service<LambdaInvocation, Response = (), Error = BoxError>,
     S::Future: Send,
 {
+    let task_id = tokio::task::id();
+    let span = info_span!("concurrent_worker_loop", task_id = %task_id);
```
jlizen: nit: `concurrent_worker_loop` is a bit verbose to see in all inner log lines, i'd suggest `worker` probably

lambda-runtime/src/runtime.rs
```diff
@@ -332,21 +396,29 @@ fn max_concurrency_from_env() -> Option<u32> {
...
     loop {
         let event = match next_event_future(client.clone()).await {
```
jlizen: Any reason not to instrument the hyper future as well? I seem to recall hyper has some sort of tracing feature support that might benefit from task id context?

jlizen: Side question - do we really need to clone the client here? Could we just pass it by reference and save the atomic increment? `call()` just requires `&self` right?

lambda-runtime/src/runtime.rs
jlizen: LGTM!

jlizen (changes requested):
Thanks for all the tweaks! New error handling, removal of `XRAY_LOGGED`, `into_stream_service_cloneable()`, etc, look great!

There are a few nits that are totally up to you.

Main load bearing feedback has to do with the usage of `Box::pin()` for the worker future. If it it turns out to be tricky to remove, that's fine / just say so, we can punt that as it is something we can change later.

I also think having a debug log indicating when sequential fallback is used in `run_concurrent`, would be pretty handy ([ref](https://github.com/aws/aws-lambda-rust-runtime/pull/1067/files/949c00231503aa8207772d642cfec778ac3a022d..3ddc948242f26201a844f22543768ead7423cbdc#r2648590077))

jlizen:
The CI failure related to MSRV seems unrelated... I'll fix separately

Ref: https://github.com/aws/aws-lambda-rust-runtime/issues/1077

lambda-runtime/src/runtime.rs
alessandrobologna: Done. I added a `debug!` log when `run_concurrent` falls back to sequential polling.

lambda-runtime/src/runtime.rs
alessandrobologna: Good point. I used `Box::pin` originally just to erase the unnameable async move type for `FuturesUnordered`, but, as you pointed out, that adds an allocation. I switched to spawning a task that returns `(task_id, Result<(), BoxError>)`, so the set holds a concrete `JoinHandle` and there’s no `Box::pin`. For panics I use `JoinError::id()` a little below.

lambda-runtime/src/runtime.rs
alessandrobologna: Agreed. I switched the aggregate error formatting to just be valid JSON so it's easy to query in Log Insights/analytics tools (if the caller logs/prints it). The per‑worker logs are unchanged.

lambda-runtime/src/runtime.rs
alessandrobologna: Yes, same reasoning here. I ended up switching the aggregate error formatting to valid JSON to keep it simple and queryable, rather than bespoke formatting

lambda-runtime/src/runtime.rs
alessandrobologna: I have changed it just "worker"

lambda-runtime/src/runtime.rs
alessandrobologna: Agreed, I wrapped the `/next` request future in the worker span so it inherits the `task_id` context. I didn't enable hyper's tracing feature (to keep deps minimal, and it seems that tracing in [1.0 is still unstable](https://github.com/hyperium/hyper/issues/2498)), but this gives us span context for the request. 

Also agreed, I have removed the cloning and passing `&ApiClient` into `next_event_future` now.

lambda-runtime/src/runtime.rs
```diff
@@ -253,14 +258,26 @@ enum WorkerError {
...
-type WorkerJoinResult = (tokio::task::Id, Result<Result<(), BoxError>, tokio::task::JoinError>);
-type WorkerJoinFuture = BoxFuture<'static, WorkerJoinResult>;
-
 #[derive(Debug)]
 struct ConcurrentWorkerErrors {
     errors: Vec<WorkerError>,
 }
 
+#[derive(Serialize)]
+struct ConcurrentWorkerErrorsPayload<'a> {
```
jlizen, line 277: Ah yeah, this is a lot cleaner, nice.

jlizen:
CI will keep failing until you pick up #1078 on your branch - probably good to validate the full successful runs, when you have a moment to rebass.

GradExMachina:
Why are we removing xray updates?

alessandrobologna:
> Why are we removing xray updates?

@FullyTyped are you referring to removing [`update_xray_trace_id_header`](https://github.com/aws/aws-lambda-rust-runtime/pull/1067/changes#diff-8d6dff89da49a82ad3db479309029ace94da82f17d113341a28d418f45ebbd36L173)? That is replaced by [`update_xray_trace_id_header_from_context`](https://github.com/aws/aws-lambda-rust-runtime/pull/1067/changes#diff-c62692e6b817476306b4d6966f541eec77a1768674e4dad5cf7f10c4faababa4R245) because the `_X_AMZN_TRACE_ID` env var cannot be used to set a per-request trace id

alessandrobologna:
> CI will keep failing until you pick up #1078 on your branch - probably good to validate the full successful runs, when you have a moment to rebass.

@jlizen:

Rebased this branch onto current upstream/main (including #1078).

  Re‑ran:

  - cargo +nightly fmt --all
  - cargo clippy --workspace --all-features -- -D warnings
  - cargo +stable test --all-features -p lambda_runtime_api_client -p lambda_runtime -p lambda_http

GradExMachina:
Generally looks good to me.

First this should be behind an experimental flag to indicate that it is still experimental.

Second we need to benchmark and test the changes thoroughly. This can be done before the next release.

Finally, we probably want to review error handling as the error shouldn't be printed before exiting, but should be sent to data plane along with a "Runtime.ExitError", and the execution should just terminate there with Ok(()). 

We can do this in a follow-up task.

alessandrobologna:
@FullyTyped thanks! 

  - Flag: I can put this behind an "experimental" flag. Do you prefer a Cargo feature (e.g. `experimental-concurrency`) or a runtime env flag? I can wire it through `lambda-runtime` and `lambda-http` and update docs/examples.
  - Benchmarks: agreed. I’ve done some light benchmarking on AWS in [my demo repo](https://github.com/alessandrobologna/rust-lmi-demo), but it probably would need to be more scientific than that :).
  - Exit errors: right now I log per‑worker errors and return `Err`. If you want a runtime‑level failure reported to the Runtime API (e.g. `Runtime.ExitError`) and then exit `Ok(())`, I can rework it. I know the per‑invocation endpoint `/2018-06-01/runtime/invocation/{id}/error`, but is there a documented init/exit error endpoint/payload that I should use?

GradExMachina:
Re flag: a cargo flag works!

Re benchmarking: I will handle the benchmarking in the second week of January ☺️

Re errors: there is the /init/error api that we are meant to use but I would rather not do those changes now as they also involve the lambda-on-demand code path, and have the PR handle just LMI code path.

jlizen:
One question I have is, is it really desirable to have workers stay stopped when they crash? For a regular runtime, the runtime exits, so it's not an issue. But, here, could we wind up in a state where a managed instance is permanently below the desired concurrency?

Does the lambda control plane somehow track worker exits or current vs desired concurrency and trigger the entire runtime to reboot at some point if it is below target?

If not, would it make more sense to either do some sort of graceful shutdown after the first worker exits, or restart the worker?

(Probably a question for @FullyTyped as I imagine this was already considered in the LMI design?)

GradExMachina:
Indeed Control Plane has no internal understanding of what concurrency means in this context as every implementation can handle it however it desires.

I need to review the documents again, but afaik it should be a graceful exit into a complete termination. Data plane will observe decrease in throughput which will trigger auto-scaling to provision more instances until the current sandbox is retired.

alessandrobologna:
There was this conversation that may be relevant: https://github.com/aws/aws-lambda-rust-runtime/issues/1065#issuecomment-3672051824

I think the approach chosen was to possibly improve on the failure handling in a second pass. Maybe then that could be a possible approach.

GradExMachina:
I think this approach is fine for now, and we can make adjustments as needed down the road.

alessandrobologna:
@FullyTyped / @jlizen / @bnusunny  just a quick clarification on the experimental flag. Do you want full gating (default build = current main behavior), e.g.:

```rust
// main behavior (feature off)
#[cfg(not(feature = "experimental-concurrency"))]
fn update_xray_trace_id_header(headers: &mut HeaderMap) {
    if let Ok(xray_trace_id) = env::var("_X_AMZN_TRACE_ID") {
        if let Ok(header_value) = HeaderValue::from_str(&xray_trace_id) {
            headers.insert(HeaderName::from_static("x-amzn-trace-id"), header_value);
        }
    }
}

// concurrent behavior (feature on)
#[cfg(feature = "experimental-concurrency")]
fn update_xray_trace_id_header_from_context(headers: &mut HeaderMap, context: &Context) {
    if let Some(trace_id) = context.xray_trace_id.as_deref() {
        if let Ok(header_value) = HeaderValue::from_str(trace_id) {
            headers.insert(HeaderName::from_static("x-amzn-trace-id"), header_value);
        }
    }
}

// call site
#[cfg(feature = "experimental-concurrency")]
update_xray_trace_id_header_from_context(event.headers_mut(), &context);

#[cfg(not(feature = "experimental-concurrency"))]
update_xray_trace_id_header(event.headers_mut());
```

Or is it OK to only gate the concurrent APIs/paths, and keep small fixes like the X‑Ray header source unconditional, i.e.:

```rust
// partial gating example
fn update_xray_trace_id_header(headers: &mut HeaderMap, context: &Context) {
    if let Some(trace_id) = context.xray_trace_id.as_deref() {
        if let Ok(header_value) = HeaderValue::from_str(trace_id) {
            headers.insert(HeaderName::from_static("x-amzn-trace-id"), header_value);
        }
    }
}

#[cfg(feature = "experimental-concurrency")]
pub async fn run_concurrent(...) { ... }
```
  
The reason why I'm asking is that full gating keeps default behavior identical to main (very low risk) but adds extra cfg plumbing and complexity, while partial gating keeps the PR simpler but changes behavior even when the feature isn’t enabled. 
Let me know which way you prefer it. Thanks!

GradExMachina:
Because in the rust runtime users have explicit control over version, it is safe to do a partial gating at the top level api, and it is only needed for the concurrent api in my view. 

If Jess and Harold don't have objections, I'd say we can go ahead with partial gating and then I'll merge.

jlizen:
Gating only the concurrent API seems fine to me.

alessandrobologna:
Pushed commit 8ed50a3  that gates the concurrent APIs behind the new `experimental-concurrency` feature. Default behavior stays the same apart from the existing `run()` guard and X‑Ray header update. Also fixes default cargo clippy warnings by gating concurrent‑only imports, and updates docs and the concurrent example dependency.

GradExMachina:
Thanks a lot for your work! Much appreciated!

PR #1068, opened 2025-12-12
fix(lambda-runtime): ensure if a client disconnects while streaming, the handler does not panic

zakhenry:
✍️ *Description of changes:*
Fixes unwrap causing panic when client disconnects from a streaming response

### DIFF
--- a/lambda-runtime/src/requests.rs
+++ b/lambda-runtime/src/requests.rs
@@ -3,7 +3,7 @@ use bytes::Bytes;
 use http::{header::CONTENT_TYPE, Method, Request, Uri};
 use lambda_runtime_api_client::{body::Body, build_request};
 use serde::Serialize;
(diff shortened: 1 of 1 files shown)


ghost:
Hi, thanks for the PR, I will be checking it shortly. In the mean time, could you tell me whether an incident happened that prompted it?

ghost:
So after further inspection, indeed the background thread panics, but it isn't evident on the test by itself.

```
thread 'requests::tests::streaming_send_data_error_is_ignored' (4597434) panicked at lambda-runtime/src/requests.rs:112:65:
called `Result::unwrap()` on an `Err` value: Error { inner: ChannelClosed }
```

ghost:
Thanks for the PR! Nice catch!

I have updated the test to make it fail without the unwraps and removed an accidental clone there.

PR #1083, opened 2026-01-06
Enable automated releases

FullyTyped:
📬 *Issue #, if available:*

✍️ *Description of changes:*
Adds support for automated releases

### DIFF
--- a/.github/workflows/release-plz.yml
+++ b/.github/workflows/release-plz.yml
@@ -1,4 +1,4 @@
-name: Release-plz
+name: Publish release
 
 on:
   push:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-plz.yml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  release-plz-release:
+    name: Release-plz release
+    runs-on: ubuntu-latest
+    if: ${{ github.repository_owner == 'aws' }}
```
maxday, line 12: What's the purpose of this check? Making sure that the push is not coming from a forked repo?

GradExMachina: Yups

GradExMachina: Well, it's so that it doesn't run on forks.

But even if somebody removed it, it still won't pus to crates io.

jlizen: Yeah just a convenience to avoid ci failures on forks. It would be great to add similar for the integration test workflow.

release-plz.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+semver_check = true
```
jlizen, line 2: This is somewhat redundant with our cargo semver check workflow right? I don't mind keeping it though for defense in depth.

PR #1086, opened 2026-01-13
test: verify request-ID isolation in concurrent exec

FullyTyped:
Adds test to ensure request IDs are properly isolated in concurrent
Lambda execution with structured logging. Verifies each concurrent
invocation maintains unique request ID context.

📬 *Issue #, if available:*

✍️ *Description of changes:*

### DIFF
--- a/lambda-runtime/src/runtime.rs
+++ b/lambda-runtime/src/runtime.rs
@@ -922,12 +922,14 @@ mod endpoint_tests {
         tracing::subscriber::set_global_default(subscriber).unwrap();
 
         let request_count = Arc::new(AtomicUsize::new(0));
+        let done = Arc::new(tokio::sync::Notify::new());
         let listener = TcpListener::bind("127.0.0.1:0").await?;
         let addr = listener.local_addr()?;
         let base: http::Uri = format!("http://{addr}").parse()?;
 
         let server_handle = {
             let request_count = request_count.clone();
+            let done = done.clone();
             tokio::spawn(async move {
                 loop {
                     let (tcp, _) = match listener.accept().await {
@@ -936,8 +938,10 @@ mod endpoint_tests {
                     };
 
                     let request_count = request_count.clone();
+                    let done = done.clone();
                     let service = service_fn(move |req: Request<Incoming>| {
                         let request_count = request_count.clone();
+                        let done = done.clone();
                         async move {
                             let (parts, body) = req.into_parts();
                             if parts.method == Method::POST {
@@ -957,6 +961,7 @@ mod endpoint_tests {
                                         .unwrap();
                                     return Ok::<_, Infallible>(res);
                                 } else {
+                                    done.notify_one();
                                     let res = Response::builder()
                                         .status(StatusCode::NO_CONTENT)
                                         .body(Full::new(Bytes::new()))
@@ -1023,14 +1028,9 @@ mod endpoint_tests {
 
         let runtime_handle = tokio::spawn(async move { runtime.run_concurrent().await });
 
-        loop {
-            tokio::time::sleep(Duration::from_millis(100)).await;
-            let count = request_count.load(Ordering::SeqCst);
-            if count >= 300 {
-                tokio::time::sleep(Duration::from_millis(500)).await;
-                break;
-            }
-        }
+        done.notified().await;
+        // Give handlers time to complete after server signals done
+        tokio::time::sleep(Duration::from_millis(500)).await;
 
         runtime_handle.abort();
         server_handle.abort();


lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,188 @@ mod endpoint_tests {
...
+                        );
+                    }
+                }
+                event.record(&mut FieldVisitor(&mut fields));
+                self.logs.lock().unwrap().push(fields);
+            }
+        }
+
+        let log_capture = LogCapture::new();
+        let subscriber = tracing_subscriber::registry().with(log_capture.clone());
```
GradExMachina: We hijack the logging impl to keep track of all the emitted fields.

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,188 @@ mod endpoint_tests {
...
+
+                    let io = TokioIo::new(tcp);
+                    tokio::spawn(async move {
+                        let _ = ServerBuilder::new(TokioExecutor::new())
+                            .serve_connection(io, service)
+                            .await;
+                    });
+                }
+            })
+        };
```
GradExMachina, line 997: We create a small emulator for the client to communicate with.

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,188 @@ mod endpoint_tests {
...
+        let runtime_handle = tokio::spawn(async move { runtime.run_concurrent().await });
+
+        loop {
+            tokio::time::sleep(Duration::from_millis(100)).await;
+            let count = request_count.load(Ordering::SeqCst);
+            if count >= 300 {
+                tokio::time::sleep(Duration::from_millis(500)).await;
+                break;
+            }
+        }
```
GradExMachina: Issue the requests, the tracker will collect the request-ids.

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,188 @@ mod endpoint_tests {
...
+                observed_id.starts_with("test-request-"),
+                "Request ID should match pattern: {}",
+                observed_id
+            );
+            assert!(
+                seen_ids.insert(observed_id.clone()),
+                "Request ID should be unique: {}",
+                observed_id
+            );
+        }
```
GradExMachina, line 1079: Verify uniqueness. If there's contamination, we won't observe uniqueness.

FullyTyped (commented):
I took this route rather than an integration test via SAM since it was harder to setup via SAM, and claude has done a good enough job setting up a small and self-contained emulator.

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,239 @@ mod endpoint_tests {
...
+    async fn test_concurrent_structured_logging_isolation() -> Result<(), Error> {
+        use std::{
+            collections::{HashMap, HashSet},
+            sync::Mutex,
+        };
+        use tracing::{info, subscriber::set_global_default};
+        use tracing_subscriber::{layer::SubscriberExt, Layer};
+
+        #[derive(Clone)]
+        struct LogCapture {
```
jlizen: Is there a reason for the handrolled layer rather than using https://docs.rs/tracing-test/latest/tracing_test/ ?

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,239 @@ mod endpoint_tests {
...
+        {
+            fn on_new_span(
+                &self,
+                attrs: &tracing::span::Attributes<'_>,
+                id: &tracing::Id,
+                _ctx: tracing_subscriber::layer::Context<'_, S>,
+            ) {
+                if attrs.metadata().name() == "Lambda runtime invoke" {
+                    let mut fields = HashMap::new();
+                    struct FieldVisitor<'a>(&'a mut HashMap<String, String>);
```
jlizen: If we do need to handroll our test layer, is there a reason to define `FieldVisitor` inline in each method? It seems to be the same for each?

lambda-runtime/src/runtime.rs
jlizen: Nice, this will be useful for other things too, has long irritated me that we don't have a great way of testing e2e.

Would it be worth extracting into a helper instead of keeping it inline in the test?

lambda-runtime/src/runtime.rs
jlizen: Is there a reason for handrolling this compared to just using [tracing_test](https://crates.io/crates/tracing-test/)?

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,239 @@ mod endpoint_tests {
...
+
+            // Verify span request ID matches logged request ID
+            assert_eq!(
+                observed_id, span_request_id,
+                "Span request ID should match logged request ID: span={}, logged={}",
+                span_request_id, observed_id
+            );
+        }
+
+        println!(
```
jlizen: nit: i could do without this, libtest harness anyway has a success output

lambda-runtime/src/runtime.rs
```diff
@@ -908,4 +908,239 @@ mod endpoint_tests {
...
+                log_stream: "test_stream".to_string(),
+                log_group: "test_log".to_string(),
+            }),
+            service,
+            concurrency_limit: 3,
+        };
+
+        let runtime_handle = tokio::spawn(async move { runtime.run_concurrent().await });
+
+        loop {
```
jlizen: Minor: This is a bit odd - if all we want to do is wait until all 300 requests are done, why not just use a oneshot channel that we write to on our 300th request? (And store it as an `Option<Sender>` in the server so that we can consume it)

jlizen (approved):
Thanks for this! Approving because the test does probe the things it wants to.

I had some non-blocking suggestions for simplification / making it more reusable.

Up to you if you think it is worth the trouble. Might be worth at least leaving some TODO comments so that if somebody else needs N+1 tracing/runtime simulation test, they have a breadcrumb to follow?

lambda-runtime/src/runtime.rs
jlizen: This note might be a nice comment

lambda-runtime/src/runtime.rs
jlizen: This note might be a nice comment

lambda-runtime/src/runtime.rs
GradExMachina: Agreed!

lambda-runtime/src/runtime.rs
GradExMachina: I seem to have missed that 🤣 no reason at all.

lambda-runtime/src/runtime.rs
GradExMachina: So one of the things I wanted to do was have a package, something like `aws-lambda-simulator`, that will provide a small server implementation, like above, and will allow users to customize behaviour, run benchmarks, etc, all in their integration suite, without needing to deploy.

I don't think we will, or should get there at this point in time, but I think it is worth considering.

jlizen: Yeah, that would be nice. Or even just exposing this as a `test-util` feature of the runtime.

I've dabbled with doing similar using `cargo-lambda`'s simulator, but it was a bit unwieldy.

Might merit a backlog GitHub issue?

PR #1089, opened 2026-01-15
fix(lambda-events): Deprecate authorizer-specific fields in ApiGatewayV2httpRequest

### LINKED ISSUE #1080: ApiGatewayV2httpRequest.http_method may return incorrect method for standard v2 payloads
Hi. I have stumbled upon this issue, and I am not sure if there's something I am missing that would explain it, but it seems it could possibly be worth looking at.

In [`ApiGatewayV2httpRequest`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-events/src/event/apigw/mod.rs#L143-L195), the `http_method` is a top level field, but it's not really part of the standard ApiGatewayV2 payload, as far as I know. The actual method field is in `requestContext.http.method`. 

By silently defaulting to `GET` when the field is missing (which is always the case for standard HTTP API v2 requests), user code that relies on this field could incorrectly handle requests, treating a POST, PUT, or DELETE as a GET. This could lead to incorrect routing, bypassed validation logic, or potential security issues if some logic relies on `GET` being safe.

As I said, I am not sure if there are reasons for that, but even looking at the test fixtures `example-apigw-v2-*`, the JSON examples do not present the `httpMethod` field in that location.

There is one test [`example_apigw_v2_custom_authorizer_v1_request`](https://github.com/awslabs/aws-lambda-rust-runtime/blob/main/lambda-events/src/event/apigw/mod.rs#L1250-L1260) that parses a custom authorizer v1 payload into `ApiGatewayV2httpRequest` and checks `http_method`. However, there's already a dedicated `ApiGatewayV2CustomAuthorizerV1Request` struct for that payload format.

The same applies to `kind`, `method_arn`, `identity_source`, and `authorization_token` — all are custom authorizer v1 specific fields that don't appear in standard HTTP API v2 payloads.

Possible solutions that I can think of:
- Deprecate these authorizer-specific fields on `ApiGatewayV2httpRequest` and guide users to `ApiGatewayV2CustomAuthorizerV1Request` for that use case
- Add documentation warning users to use `request_context.http.method` instead for standard HTTP API v2 requests

Happy to work on a small PR if it's confirmed to be an issue.

dap0am:
## Description

This PR addresses issue #1080 by deprecating fields in `ApiGatewayV2httpRequest` that are only present in Custom Authorizer v1 payloads but not in standard HTTP API v2 requests.

## Changes

Deprecates the following fields in `ApiGatewayV2httpRequest`:
- `kind` (type)
- `method_arn`
- `http_method`
- `identity_source`
- `authorization_token`

## Problem

The `http_method` field incorrectly defaults to `GET` when missing from the JSON payload. This is problematic because:
- Standard HTTP API v2 requests do not include a top-level `httpMethod` field
- The actual HTTP method is located in `requestContext.http.method`
- Defaulting to `GET` can cause POST/PUT/DELETE requests to be incorrectly treated as GET requests
- This can lead to security vulnerabilities where unsafe methods bypass validation logic

## Solution

1. **Deprecation**: Added `#[deprecated]` attributes to all authorizer-specific fields with clear guidance
2. **Documentation**: Added comprehensive struct-level and field-level documentation explaining:
   - For standard HTTP API v2 requests, use `request_context.http.method`
   - For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request`
3. **Testing**: Added a new test `apigw_v2_correct_http_method_usage()` demonstrating the correct usage pattern

## Backward Compatibility

All deprecated fields remain functional and existing code will continue to work. Users will receive deprecation warnings guiding them to the correct approach.

## Semver Check Status

⚠️ The `semver` check failure is **expected and correct**. Marking public fields as deprecated requires a **minor version bump** (1.0.3 → 1.1.0) according to semantic versioning rules, which will be handled by maintainers during the release process.

## Testing

- ✅ All 27 existing API Gateway tests pass
- ✅ New test added to demonstrate correct HTTP method access
- ✅ Code formatted with `rustfmt`
- ✅ Deprecation warnings work as expected

Fixes #1080

### DIFF
--- a/lambda-events/src/event/apigw/mod.rs
+++ b/lambda-events/src/event/apigw/mod.rs
@@ -156,7 +156,7 @@ pub struct ApiGatewayV2httpRequest {
     /// This field is not present in standard HTTP API v2 requests.
     /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
     #[deprecated(
-        since = "0.14.0",
+        since = "1.1.0",
         note = "This field is only present in Custom Authorizer v1 requests. Use `ApiGatewayV2CustomAuthorizerV1Request` for authorizer payloads."
     )]
     #[serde(default, rename = "type")]
@@ -166,7 +166,7 @@ pub struct ApiGatewayV2httpRequest {
     /// This field is not present in standard HTTP API v2 requests.
     /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
     #[deprecated(
-        since = "0.14.0",
+        since = "1.1.0",
         note = "This field is only present in Custom Authorizer v1 requests. Use `ApiGatewayV2CustomAuthorizerV1Request` for authorizer payloads."
     )]
     #[serde(default)]
@@ -182,7 +182,7 @@ pub struct ApiGatewayV2httpRequest {
     /// This top-level `http_method` field is only present in Custom Authorizer v1 requests.
     /// For those payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
     #[deprecated(
-        since = "0.14.0",
+        since = "1.1.0",
         note = "For standard HTTP API v2 requests, use `request_context.http.method` instead. This field defaults to GET when missing and may return incorrect values. For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request`."
     )]
     #[serde(with = "http_method", default = "default_http_method")]
@@ -192,7 +192,7 @@ pub struct ApiGatewayV2httpRequest {
     /// This field is not present in standard HTTP API v2 requests.
     /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
     #[deprecated(
-        since = "0.14.0",
+        since = "1.1.0",
         note = "This field is only present in Custom Authorizer v1 requests. Use `ApiGatewayV2CustomAuthorizerV1Request` for authorizer payloads."
     )]
     #[serde(default)]
@@ -202,7 +202,7 @@ pub struct ApiGatewayV2httpRequest {
     /// This field is not present in standard HTTP API v2 requests.
     /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
     #[deprecated(
-        since = "0.14.0",
+        since = "1.1.0",
         note = "This field is only present in Custom Authorizer v1 requests. Use `ApiGatewayV2CustomAuthorizerV1Request` for authorizer payloads."
     )]
     #[serde(default)]


dap0am:
## Semver Check Failure - Expected Behavior

The `semver` check is failing as expected because this PR marks several public fields as `#[deprecated]`, which requires a **minor version bump** according to semantic versioning rules.

### What the semver checker found:
- Marking public struct fields as deprecated changes the user experience (compiler warnings)
- This requires bumping from `1.0.3` → `1.1.0` (minor version bump)

### Resolution:
This is expected and correct behavior. The version bump will be handled by the maintainers during the release process. The deprecations are necessary to guide users away from fields that can cause security issues (as described in #1080).

### Why this change is important:
The deprecated `http_method` field defaults to `GET` when missing, which can cause POST/PUT/DELETE requests to be incorrectly treated as GET requests, potentially leading to security vulnerabilities. The deprecation warnings will guide users to use the correct `request_context.http.method` field instead.

lambda-events/src/event/apigw/mod.rs
```diff
@@ -137,18 +137,74 @@ pub struct ApiGatewayProxyRequestContext {
...
+    /// **Warning**: This field defaults to `GET` if not present in the payload, which may lead to
+    /// incorrect behavior. For standard HTTP API v2 requests, this field is not present in the JSON payload.
+    ///
+    /// **Use `request_context.http.method` instead for standard HTTP API v2 requests**, which contains
+    /// the actual HTTP method for the request.
+    ///
+    /// This top-level `http_method` field is only present in Custom Authorizer v1 requests.
+    /// For those payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
+    #[deprecated(
+        since = "0.14.0",
```
GradExMachina: I may be missing something, but why since 0.14.0?

lambda-events/src/event/apigw/mod.rs
```diff
@@ -137,18 +137,74 @@ pub struct ApiGatewayProxyRequestContext {
...
+        note = "For standard HTTP API v2 requests, use `request_context.http.method` instead. This field defaults to GET when missing and may return incorrect values. For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request`."
+    )]
     #[serde(with = "http_method", default = "default_http_method")]
     pub http_method: Method,
+    /// The identity source from Custom Authorizer v1 requests.
+    ///
+    /// This field is not present in standard HTTP API v2 requests.
+    /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
+    #[deprecated(
+        since = "0.14.0",
```
GradExMachina: Ditto

lambda-events/src/event/apigw/mod.rs
```diff
@@ -137,18 +137,74 @@ pub struct ApiGatewayProxyRequestContext {
...
+        note = "This field is only present in Custom Authorizer v1 requests. Use `ApiGatewayV2CustomAuthorizerV1Request` for authorizer payloads."
+    )]
     #[serde(default)]
     pub identity_source: Option<String>,
+    /// The authorization token from Custom Authorizer v1 requests.
+    ///
+    /// This field is not present in standard HTTP API v2 requests.
+    /// For Custom Authorizer v1 payloads, use `ApiGatewayV2CustomAuthorizerV1Request` instead.
+    #[deprecated(
+        since = "0.14.0",
```
GradExMachina: Ditto

dap0am:
Thanks for the quick review! 

I've updated the deprecation version, but I want to confirm the correct value for the `since` parameter:

- Current version in `Cargo.toml`: **1.0.3**
- You mentioned bumping to: **1.1.0**

Should the `since` value be:
1. **`1.0.3`** (current version)
2. **`1.1.0`** (the version where this deprecation will be released)

I currently have it set to `1.1.0` since that's when users will first see the deprecation warnings. Please let me know if this should be different!

GradExMachina:
Sorry yeah, 1.1.0 is appropriate. 

We will be bumping that as we are adding the builders.

dap0am:
> Sorry yeah, 1.1.0 is appropriate.
> 
> We will be bumping that as we are adding the builders.

Thanks, I will commit the changes now.

dap0am:
✅ Updated! All deprecation attributes now use `since = "1.1.0"` as confirmed. The changes have been pushed.

PR #1090, opened 2026-01-15
Add builder pattern support for event response types

### LINKED ISSUE #1060: [aws_lambda_events] No longer able to construct response types a lambda may return as a response (they are `non_exhaustive`)
A number of the types in this crate are responses that a lambda might return. With the [recent change to mark structs as `non_exhaustive`](#1016), these types can no longer be constructed. Some examples:

* [`SqsBatchResponse`](https://docs.rs/aws_lambda_events/latest/aws_lambda_events/event/sqs/struct.SqsBatchResponse.html) (and its child types) are [returned from an SQS lambda](https://docs.aws.amazon.com/lambda/latest/dg/example_serverless_SQS_Lambda_batch_item_failures_section.html)
* [`AppSyncLambdaAuthorizerResponse`](https://docs.rs/aws_lambda_events/latest/aws_lambda_events/event/appsync/struct.AppSyncLambdaAuthorizerResponse.html) (and its child types) are [returned from an AppSync authorizer lambda](https://docs.aws.amazon.com/appsync/latest/devguide/security-authz.html#aws-lambda-authorization).
* [`ApiGatewayCustomAuthorizerResponse`](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html) (and its child types) are [returned from an API Gateway authorizer lambda](https://docs.aws.amazon.com/apigateway/latest/developerguide/api-gateway-lambda-authorizer-output.html).

I'm sure there are other response types in this crate that lambdas would construct, but these are the ones I'm currently constructing and am no longer able to after they've been marked `non_exhaustive`.

FullyTyped:
Add derived builders to events.
Builders are conditionally compiled with the "builders" feature flag.

📬 *Issue #, if available:*
fixes #1060

### DIFF
--- a/lambda-events/src/event/sqs/mod.rs
+++ b/lambda-events/src/event/sqs/mod.rs
@@ -500,6 +500,4 @@ mod test {
         assert_eq!(response.batch_item_failures.len(), 1);
         assert_eq!(response.batch_item_failures[0].item_identifier, "msg-4");
     }
-
-
 }


GradExMachina:
Hi folks, so with this PR we aim to address the issues that were introduced when we tagged events as non-exhaustive.

@lmammino @jeastham1993 @ikai104 your feedback is appreciated. I looked at the examples you've posted in #1060 and added before and after in the examples folder.

lambda-events/Cargo.toml
```diff
@@ -20,6 +20,7 @@ edition = "2021"
 base64 = { workspace = true }
 bytes = { workspace = true, features = ["serde"], optional = true }
 chrono = { workspace = true, optional = true }
+derive_builder = { version = "0.20", optional = true }
```
jlizen: How would you feel about using `bon` instead? The big reason would be that it uses typestate for compile time verification, which is pretty nice for ergonomics.

The better refactoring support probably doesn't affect us since we are anyway all pub so would break semver either way on internal changes.

The function builders we don't need yet, though I wouldn't be surprised if it is useful later. For instance, we have split into separate APIs for the concurrent mode. Future features like that would be nice to add as optional parameters via a builder.

lambda-events/README.md
```diff
@@ -27,6 +27,56 @@ This crate divides all Lambda Events into features named after the service that
 cargo add aws_lambda_events --no-default-features --features apigw,alb
 ```
 
+### Builder pattern support
+
+The crate provides an optional `builders` feature that adds builder pattern support for event types using the [bon](https://crates.io/crates/bon) crate. This enables type-safe, immutable construction of event responses with a clean, ergonomic API.
```
jlizen: Nit: `bon` feels like an implementation detail. We are not exposing bon specific types in our public API contract and might change what builder we use in the future (while preserving the API). Not sold on mentioning it in the README, is there anything we expect users to need to go and look up in bon docs to understand?

lambda-events/README.md
```diff
@@ -27,6 +27,56 @@ This crate divides all Lambda Events into features named after the service that
...
+    let response = ApiGatewayV2CustomAuthorizerSimpleResponse::builder()
+        .is_authorized(true)
+        .context(context)
+        .build();
+
+    Ok(response)
+}
+```
+
+Key benefits of bon builders:
```
jlizen: I would probably ditch this, unless we want to have some sort of INTERNALS.md to explain design decisions. It doesn't feel like it belongs in end-user docs though.

lambda-events/examples/comprehensive-builders.rs
```diff
@@ -0,0 +1,44 @@
+// Example demonstrating builder pattern usage for AWS Lambda events
+#[cfg(feature = "builders")]
+use aws_lambda_events::event::{
+    dynamodb::Event as DynamoDbEvent, kinesis::KinesisEvent, s3::S3Event,
+    secretsmanager::SecretsManagerSecretRotationEvent, sns::SnsEvent, sqs::SqsEvent,
+};
+
+#[cfg(feature = "builders")]
+fn main() {
+    // S3 Event - Object storage notifications
```
jlizen: Minor: These would be much more useful examples if they included at least one record.

lambda-events/examples/lambda-runtime-authorizer-builder.rs
```diff
@@ -0,0 +1,145 @@
+// Example showing how builders solve the Default trait requirement problem
+// when using lambda_runtime with API Gateway custom authorizers
+//
+// ❌ OLD WAY (with Default requirement):
```
jlizen: Nit: I'm not sure there is much value in showing the old way. It adds noise to this example. Shouldn't readers just be pointed to the recommended pattern without the history lesson?

lambda-events/examples/lambda-runtime-sqs-batch-builder.rs
```diff
@@ -0,0 +1,135 @@
+// Example showing how builders simplify SQS batch response construction
+// when handling partial batch failures
+//
+// ❌ OLD WAY (with Default):
```
jlizen: Same feedback on ditching old way

lambda-events/src/event/sqs/mod.rs
```diff
@@ -476,4 +500,52 @@ mod test {
         assert_eq!(response.batch_item_failures.len(), 1);
         assert_eq!(response.batch_item_failures[0].item_identifier, "msg-4");
     }
+
+    #[test]
+    #[cfg(all(feature = "sqs", feature = "builders"))]
+    fn test_bon_sqs_event_builder() {
```
jlizen: nit: i'd probably ditch bon from the name from all of these. Someday we might want to change out the framework, and anyway we would need to maintain the api. What this is validating is the API contract, not bon specifically.

lambda-events/examples/lambda-runtime-authorizer-builder.rs
jlizen: If anything, this would be a great code snippet to include in the release notes

Cargo.toml
```diff
@@ -7,6 +7,7 @@ members = [
     "lambda-runtime",
     "lambda-extension",
     "lambda-events",
+    "lambda-events/lambda-events-examples",
 ]
 
 exclude = ["examples"]
```
GradExMachina:
```suggestion
]

exclude = ["examples","lambda-events/lambda-events-examples"]
```

PR #1091, opened 2026-01-19
Log timeouts when DP responds with timedout

FullyTyped:
- **base test impl**
- **test-impl**

📬 *Issue #, if available:*

✍️ *Description of changes:*

In concurrent execution, when the runtime times-out, Dataplane will not stop the execution, but instead will inform the runtime through the trailers that we've timed-out when we respond. This PR intends to consume the trailers to ensure that in those scenarios the runtime correctly logs. 

### DIFF
--- a/lambda-runtime/src/layers/api_client.rs
+++ b/lambda-runtime/src/layers/api_client.rs
@@ -14,7 +14,7 @@ use tracing::{error, warn};
 /// This type is only meant for internal use in the Lambda runtime crate. It neither augments the
 /// inner service's request type nor its error type. However, this service returns an empty
 /// response `()` as the Lambda request has been completed.
-pub struct RuntimeApiClientService<S> {
+pub(crate) struct RuntimeApiClientService<S> {
     inner: S,
     client: Arc<Client>,
 }
@@ -78,7 +78,7 @@ where
 /// The type parameter `B` represents the response body type. In production this is `hyper::body::Incoming`,
 /// but for testing it can be any body type that implements `http_body::Body`.
 #[pin_project(project = RuntimeApiClientFutureProj)]
-pub enum RuntimeApiClientFuture<F, B> {
+pub(crate) enum RuntimeApiClientFuture<F, B> {
     /// **Invoke Phase**: Polling the inner service to build the Lambda response request.
     ///
     /// Contains:


jlizen:
I haven't looked closely but it looks like this impl would require adding generics for the runtime API client and other breaking changes? Customers are unlikely to name them but we probably still can't do break that API... Probably we need to cut over to a new type and deprecate the old? What are you thinking?

GradExMachina:
Adding a non-constrained generic shouldn't be a problem, and it is pretty beneficial in this scenario. 

I guess the alternative is injecting a client that communicates over channels.

I am not sure. Maybe there's a different way to test this that I am just missing.

jlizen:
> I guess the alternative is injecting a client that communicates over channels.

Big ++ to this - some sort of `test-util`-feature-gated utility that allows overriding the client with an arbitrary trait implementation, seems great.

> Adding a non-constrained generic shouldn't be a problem

Not a problem, but it is semver breaking, in case anybody names the type. (They are unlikely to, but still). Along similar lines, we are adding a new variant to `RuntimeApiClientFuture`, which is not tagged as `#[non_exhaustive]`, which is also breaking.

I don't think it's a big deal, these are unlikely to be used by end user code, but we probably should play by the rules and add new names for these and deprecate the old? Anyway if we cut over all of our other crates relying on these types, end user churn would be minimal.

darklight3it:
I would say `RuntimeApiClientFuture` is not accessible outside layer because of this rule 

https://github.com/aws/aws-lambda-rust-runtime/blob/61855c2eabfc725f93cad6deabb7d6fa203e35a9/lambda-runtime/src/layers/mod.rs#L3

The same is for `RuntimeApiClientService` which has `pub(crate)` visibility. 

https://github.com/aws/aws-lambda-rust-runtime/blob/61855c2eabfc725f93cad6deabb7d6fa203e35a9/lambda-runtime/src/layers/mod.rs#L10

So cx cannot infer the breaking enum even through the struct using it.

In this case there's not much sense also in putting also the non_exhaustive

darklight3it:
@FullyTyped I'll have a look at this stuff and try to make it prod-ready.

jlizen:
> I would say RuntimeApiClientFuture [and Serivce] is not accessible outside layer because of this rule

Looking through, I think that you're right!

It doesn't seem to be exposed as an unnameable type either (ie, using a public type in a private module, but that shows up in a public api). For that, we still might have a problem, since you can still match on it via field accessor syntax.

Can we just move both to `pub(crate)` in that case? That would pretty clearly flush out public exposure that we are missing.

darklight3it:
Moving the discussion to this CR

https://github.com/aws/aws-lambda-rust-runtime/pull/1105

Issue #977, opened 2025-04-09
Do the TelemetryAPI records support JSON record fields?


mheffner:
According to the TelemetryAPI [docs](https://docs.aws.amazon.com/lambda/latest/dg/telemetry-schema-reference.html#telemetry-api-function), if you're using a more recent schema version than "2022-12-13", the `record` field of function and extension logs may be a string OR a JSON record.

From the code it would appear that these only support the string records. Do these need to be updated to something like `serde_json::Value` to handle both cases?

https://github.com/awslabs/aws-lambda-rust-runtime/blob/f69280e9b92e9a7acc92cfa04a860d9b2b83b5c0/lambda-extension/src/telemetry.rs#L25-L30

Thanks!

jlizen:
Looks like you're right! Feel free to modify this issue into a feature request, or take a stab at implementing it!

It's a bit awkward since we're relying heavily on `serde` to do the deserialization implicitly. So drilling through state information about which record type is enabled from something stored in extension or runtime state, is pretty unpleasant.

I could think of some ugly things involving global state and a custom deserialize impl for some newtyped inner structs (probing for `AWS_LAMBDA_LOG_FORMAT` and caching it, testing the first record you see to find out if it is JSON and stashing that result, etc).

Meanwhile the 'rusty' thing to do might be to encode it at the type level via a generic (or more likely a separate struct, since adding a generic to `LambdaTelemetryRecord` would cause semver breakage). And then the caller could just specify that in their handler if they want to receive processed JSON.  The downside of that approach though is that it would be effectively hardcoding a choice around log format, whereas advanced logging controls are designed to be changeable in console, etc. I guess probably the caller could write two different handlers and choose between them based on the same ENV, but it's all pretty klunky.

Just sharing my perspective! (Not a maintainer / not offering to implement this)

jlizen:
cc @bnusunny for perspective on this one... my knee jerk is, probably not good to cut to modeling a `serde_json::Value` type, even across a breaking semver boundary. Seems like a future semver hazard and anyway it's trivial for callers to choose to parse the String into JSON themselves. If we did want to do it, we also probably would want to make it a [RawValue](https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html) to avoid extra cost if caller is just passing it through without interacting with fields.

That said I do think a `.json()` method might be good as a middle ground to help with ergonomics.

jlizen:
[...]

PR #1098, opened 2026-02-11
Make Telemetry API log record type generic

raphael-theriault-swi:
📬 *Issue #, if available:* #977

✍️ *Description of changes:*

This adds a generic type parameter to `LambdaTelemetryRecord` for logs to support structured JSON logs. I initially tried to implement [the `RawValue` newtype based solution](https://github.com/aws/aws-lambda-rust-runtime/issues/977#issuecomment-2853186504) suggested in #977, but it turns out [you can't use `RawValue` within `#[serde(flatten)]`](https://github.com/serde-rs/json/issues/1051).

This is a breaking change given it introduces a new generic parameter, but IMO it's better than introducing a whole new set of duplicate types, especially given supporting managed instances may already require breakage soon.

Unlike suggested in #977, this doesn't lock users into either supporting text or JSON logs at compile time, since they can just use an untagged enum:

```rs
#[derive(Deserialize)]
#[serde(untagged)]
enum LogRecord {
    Text(String),
    Json(StructuredLogRecord),
}
```

I also opted against adding a new struct for JSON logs since they don't have a fixed schema, some extension authors might want to use `serde_json::Value`, some might want a more specialized type.

### DIFF
--- a/lambda-extension/src/extension.rs
+++ b/lambda-extension/src/extension.rs
@@ -185,7 +185,7 @@ where
     }
 
     /// Create a new [`Extension`] with a service that receives Lambda telemetry data.
-    /// 
+    ///
     /// By default, telemetry log records are deserialized as `String`, but
     /// it's possible to configure Lambda functions to emit logs in JSON format.
     /// For more information, refer to [`Self::with_telemetry_record_type`].


jlizen:
> This is a breaking change given it introduces a new generic parameter, but IMO it's better than introducing a whole new set of duplicate types, especially given supporting managed instances may already require breakage soon.

Managed instances will not require semver breaking. We are a long ways off from release 2.x.

With that in mind, do you have any thoughts about a semver-safe way to do it? I guess we just add a new API, deprecate the old, but allow continuing to use old until next release version?

jlizen:
(Also, thanks for picking this up! This is definitely in an awkward spot right now)

raphael-theriault-swi:
I'm thinking maybe an opt-in feature to enable the generic might be more elegant ? Then the feature can just be removed when a future `2.x` comes out.

jlizen:
The opt-in feature will work, but it seems a bit confusing to have a feature that shifts runtime behavior like this, especially for such a limited feature scope. Let me get a temperature check from the other maintainers on this and follow up.

jlizen:
Coming back to this - 

Unfortunately a feature flag doesn't escape semver constraints, even if it is opt in (given feature unification).

Per cargo:
> Enabling a feature should not introduce a SemVer-incompatible change. For example, the feature shouldn’t change an existing API in a way that could break existing uses.

[ref](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility)

And then, any introduction of generics that don't have an default inference available, is breaking. Which, unfortunately, the function signatures `run()` and `register` run afoul of. The types are otherwise fine.

EDIT: Hold on a sec, are those just methods on extension? We might be able to work around it by hoisting it to be struct-wide, with a default.

jlizen:
Yeah, it seems to be working ok with an extra generic on `Extension`:

```
pub struct Extension<'a, E, L, T, R = String>
 {
    extension_name: Option<&'a str>,
    ...
    _record_type: PhantomData<R>
}
```
(Maybe there is a way to do it without the PhantomData as well?)

And then we can hoist the telemetry where clause back to the struct definition:
```
impl<'a, E, L, T, R> Extension<'a, E, L, T, R>
where
    E: Service<LambdaEvent>,
    E::Future: Future<Output = Result<(), E::Error>>,
    E::Error: Into<Error> + fmt::Display + fmt::Debug,

    // Fixme: 'static bound might be too restrictive
    L: MakeService<(), Vec<LambdaLog>, Response = ()> + Send + Sync + 'static,
    L::Service: Service<Vec<LambdaLog>, Response = ()> + Send + Sync,
    <L::Service as Service<Vec<LambdaLog>>>::Future: Send + 'a,
    L::Error: Into<Error> + fmt::Debug,
    L::MakeError: Into<Error> + fmt::Debug,
    L::Future: Send,
    T: MakeService<(), Vec<LambdaTelemetry<R>>, Response = ()> + Send + Sync + 'static,
    T::Service: Service<Vec<LambdaTelemetry<R>>, Response = ()> + Send + Sync,
    <T::Service as Service<Vec<LambdaTelemetry<R>>>>::Future: Send + 'a,
    T::Error: Into<Error> + fmt::Debug,
    T::MakeError: Into<Error> + fmt::Debug,
    T::Future: Send,
    R: DeserializeOwned + Send + 'static,
{ }
```

jlizen:
And then, it's unpleasant for the user to have to specify all the generic parameters as `Extension<_, _, _, serde_json::Value>`.

But, seem to have consistent inference if we update `with_telemetry processor` to accept an incoming type:
```
impl<'a, E, L, T, R> Extension<'a, E, L, T, R> {
   ...
    pub fn with_telemetry_processor<N, NS, NewR>(self, lp: N) -> Extension<'a, E, L, N, NewR> { }

}
```

And then you can specify it in your handler that this function receives:
```
async fn handler(events: Vec<LambdaTelemetry<serde_json::Value>>) -> Result<(), Error> {
   println!("{events:#}");
    Ok(())
}

#[tokio::main]
fn main() {
    let telemetry_processor = SharedService::new(service_fn(handler));

    Extension::new()
        .with_telemetry_processor(telemetry_processor)
        .run()
        .await.unwrap()
}
```

(Or not, and then it defaults to String, and stays backwards compatible)

jlizen:
What do you think about the proposed API @raphael-theriault-swi ?

I think we should flip the generic default next major version, but that might be a ways off. (At which point, your untagged `LogRecord` seems nice)

Are these ergonomics adequate in the meantime? No feature flags at least?

Probably a small `examples/` entry would be good too to show how to use this (or update `extension-telemetry-basic`).

raphael-theriault-swi:
The thing is that adding the default generic type to the enum itself is technically a breaking change the moment you try to use an enum literal of a variant that doesn't include the generic.

```rs
enum Generic<T = &'static str> {
    Generic(T),
    Fixed(&'static str),
}

fn main() {
    let generic = Generic::Generic("inference succeeds");
    let fixed = Generic::Fixed("inference fails");
}
```

jlizen:
> The thing is that adding the default generic type to the enum itself is technically a breaking change the moment you try to use an enum literal of a variant that doesn't include the generic.

That may be, but in the current approach, `LambdaTelemetryRecord` is using the generic on each variant, and is already `pub` and not `#[non_exhaustive]`, so we couldn't add a new variant without the generic anyway, right? Or else that would be a breaking change regardless?

You can see that the [semver check](https://github.com/aws/aws-lambda-rust-runtime/actions/runs/21926881586/job/63780471741?pr=1098) is only complaining about the new generics in the methods, which my suggestion avoids.

Please correct me if I am missing something!

raphael-theriault-swi:
Yeah, I do't think there's anyway to make this a totally non-breaking change without duplicating pretty large parts of the API. Personally I feel the solution you suggested in your recent replies is the best way to go, I just wanted to bring up the potential breakage.

Should I go ahead and implement it that way ?

rcoh:
Typically, inference failure is not considered SemVer breaking: https://doc.rust-lang.org/cargo/reference/semver.html#fn-generalize-compatible

One option (I have not totally read all context in here) is to introduce a new type `LambdaTelemetryRecordV2<T>` (bikeshed), replacing `LambdaTelemetryRecord` with `pub type LambdaTelemetryRecord = LambdaTelemetryRecordV2<String>`

you could then add `register_generic_extension<TL>` that used `LambdaRecordV2<TL>` to allow for using the generic.

Since `register_extension` would delegate directly to `register_generic_extension`, the API itself should remain largely unchanged and the only change in surface area would be the new methods.

jlizen:
I think Russell's solution is a clean one that would avoid duplication:
- use a type alias `pub type LambdaTelemetryRecord = GenericLambdaTelemetryRecord<String>`
- under the hood, remove `LambdaTelemetryRecord` entirely and replace it with `GenericLambdaTelemetryRecord`, which looks like the generic version of the original that we have been working with
- add a new API to the builder, `with_generic_telemetry_processor()`, which accepts `GenericLambdaTelemetryRecord` for an arbitrary type
- probably also add, `JsonLambdaTelemetryRecord = GenericLambdaTelemetryRecord<serde_json::Value>`, and `with_json_telemetry_processor` for ergonomics?

And then no inference breaks on existing `LambdaTelemetryRecord` construction, since `LambdaTelemetryRecord` is a type alias with no generics.

And usage looks fine:
```
# record is a string
async fn handler(events: Vec<LambdaTelemetryRecord) -> Result<(), Error> {
   println!("{events:#}");
    Ok(())
}

#[tokio::main]
fn main() {
    let telemetry_processor = SharedService::new(service_fn(handler));

    Extension::new()
        .with_telemetry_processor(telemetry_processor)
        .run()
        .await.unwrap()
}
```


```
# record is a serde_json::Value
async fn handler(events: Vec<JsonLambdaTelemetryRecord) -> Result<(), Error> {
   println!("{events:#}");
    Ok(())
}

#[tokio::main]
fn main() {
    let telemetry_processor = SharedService::new(service_fn(handler));

    Extension::new()
        .with_json_telemetry_processor(telemetry_processor)
        .run()
        .await.unwrap()
}
```

(and you can imagine for the generic)

What do you think about that, @raphael-theriault-swi ?

If there are blockers to this approach, I think we are probably ok to move forward with the default generic that might break inference on direct `LambdaTelemetryRecord` usage, with just a minor version bump (which should go in this PR for `lambda-extension` `Cargo.toml` / lambda-runtime's dependency, since I don't think cargo-semver-checks is smart enough to flag it).

raphael-theriault-swi:
Sounds like it would solve pretty much every issue brought up in the thread, love it. I'll update the PR !

raphael-theriault-swi:
Should be ready for another look @jlizen

jlizen:
Oof - sorry @raphael-theriault-swi , but it looks like that is actually more breaking than the risk of inference failure - 
ref: [failed run](https://github.com/aws/aws-lambda-rust-runtime/actions/runs/22319772619/job/64578715439?pr=1098):
```

--- failure enum_missing: pub enum removed or renamed ---

Description:
A publicly-visible enum cannot be imported by its prior path. A `pub use` may have been removed, or the enum itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.46.0/src/lints/enum_missing.ron

Failed in:
  enum lambda_extension::LambdaTelemetryRecord, previously in file /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lambda-extension-1.0.2/src/telemetry.rs:25

--- failure struct_missing: pub struct removed or renamed ---

Description:
A publicly-visible struct cannot be imported by its prior path. A `pub use` may have been removed, or the struct itself may have been renamed or removed entirely.
        ref: https://doc.rust-lang.org/cargo/reference/semver.html#item-remove
       impl: https://github.com/obi1kenobi/cargo-semver-checks/tree/v0.46.0/src/lints/struct_missing.ron

Failed in:
  struct lambda_extension::LambdaTelemetry, previously in file /home/runner/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/lambda-extension-1.0.2/src/telemetry.rs:14

```

Some more context: https://predr.ag/blog/re-exporting-enum-with-type-alias-breaking-change-not-major/
(it appears that the conclusion was that in fact it was a major breaking change, after all).

Based on that, I think we should go ahead with the plan from before this one - hoist the generic to the struct level, use the default, and risk breaking inference if somebody is directly constructing the enum. I think it's a niche enough use case, and trivially fixable, that it is sufficient to do the minor version bump and document it as such in our release notes.

Sorry for the churn!

If you go ahead with that tweak, I can stage the minor version bump myself in a separate PR so we don't forget (since i think cargo-semver-checks won't recognize it as needed).

raphael-theriault-swi:
No worries about the churn, I had no idea either, been learning of a few edge cases on this one :)

jlizen:
@raphael-theriault-swi -

The recent changes won't work either, since introducing new generics on a method is breaking (and we can't set defaults on method generics). Meanwhile, hoisting to struct level alone isn't enough, since then `new()` and `default()` will constrain to `String`, unless they themselves have a generic, which is also breaking.

Instead, what we can do would be add a standalone method, `with_telemetry_record_type`, that can be called before `with_telemetry_processor` and constrain its type. It would need its own standalone impl block, I think, if we want to avoid footguns by constraining it to only be callable BEFORE the `with_telemetry_processor`.

It's a bit ugly, but not hideous, just requires a simple turbofish:

```
impl<'a, E, L> Extension<'a, E, L, MakeIdentity<Vec<LambdaTelemetry>>> {
    /// Set the deserialization type for telemetry records.
    ///
    /// By default, telemetry records are deserialized as `String`.
    /// Use this method to deserialize into a different type, such as `serde_json::Value`.
    /// Must be called before [`Self::with_telemetry_processor`].
    ///
    /// ```
    /// use lambda_extension::{Extension, LambdaTelemetry, SharedService, service_fn};
    ///
    /// async fn handler(events: Vec<LambdaTelemetry<serde_json::Value>>) -> Result<(), lambda_extension::Error> {
    ///     for event in &events {
    ///         println!("{event:?}");
    ///     }
    ///     Ok(())
    /// }
    ///
    /// let _ext = Extension::new()
    ///     .with_telemetry_record_type::<serde_json::Value>()
    ///     .with_telemetry_processor(SharedService::new(service_fn(handler)));
    /// ```
    pub fn with_telemetry_record_type<NewTL>(self) -> Extension<'a, E, L, MakeIdentity<Vec<LambdaTelemetry<NewTL>>>, NewTL> {
        Extension {
            _telemetry_record_type: PhantomData,
            telemetry_processor: None,
            events_processor: self.events_processor,
            extension_name: self.extension_name,
            events: self.events,
            log_types: self.log_types,
            log_buffering: self.log_buffering,
            logs_processor: self.logs_processor,
            log_port_number: self.log_port_number,
            telemetry_types: self.telemetry_types,
            telemetry_buffering: self.telemetry_buffering,
            telemetry_port_number: self.telemetry_port_number,
        }
    }
}
```

I got it compiling locally with your example, with no inference issues (besides the risk that you've already called out around inference on `LambdaTelemetryRecord`, which we are willing to accept with a minor version bump).

(And then, `with_telemetry_processor` becomes roughly:
```
    /// Create a new [`Extension`] with a service that receives Lambda telemetry data.
    ///
    /// By default, telemetry records are deserialized as `String`.
    /// To parse into other record types, such as json, use
    /// [`Self::with_telemetry_record_type`] before calling this method.
    pub fn with_telemetry_processor<N, NS>(self, lp: N) -> Extension<'a, E, L, N, TL>
    where
        N: Service<()>,
        N::Future: Future<Output = Result<NS, N::Error>>,
        N::Error: Into<Error> + fmt::Display,
    {
        Extension {
            telemetry_processor: Some(lp),
            _telemetry_record_type: PhantomData,
            events_processor: self.events_processor,
            extension_name: self.extension_name,
            events: self.events,
            log_types: self.log_types,
            log_buffering: self.log_buffering,
            logs_processor: self.logs_processor,
            log_port_number: self.log_port_number,
            telemetry_types: self.telemetry_types,
            telemetry_buffering: self.telemetry_buffering,
            telemetry_port_number: self.telemetry_port_number,
        }
    }
```
)

raphael-theriault-swi:
@jlizen should be good now, hopefully

Issue #1099, opened 2026-02-16
Mismatch type description for Cognito CognitoEventUserPoolsPreTokenGenV2 Event


AbhigyaKrishna:
There is a type mismatch for [`claims_to_add_or_override`](https://github.com/aws/aws-lambda-rust-runtime/blob/61855c2eabfc725f93cad6deabb7d6fa203e35a9/lambda-events/src/event/cognito/mod.rs#L572) in [CognitoEventUserPoolsPreTokenGenV2](https://github.com/aws/aws-lambda-rust-runtime/blob/61855c2eabfc725f93cad6deabb7d6fa203e35a9/lambda-events/src/event/cognito/mod.rs#L473) event. `claims_to_add_or_override` should accept any of the json types instead of only strings.

The [documentation at AWS](https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-user-pools-lambda-trigger-syntax-pre-token-generation) mentions the following type description for the event as:
```json
{
    "request": {
        "userAttributes": {
            "string": "string"
        },
        "scopes": ["string", "string"],
        "groupConfiguration": {
            "groupsToOverride": ["string", "string"],
            "iamRolesToOverride": ["string", "string"],
            "preferredRole": "string"
        },
        "clientMetadata": {
            "string": "string"
        }
    },
    "response": {
        "claimsAndScopeOverrideDetails": {
            "idTokenGeneration": {
                "claimsToAddOrOverride": {
                    "string": [accepted datatype]
                },
                "claimsToSuppress": ["string", "string"]
            },
            "accessTokenGeneration": {
                "claimsToAddOrOverride": {
                    "string": [accepted datatype]
                },
                "claimsToSuppress": ["string", "string"],
                "scopesToAdd": ["string", "string"],
                "scopesToSuppress": ["string", "string"]
            },
            "groupOverrideDetails": {
                "groupsToOverride": ["string", "string"],
                "iamRolesToOverride": ["string", "string"],
                "preferredRole": "string"
            }
        }
    }
}
```

##### Accepted datatypes:
* String
* Number
* Boolean
* Array of strings, numbers, booleans, or a combination of any of these
* JSON

#### Proposed Solution
```diff
- pub claims_to_add_or_override: HashMap<String, String>
+ pub claims_to_add_or_override: HashMap<String, serde_json::Value>
```

PR #1100, opened 2026-02-16
Fix type for claims_to_add_or_override in CognitoEventUserPoolsPreTokenGenV2

AbhigyaKrishna:
📬 *Issue #, if available:* #1099

✍️ *Description of changes:*
`claims_to_add_or_override` now has the correct type for overriding complex structure in claims. Pre token generation Lambda trigger `V2_0` allows to pass complex datatypes to ID and access token claim values.

### DIFF
--- a/lambda-events/src/event/cognito/mod.rs
+++ b/lambda-events/src/event/cognito/mod.rs
@@ -551,7 +551,7 @@ pub struct ClaimsAndScopeOverrideDetailsV2 {
 #[derive(Clone, Debug, Default, Deserialize, Eq, PartialEq, Serialize)]
 #[serde(rename_all = "camelCase")]
 pub struct CognitoIdTokenGenerationV2 {
-    pub claims_to_add_or_override: HashMap<String, String>,
+    pub claims_to_add_or_override: HashMap<String, Value>,
     pub claims_to_suppress: Vec<String>,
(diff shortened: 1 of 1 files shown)


jlizen:
This will be flagged as semver breaking, but the argument is that the previous case was a bug and would lead to incorrect deserialization failures? (Or would it? I think no, it would just cast everything to `String`)?

According to the bug argument will need to bypass cargo semver check for it I guess?

But, I'm wondering if we should instead be preserving old String and offering a new API for json? This seems parallel to: https://github.com/aws/aws-lambda-rust-runtime/issues/977

jlizen (approved):
I reread the upstream docs, and they are very explicit that this field is a complex object:
https://docs.aws.amazon.com/cognito/latest/developerguide/user-pool-lambda-pre-token-generation.html#cognito-pretoken-complex-objects-note

And, v1 explicitly does not have the complex type, so this is likely something that was present since the start of the protocol version.

So - sorry to anybody that this might break, but this is a bug that needs to be fixed, albeit very belatedly. It is blocking conventional use of the field for future users.

Thanks for the contribution, @AbhigyaKrishna !

jlizen:
Semver check failures are expected as this is a bugfix of a public type.

PR #1103, opened 2026-02-20
Introducing Harness Testing

darklight3it:
📬 *Issue #, if available:*

N/A

✍️ *Description of changes:*

This PR adds Docker-based testing infrastructure using AWS's `containerized-test-runner-for-aws-lambda`, enabling automated testing of Lambda functions in a containerized environment that closely mirrors the AWS Lambda execution environment.

The PR introduces a `test-dockerized` Makefile target that runs test suites defined in `test/dockerized/*.json` files. These test suites specify handlers to test (from the examples directory), request payloads, and expected response assertions with optional jq transforms for validation.

The infrastructure reuses Lambda binaries from the `/examples` folder as test handlers, demonstrating the concept with an initial test case for `basic-lambda`. Additional tests and multi-concurrency scenarios can be added by creating new test suite JSON files.

A GitHub Actions workflow provides CI/CD integration for automated testing on pull requests.

## Testing

Run dockerized tests locally:
```bash
make test-dockerized
```

Run RIE tests:
```bash
make test-rie
HANDLERS_TO_BUILD="basic-lambda basic-sqs" make test-rie
```

Build specific examples:
```bash
EXAMPLES="basic-lambda basic-lambda-concurrent" make build-examples
```

- [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.on.


.github/workflows/dockerized-test.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - name: Build the image
+      run: docker build . -t local/test -f Dockerfile.rie
+
+    - name: Run tests
+      uses: aws/containerized-test-runner-for-aws-lambda@v1
```
jlizen: Does this action exist? The ci is failing to find it, and I don't see references elsewhere on the Internet.

I wonder if the existing [test-rie](https://github.com/aws/aws-lambda-rust-runtime/blob/main/.github/workflows/test-rie.yml) workflow is close to what you want here, and could be adapted?

Seems like the main thing missing would be a small script to translate the request mappings into curl requests that assert against the response? Along with some parameterizing to feed in the proper lambda binary and test assets to the dockerfile?

jlizen: (I like what you are trying to do here with the testing harness though, seems super handy... If the referenced action is in development / not yet public... Excited to see it!)

darklight3it: Sorry 😢  for this you miss some context here, and I had no time to write this in the appropriate channel. This stuff is agreed with @maxday.

That action is already used in the ruby repo and Maxime is actively working in making it  public. The idea is to use the same harness test framework in all lambda runtimes.

I created a CR on that repo and this one is dependent on that to be working.

.github/workflows/dockerized-test.yml
```diff
@@ -0,0 +1,24 @@
...
+    - uses: actions/checkout@v3
+
+    - name: Build the image
+      run: docker build . -t local/test -f Dockerfile.rie
+
+    - name: Run tests
+      uses: aws/containerized-test-runner-for-aws-lambda@v1
+      with:
+        suiteFileArray: '["./tests/dockerized/*.json"]'
+        dockerImageName: 'local/test'
```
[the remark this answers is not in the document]
darklight3it: @maxday how did you solve this one? I included the restriction it's talking about but it is still complaining.

.github/workflows/dockerized-test.yml
darklight3it: BTW I think the existing `test-rie` way of dealing with those tests will be removed in favor how the harness test standard that easily allow you to make assertions on the lambda response.

GradExMachina:
There's already docker infra scripts written by Antonia.

darklight3it:
I know but this is doing something different. It's integrating this stuff in the harness test GHA created by @maxday .

The idea is to port every RIC to use the same test framework.

Beside that this one also allows to use examples as task to be run inside the harness test framework.

GradExMachina:
I don't think that is needed as we already do e2e tests on PR. I understand the need for uniformity, but I think other runtimes should move to this model, rather than introducing a language nobody wants to learn.

For rust in particular, we can inject a custom client into the runtime and get this without dealing with additional tooling, and we can also test the client by itself with just a custom server — this server simulator code is already available.

Ultimately, not my decision, but I think this is a waste of time and resources.

maxday:
Thanks for your feedback @FullyTyped !
We believe there is a need for harness testing, as it could quickly detect some issues without having to deploy to a real Lambda ecosystem. It allows for a fast feedback loop during development.
We don't use a custom server, but the RIE, also maintained by AWS: https://github.com/aws/aws-lambda-runtime-interface-emulator

.github/workflows/dockerized-test.yml
```diff
@@ -0,0 +1,56 @@
...
+  pull_request:
+    branches: [ '*' ]
+  workflow_dispatch:
+
+
+jobs:
+  dockerized-test:
+    runs-on: ubuntu-latest
+    env:
+      ACTIONS_STEP_DEBUG: true
```
maxday: do we need this?

.github/workflows/dockerized-test.yml
```diff
@@ -0,0 +1,56 @@
...
+
+    steps:
+    - uses: actions/checkout@v6
+
+    - uses: dtolnay/rust-toolchain@stable
+    - uses: Swatinem/rust-cache@v2
+
+    - name: Set up Python 3.14
```
maxday: I think you can now remove this as the action is now Dockerized, no need for python anymore

.github/workflows/dockerized-test.yml
```diff
@@ -0,0 +1,56 @@
...
+          set +a
+          echo "HANDLERS_TO_BUILD=${HANDLERS_TO_BUILD}" >> $GITHUB_ENV
+          echo "OUTPUT_DIR=${OUTPUT_DIR}" >> $GITHUB_ENV
+        fi
+    
+    - name: Build Lambda artifacts for testing
+      run: |
+        mkdir -p test/dockerized/tasks
+        OUTPUT_DIR="$(pwd)/test/dockerized/tasks" make build-examples
+        ls -la test/dockerized/tasks/
```
maxday, line 36: should we remove this?

PR_DESCRIPTION.md
```diff
@@ -0,0 +1,39 @@
...
+```bash
+EXAMPLES="basic-lambda basic-lambda-concurrent" make build-examples
+```
+
+🔏 *By submitting this pull request*
+
+- [x] I confirm that I've ran `cargo +nightly fmt`.
+- [x] I confirm that I've ran `cargo clippy --fix`.
+- [x] I confirm that I've made a best effort attempt to update all relevant documentation.
+- [x] I confirm that my contribution is made under the terms of the Apache 2.0 license.on.
```
rarepolz: nit: Looks like a duplicate checkbox + stray .on. at the end of the license line.

.github/workflows/dockerized-test.yml
darklight3it: no, removing debug

.github/workflows/dockerized-test.yml
darklight3it: you're right, forgot it

GradExMachina:
I wrote the server in this crate in other tests because it was far easier than using RIE and harness tests, and it allows you to evaluate the actual state of the runtime programmatically in a white-box approach.

There is a reason I did that way when I wrote those tests, Max.

~ Quinn 

Let's maybe not assume ignorance next time we talk to contributors :)

.github/workflows/dockerized-test.yml
darklight3it: No this is important, I need to build the binaries in some ways and put them where it is needed for the action to be taken. In this way the action is woring exactly as in Ruby but I'm taking binaries instead of source file.

PR_DESCRIPTION.md
darklight3it: Sorry removing it from the PR.

maxday:
I meant we don't use custom server in the GitHub action as we use the RIE sorry if that was not clear. 

Also I'm definitely not assuming anything, I'm simply explaining why we are introducing the RIE and providing a link for those who aren't familiar with it and might find this PR later.

rarepolz (approved):

PR #1105, opened 2026-02-23
feat(lambda-runtime): log non-2xx Lambda Runtime API responses with status and body

darklight3it:
## Summary

When the Lambda Runtime API returns a non-2xx response (e.g. HTTP 410 with `{"errorMessage":"Invoke timeout","errorType":"InvokeTimeout"}`), the previous implementation silently discarded the response body. Customers had no visibility into why their handler response was rejected.

## Solution

Read and log the response body for non-2xx responses from the Lambda Runtime API.

The main challenge is that `RuntimeApiClientFuture` is a pub enum without` #[non_exhaustive]`, so adding a new variant would be a breaking change for any downstream code that exhaustively matches on it. To avoid this, we instead spawn a background task via `tokio::task::spawn` to read the body asynchronously without modifying the enum's public surface.

## Changes

- `RuntimeApiClientService` now captures a `tokio::runtime::Handle` at construction time via `Handle::try_current()`, with a fallback minimal current-thread runtime for non-tokio contexts (e.g. tests)
- On non-2xx responses, a background task is spawned to collect and log the response body including the HTTP status code
- Added a `log_or_print!` macro in `lib.rs` that dispatches to tracing::error! when a tracing dispatcher is set, and falls back to `eprintln!` otherwise — consistent with the existing codebase pattern
- Added unit tests covering 2xx success, 410 InvokeTimeout, and inner service errors, using `httpmock`

## Example log output

```
ERROR lambda_runtime::layers::api_client: Lambda Runtime API returned non-200 response status=410 body={"errorMessage":"Invoke timeout","errorType":"InvokeTimeout"}
```

### DIFF
--- a/lambda-runtime/src/layers/api_client.rs
+++ b/lambda-runtime/src/layers/api_client.rs
@@ -1,9 +1,11 @@
 use crate::LambdaInvocation;
 use futures::{future::BoxFuture, ready, FutureExt, TryFutureExt};
+use http_body_util::BodyExt;
 use hyper::body::Incoming;
 use lambda_runtime_api_client::{body::Body, BoxError, Client};
 use pin_project::pin_project;
 use std::{future::Future, pin::Pin, sync::Arc, task};
+use tokio::runtime::{Handle, Runtime};
 use tower::Service;
 use tracing::error;
 
@@ -16,11 +18,38 @@ use tracing::error;
 pub struct RuntimeApiClientService<S> {
     inner: S,
     client: Arc<Client>,
+    /// Handle to the tokio runtime used to spawn background tasks (e.g. logging error bodies).
+    /// Obtained from the ambient runtime at construction time, or from a self-managed fallback.
+    handle: Handle,
+    /// Keeps the self-managed fallback runtime alive for as long as this service exists.
+    /// `None` when constructed inside an existing tokio context (the common case).
+    /// Must not be dropped before `handle`, as dropping the runtime invalidates the handle.
+    fallback_runtime: Option<Arc<Runtime>>,
 }
 
 impl<S> RuntimeApiClientService<S> {
     pub fn new(inner: S, client: Arc<Client>) -> Self {
-        Self { inner, client }
+        let (handle, fallback) = match Handle::try_current() {
+            Ok(h) => (h, None),
+            Err(_) => {
+                // Fallback: build a minimal current-thread runtime and keep it alive alongside
+                // the service. This should only happen outside of a tokio context (e.g. some tests).
+                let rt = Arc::new(
+                    tokio::runtime::Builder::new_current_thread()
+                        .enable_all()
+                        .build()
+                        .expect("failed to build fallback tokio runtime"),
+                );
+                let h = rt.handle().clone();
+                (h, Some(rt))
+            }
+        };
+        Self {
+            inner,
+            client,
+            handle,
+            fallback_runtime: fallback,
+        }
     }
 }
 
@@ -52,13 +81,21 @@ where
         Self {
             inner: self.inner.clone(),
             client: self.client.clone(),
+            handle: self.handle.clone(),
+            fallback_runtime: self.fallback_runtime.clone(),
         }
     }
 }
 
+// Despite this enum not being in mod.rs, it is `pub` and not marked as `#[non_exhaustive]`.
+// This means that adding a new variant would be a breaking change for any downstream code
+// that exhaustively matches on it. To consume the error response body without introducing
+// a new variant, we instead spawn a background task via `tokio::task::spawn`.
 #[pin_project(project = RuntimeApiClientFutureProj)]
 pub enum RuntimeApiClientFuture<F> {
+    /// The inner service is building the HTTP request to send to the Lambda Runtime API.
     First(#[pin] F, Arc<Client>),
+    /// The HTTP request is in-flight; waiting for the Lambda Runtime API response.
     Second(#[pin] BoxFuture<'static, Result<http::Response<Incoming>, BoxError>>),
 }
 
@@ -79,7 +116,10 @@ where
                         let next_fut = client
                             .call(ok)
                             .map_err(|err| {
-                                error!(error = ?err, "failed to send request to Lambda Runtime API");
+                                log_or_print!(
+                                    tracing: error!(error = ?err, "failed to send request to Lambda Runtime API"),
+                                    fallback: eprintln!("Lambda Runtime API request failed: {err}")
+                                );
                                 err
                             })
                             .boxed();
@@ -87,8 +127,128 @@ where
                     }
                     Err(err) => break Err(err),
                 },
-                RuntimeApiClientFutureProj::Second(fut) => break ready!(fut.poll(cx)).map(|_| ()),
+                RuntimeApiClientFutureProj::Second(fut) => match ready!(fut.poll(cx)) {
+                    Ok(resp) if !resp.status().is_success() => {
+                        let status = resp.status();
+                        // Spawn a background task to read and log the error body without
+                        // blocking the current poll or requiring a new enum variant.
+                        Handle::current().spawn(async move {
+                            match resp.into_body().collect().await {
+                                Ok(body) => {
+                                    let body = body.to_bytes();
+                                    log_or_print!(
+                                        tracing: error!(status = %status, body = ?body, "Lambda Runtime API returned non-200 response"),
+                                        fallback: eprintln!("Lambda Runtime API returned non-200 response: status={status}, body={body:?}")
+                                    )
+                                }
+                                Err(e) => {
+                                    log_or_print!(
+                                        tracing: error!(status = %status, error = ?e, "Lambda Runtime API returned non-200 response (body unreadable)"),
+                                        fallback: eprintln!("Lambda Runtime API returned non-200 response (body unreadable): status={status}, error={e:?}")
+                                    )
+                                }
+                            }
+                        });
+                        break Ok(());
+                    }
+                    Ok(_) => break Ok(()),
+                    Err(err) => break Err(err),
+                },
             }
         })
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::types::Context;
+    use bytes::Bytes;
+    use httpmock::prelude::*;
+    use std::future;
+    use tower::Service;
+
+    fn make_invocation() -> LambdaInvocation {
+        let (parts, _) = http::Response::new(()).into_parts();
+        LambdaInvocation {
+            parts,
+            body: Bytes::new(),
+            context: Context::default(),
+        }
+    }
+
+    fn make_client(server: &MockServer) -> Arc<Client> {
+        let uri = format!("http://{}:{}", server.host(), server.port()).parse().unwrap();
+        Arc::new(Client::builder().with_endpoint(uri).build().unwrap())
+    }
+
+    fn mock_inner(
+        req: http::Request<Body>,
+    ) -> impl Service<
+        LambdaInvocation,
+        Response = http::Request<Body>,
+        Error = BoxError,
+        Future = impl Future<Output = Result<http::Request<Body>, BoxError>>,
+    > {
+        tower::service_fn(move |_inv: LambdaInvocation| {
+            let req = http::Request::builder()
+                .uri(req.uri().clone())
+                .method(req.method().clone())
+                .body(Body::empty())
+                .unwrap();
+            future::ready(Ok::<_, BoxError>(req))
+        })
+    }
+
+    #[tokio::test]
+    async fn test_2xx_response_succeeds() {
+        let server = MockServer::start();
+        server.mock(|when, then| {
+            when.any_request();
+            then.status(200).body("ok");
+        });
+
+        let client = make_client(&server);
+        let req = http::Request::builder().uri("/some/path").body(Body::empty()).unwrap();
+        let mut svc = RuntimeApiClientService::new(mock_inner(req), client);
+
+        let result = svc.call(make_invocation()).await;
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_invoke_timeout_410_is_logged_and_succeeds() {
+        // Simulates the Lambda Runtime API returning 410 with an InvokeTimeout body,
+        // e.g.: {"errorMessage":"Invoke timeout","errorType":"InvokeTimeout"}
+        // The runtime should treat this as a non-fatal event: log it and return Ok(()).
+        let server = MockServer::start();
+        server.mock(|when, then| {
+            when.any_request();
+            then.status(410)
+                .header("Content-Type", "application/json")
+                .body(r#"{"errorMessage":"Invoke timeout","errorType":"InvokeTimeout"}"#);
+        });
+
+        let client = make_client(&server);
+        let req = http::Request::builder().uri("/some/path").body(Body::empty()).unwrap();
+        let mut svc = RuntimeApiClientService::new(mock_inner(req), client);
+
+        // Non-2xx should still resolve Ok(()) — the error body is logged in a background task.
+        let result = svc.call(make_invocation()).await;
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_inner_service_error_propagates() {
+        let server = MockServer::start();
+        let client = make_client(&server);
+
+        let failing_inner = tower::service_fn(|_: LambdaInvocation| {
+            future::ready(Err::<http::Request<Body>, BoxError>("inner error".into()))
+        });
+        let mut svc = RuntimeApiClientService::new(failing_inner, client);
+
+        let result = svc.call(make_invocation()).await;
+        assert!(result.is_err());
     }
 }
--- a/lambda-runtime/src/lib.rs
+++ b/lambda-runtime/src/lib.rs
@@ -19,6 +19,17 @@ use tokio_stream::Stream;
 use tower::util::ServiceFn;
 pub use tower::{self, service_fn, Service};
 
+/// Logs using tracing `error!` if a dispatcher is set, otherwise falls back to `eprintln!`.
+macro_rules! log_or_print {
+    (tracing: $tracing_expr:expr, fallback: $fallback_expr:expr) => {
+        if tracing::dispatcher::has_been_set() {
+            $tracing_expr;
+        } else {
+            $fallback_expr;
+        }
+    };
+}
+
 /// Diagnostic utilities to convert Rust types into Lambda Error types.
 pub mod diagnostic;
 pub use diagnostic::Diagnostic;


lambda-runtime/src/layers/api_client.rs
```diff
@@ -56,13 +58,17 @@ where
     }
 }
 
-#[pin_project(project = RuntimeApiClientFutureProj)]
-pub enum RuntimeApiClientFuture<F> {
-    First(#[pin] F, Arc<Client>),
-    Second(#[pin] BoxFuture<'static, Result<http::Response<Incoming>, BoxError>>),
+#[pin_project(project = RuntimeApiClientFutureStateProj)]
+pub enum RuntimeApiClientFutureState<F> {
```
jlizen: This will be a breaking change. It is an internal type, that probably nobody is using, but it is part of our public API.

If we did want to go this route, I'd suggest making a new one, and marking the old one as deprecated. And cutting over all usage.

Given that probably nobody is using it, the new type is fairly painless. I don't mind that option.

That would only require a minor version bump (the deprecated old enum).

jlizen: It wouldn't be breaking if we had tagged it `#[non_exhaustive]`, but we missed this one.

Definitely if we make a new type, we should tag it non exhaustive :)

darklight3it: sorry I was already fixing that. It's still `RuntimeApiClientFutureState`. The CR was still a draft.

darklight3it: we changed the internals ase we said in the other CR. Cannot make it pub (crate) from api_client but still it isn't exposed in mod.rs

jlizen: I discussed this with you and rcoh@ offline - 

Technically this is a breaking change, whether we change the innards of `RuntimeApiClientFuture` or substitute it for a different enum. The type is accessible as an associated type on the `RuntimeApiClientService` which is publicly visible.

None of our internal machinery relies on this specific type surface - we are just polling the service that the runtime api client exposes. I'm pretty sure whoever implemented this just didn't realize that `pub enum` exposed all of its variants and field contents.

It would generally be risky to try to deploy something in the runtime api client, that the `lambda-runtime` needs to be upgraded in lockstep to, else they will have compile errors. And then that's fragile to do except across a breaking version boundary.

BUT - we aren't in that situation, the only people that would be impacted would be people that built their own higher level runtime on top of our own, or otherwise are doing things with the innards of the runtime client.

I've confirmed via [code search](https://github.com/search?q=RuntimeApiClientFuture&type=code) that the only cases of `RuntimeApiClientFuture` being used are forks of the api client.

So, with that in mind, I think this is probably a fairly low risk change to make.

jlizen: And then, the lowest risk change available would be to just keep the current future name, since then anybody that is naming the future as part of a tower service stack, wouldn't be broken.

And with the future, it is safer to add a new field to the existing variant, rather than add a new variant. Since that won't break match statements.

Unfortunately, that doesn't get us out of this mess for next time, but it is the least risky.

jlizen: That said - what field to add - well, we already have the contents we want inside the `RuntimeApiClientFuture`, the issue is that we can't poll it to completion from inside the runtime (specifically, the body field of the `http::Response` future stored inside Second)

For that, we really would prefer a third state transition which is riskier than a new field .

I guess, we could also just change the second variant's contents to an enum (which is clearly sillier than a third variant, but less breaking-prone maybe?)

jlizen: Maybe instead we could do something hacky like, check the status code, and bubble up the inner future inside of the `BoxError`, and downcast the `BoxError` from inside our managing runtime to decide whether to finish polling the body?

And then, another option we discussed was just storing a client current thread runtime or something to drive the future, but that's pretty awful... but we don't currently require that this be run inside a tokio runtime, so i don't think we can rely on `tokio::spawn`.

I guess we could also do something with a sync thread and `block_on`, but that is pretty terrible :)

darklight3it: I went with the `tokio::spawn` idea, the bubbling up of the error required a changing in then `RuntimeApiClientFuture` signature that while respecting the current enum and variant structure would have changed the signature constituting a breaking change.

This seems to be the only way for this to work without any signature change. The integration tests are working and the test I added are working too.

lambda-runtime/src/layers/api_client.rs
```diff
@@ -16,11 +18,38 @@ use tracing::error;
...
 
 impl<S> RuntimeApiClientService<S> {
     pub fn new(inner: S, client: Arc<Client>) -> Self {
-        Self { inner, client }
+        let (handle, fallback) = match Handle::try_current() {
+            Ok(h) => (h, None),
+            Err(_) => {
+                // Fallback: build a minimal current-thread runtime and keep it alive alongside
+                // the service. This should only happen outside of a tokio context (e.g. some tests).
+                let rt = Arc::new(
```
jlizen, line 37: Is there a reason that the rt needs to be inside an arc? It is stored inside a struct that doesn't have `Clone` anyway

lambda-runtime/src/layers/api_client.rs
```diff
@@ -79,16 +116,139 @@ where
...
                     }
                     Err(err) => break Err(err),
                 },
-                RuntimeApiClientFutureProj::Second(fut) => break ready!(fut.poll(cx)).map(|_| ()),
+                RuntimeApiClientFutureProj::Second(fut) => match ready!(fut.poll(cx)) {
+                    Ok(resp) if !resp.status().is_success() => {
+                        let status = resp.status();
+                        // Spawn a background task to read and log the error body without
+                        // blocking the current poll or requiring a new enum variant.
+                        Handle::current().spawn(async move {
```
jlizen, line 135: This will panic if the future is awaited outside of a tokio runtime. It isn't actually using the handle we store on the client.

That said, there isn't a GREAT way to pass through the runtime handle to the future without changing the public signature, since we are stuck with an enum with wide open variants and inner fields. We would essentially need to add the handle to both the first and second, and have first plumb it through to second. IE add a field to both variants.

The main non breaking way I can think of would be to store the runtime in a static [OnceLock](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) (which you initialize inside `RuntimeApiClientService::new()`, and it stores the runtime and returns the handle via `get_or_init`. So nothing stored on the client itself anymore.

At that point, you won't need to carry any data in the future. And the runtime is something of a global resource, so it's not too strange..?

lambda-runtime/src/layers/api_client.rs
```diff
@@ -79,16 +116,139 @@ where
...
+                                }
+                                Err(e) => {
+                                    log_or_print!(
+                                        tracing: error!(status = %status, error = ?e, "Lambda Runtime API returned non-200 response (body unreadable)"),
+                                        fallback: eprintln!("Lambda Runtime API returned non-200 response (body unreadable): status={status}, error={e:?}")
+                                    )
+                                }
+                            }
+                        });
+                        break Ok(());
```
jlizen, line 152: why are these breaks needed? i would have expected we can just have eg `Ok(_) => Ok(())`

lambda-runtime/src/layers/api_client.rs
```diff
@@ -79,16 +116,139 @@ where
...
+                    }
+                    Ok(_) => break Ok(()),
+                    Err(err) => break Err(err),
```
jlizen, line 155: just to confirm, we are already logging the connection related error higher up? It looked like the C++ impl was logging them all at the same site (though it's fine if we handle differently, with the same ultimate logs)

lambda-runtime/src/lib.rs
```diff
@@ -19,6 +19,17 @@ use tokio_stream::Stream;
 use tower::util::ServiceFn;
 pub use tower::{self, service_fn, Service};
 
+/// Logs using tracing `error!` if a dispatcher is set, otherwise falls back to `eprintln!`.
+macro_rules! log_or_print {
```
jlizen, line 23: Cute, can we cut over [this usage](https://github.com/aws/aws-lambda-rust-runtime/blob/c565705ddec966ba2f51964829ce023729f2b6a6/lambda-runtime/src/runtime.rs#L366) as well?

lambda-runtime/src/layers/api_client.rs
```diff
@@ -79,16 +116,139 @@ where
...
+        let client = make_client(&server);
+        let req = http::Request::builder().uri("/some/path").body(Body::empty()).unwrap();
+        let mut svc = RuntimeApiClientService::new(mock_inner(req), client);
+
+        let result = svc.call(make_invocation()).await;
+        assert!(result.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_invoke_timeout_410_is_logged_and_succeeds() {
```
jlizen, line 220: This isn't actually asserting that anything is logged, we probably want to use `#[traced_test]` to reflect against log output

lambda-runtime/src/layers/api_client.rs
jlizen: One thing we might need to do would be, add a check to see if a caller-provided tokio runtime is still alive, since it might shut down early and cause a panic? Not sure how much we need to guard against that... Will get a second opinion

rcoh: is the only thing you need to do here log the message? Rather than doing it this way, can you send the `resp` over a channel to someone else to read it out and print it?

One "dumb" option is to use the response.extensions_mut() to get the channel (or runtime handle) in there.

That avoids the hazard of attempting to read it here from the thread local.

lambda-runtime/src/layers/api_client.rs
darklight3it: I know. I can add this.

lambda-runtime/src/layers/api_client.rs
darklight3it: Yes we only need to log the message.

jlizen: One concern is, the place we would want to store a runtime handle or channel receiver in order to [write into response extensions](https://github.com/aws/aws-lambda-rust-runtime/blob/c565705ddec966ba2f51964829ce023729f2b6a6/lambda-runtime-api-client/src/lib.rs#L60), is the [underlying runtime api client](https://github.com/aws/aws-lambda-rust-runtime/blob/c565705ddec966ba2f51964829ce023729f2b6a6/lambda-runtime-api-client/src/lib.rs#L31), which is also wide open as pub / non-exhaustive.

So it wouldn't necessarily sidestep the need for static state. 

A nicer option might be, instead use a [`tokio::task_local!`](https://docs.rs/tokio/latest/tokio/task/struct.LocalKey.html) at [the point where we are spawning the tokio tasks](https://github.com/aws/aws-lambda-rust-runtime/blob/8e6dc816553c5c09f73fe5461e408bc3a401c50f/lambda-runtime/src/runtime.rs#L216), to inject a `mpsc::Sender`. And then, we can listen on it in a `tokio::select!` loop alongside its work loop.

That reduces the hazard of of the thread local reads, right @rcoh ? It also has the side benefit of allowing us to enrich the logs with per-worker context, since we have a worker id span instrumented there.

And then, inside the `RuntimeApiClientFuture`, where we read the response from second, we can (perhaps behind `cfg(feature="concurrency-tokio")` check if `LocalKey::try_get` returns something, and if yes, write the response to the channel to let the worker-managed channel receive loop await the body and write out its log?

jlizen: I guess we could also just use `tokio::runtime::Handle::current()` inside the `lambda_runtime_api_client` gated behind `tokio_concurrency`  and drill that feature down? IE use `.map_ok` on the request future, get the handle, and shove it in response extensions, and then have the `RuntimeApiClientFuture` check extensions and spawn the task.

Feels more infectious though than just task_locals, and it still has some weird edge cases if anybody is instantiating the `lambda_runtime_api_client::Client` directly, with the feature enabled, but not on a tokio runtime. (probably fine, but technically breaking if somebody has `--all-features` enabled)

lambda-runtime/src/layers/api_client.rs
darklight3it: `Arc` is necessary because:

1. `RuntimeApiClientService` does implement `Clone`, and it needs to be cloneable for concurrent workers (line [222](https://github.com/aws/aws-lambda-rust-runtime/blob/c565705ddec966ba2f51964829ce023729f2b6a6/lambda-runtime/src/runtime.rs#L222) in runtime.rs). 

2. The `Handle` becomes invalid if the `Runtime` is dropped.  `Arc` ensures the runtime stays alive as long as any service instance exists.

3.  Multiple cloned services share the same fallback runtime via reference counting instead of creating separate runtimes.

GradExMachina:
Why not do this in the actual client a-la the PR I was working on? That will be effectively the same pattern as the Managed Runtimes do it, by delegating to the C++ client, which will then log. 

That means we won't be breaking anyone along the way.

jlizen:
> That will be effectively the same pattern as the Managed Runtimes do it, by delegating to the C++ client, which will then log.

We could accomplish this, but it would be a bit awkward since it would require that we use `tokio::runtime::Handle::current()` directly, which in turn probably needs to be behind a `tokio_concurrency` feature gate since we don't force tokio usage in general.  And even then, we might want to guard against non-tokio-runtime-contexts calling the client with `--all-features` enabled.

I'm not against it though, necessarily.

Interested in your thoughts about this thread @FullyTyped : https://github.com/aws/aws-lambda-rust-runtime/pull/1105#discussion_r2848371896

GradExMachina:
Why would we need to do that? 

I am under the impression this is universal and doesn't need us to hide it under concurrency. 


IIRC the client impl doesn't depend on the tolio runtime at all.

darklight3it:
> Why would we need to do that?

> I am under the impression this is universal and doesn't need us to hide it under concurrency.
on this I agree with @FullyTyped . The logging of the error should happen also in the non multiconcurrent path.

However I want to point out that `tokio::spawn` in this context is a workaround because we cannot change in any way `RuntimeApiClientFuture` without breaking existing code.

jlizen:
> Why would we need to do that?

This is attempting to sidestep a breaking change in the `RuntimeApiClientFuture` by adding a new variant.

> The logging of the error should happen also in the non multiconcurrent path

Ah, good to know... in that case `tokio::spawn` based solutions will be pretty awkward since we don't want to assume usage of tokio runtime for non-LMI path. Really we would prefer modeling things in a single future state machine, like Quinn is pointing out.

That said, I think we could possibly just move this entire check downwards into the runtime api client, like you might have been alluding to Quinn? Since we are already working with a `BoxFuture` on the `Client`'s side. So as long as we cast whatever information we want to transmit upwards into a `BoxError`, we could use future combinators?

It would go [here](https://github.com/aws/aws-lambda-rust-runtime/blob/c565705ddec966ba2f51964829ce023729f2b6a6/lambda-runtime-api-client/src/lib.rs#L60)

And look something like:
```rust

        self.client.request(req).map_err(Into::into)
            .then(|res| async move { 
                match res {
                    Ok(resp) if !resp.status().is_success() => {
                        let status = resp.status();
                        match resp.into_body().await {
                            // convert into a BoxError containing the status and optional body
                           // which is nicely loggable in the upper layer, or you could also log it here.
                           // If you need to specifically log ONLY this case rather than other BoxErrors, you could newtype
                          // a marker type that we downcast the stderror to
                           Err(BoxError::new(format!("my message with status code + body")))
                        }
                    } else {
                        res
                    }

                }
            })
            .boxed()
```

Does that seem like it might work?

GradExMachina:
I think that this should be within the state machine.

Because all we do here is logging and apparently dataplane gives us a status code, we don't need to change the flow of the state machine. We will know that we time'd out by the status code, and therefore we don't need to consume more of the response (and thus add another state).

jlizen:
If we can get away with checking the status only, then awesome, that makes this whole problem much simpler! I was assuming there was something important in the body. Much nicer to just have a log on non-2xx status when consuming the `http:: Response ` inside the existing second state.

darklight3it:
We can definitely get away reading only the status code for now and logging a specific message. But this is not the functional equivalent of what we are doing in other runtimes where we log also the body.

If you agree to this I would go on with the simple solution and open a new issue about the visibility about `RuntimeApiErrorFuture`.

At some point we would need to set things straight deprecating that and fixing the issue.

jlizen:
Would the `then` combinator suggestion that smuggles the body in a `BoxError` be that much uglier? It seems like a relatively simple way to propagate the body upwards without needing manage a spawned task. The main downside is that it would await inside the foreground task (which probably is fine...?)

No objections to going status only though here, y'all have more context about expectations for cross language runtime parity.

lambda-runtime/src/layers/api_client.rs
darklight3it: The break is needed to exit the loop. Without it, Ok(()) would just be a value expression and the loop would continue infinitely. The break exits the loop and provides its result value, which gets wrapped in task::Poll::Ready(...).

darklight3it:
>Would the then combinator suggestion that smuggles the body in a BoxError be that much uglier? It seems like a relatively simple way to propagate the body upwards without needing manage a spawned task. The main downside is that it would await inside the foreground task (which probably is fine...?)
No objections to going status only though here, y'all have more context about expectations for cross language runtime parity.

I think that the the `then` solution is probably an implicit contract breaking solution. We are  not changing the signature, but we are significantly changing the behavior of `Client`.

Returning `ERR` every time we are getting a non 200 status code. It's true that also `Client` is probably not used that much, and this is an implicit contract change, but we had high standards so far on what is breaking.

I think the most safe solution is taking only the status code. Also we can open an issue abou this.

jlizen:
> Returning ERR every time we are getting a non 200 status code. It's true that also Client is probably not used that much, and this is an implicit contract change, but we had high standards so far on what is breaking.

This is not necessarily true, we could downcast the `BoxError` inside the `RuntimeApiClientFuture` handling to accomplish logging, and still cast to success from caller's view. (Unfortunately we can't log directly inside the `then` and without at least emitting a marker `BoxError` that we know to skip, since we can't reconstruct the `Response` for the `OK()` variant. So we can't cast back to `Ok(T)` after consuming the body.

Something like:

```rust
struct ClientResponseErrorWithStatusAndMaybeBody((StatusCode, Option<String>));
// inside the `Client`, use `.then()` to check status and await the body if non-success, and cast to the box error with this struct in it

// inside the `RuntimeApiClientFuture` second variant handling
                RuntimeApiClientFutureProj::Second(fut) => match ready!(fut.poll(cx)) {
                   ....
                    Err(err) => {
                         if let Some(response_err) = err.downcast::<ClientResponseErrorWithStatusAndMaybeBody>() {
                                // log it, return success
                         }
                         return Err(err)
                   }
```

That said, I have no objections to logging only status. Up to you if you care to cut an issue for logging the body down the road. If yes, I think this approach would work.

darklight3it:
Closed in favor of this CR

https://github.com/aws/aws-lambda-rust-runtime/pull/1109

PR #1112, opened 2026-02-26
test(all)  Add concurrent Lambda invocation tests

darklight3it:
- **chore: add pre commit script for anyone that wants to use it.**
- **chore: adding new tests**

📬 *Issue #, if available:*

✍️ *Description of changes:*

This PR adds end-to-end concurrent testing for the `basic-lambda-concurrent` example using the Lambda Managed Instance (LMI) runtime mode.

### What changed

**GitHub Actions (`.github/workflows/dockerized-test.yml`)**
- Added `dockerized-test-concurrent` job that builds only `basic-lambda-concurrent`, creates a Docker network, and runs the concurrent scenarios via the action

**New example behavior (`examples/basic-lambda-concurrent/src/main.rs`)**
- Added a concrete `HandlerError` type with `From<HandlerError> for Diagnostic`, giving clean `errorType: "HandlerError"` in error responses instead of the mangled type name from `type_name_of_val`
- Added startup log for `AWS_LAMBDA_MAX_CONCURRENCY` value
- Handler returns `Err(HandlerError(...))` for `{"command": "fail"}` payloads

**New test scenarios (`test/dockerized/scenarios/concurrent_scenarios.py`)**
- `concurrent_happy_path`: fires 10 unique commands concurrently, asserts all succeed
- `concurrent_error_isolation`: fires 9 failing requests + 1 valid survivor concurrently, asserts the survivor still succeeds despite the errors

**Makefile**
- Extracted shared Docker setup into `build-test-runner` target
- Added `test-dockerized-concurrent` target with Docker network setup for DinD
- `TEST_RUNNER_BRANCH` variable (defaulting to `main`, overridable via `.env`) controls which branch of the test runner is cloned

**IMPORTANT**
- To make this work I'm running on a custom version of containerized test. We need some more work upstream to make multiconcurrent testing compatible with what we are trying to do.


Makefile
```diff
@@ -179,6 +196,7 @@ help: ## Show this help message
 	@echo '                        Usage: HANDLERS_TO_BUILD="basic-lambda" HANDLER="basic-lambda" make test-rie'
 	@echo '  test-rie-lmi          Test RIE in Lambda Managed Instance mode'
 	@echo '                        Usage: RIE_MAX_CONCURRENCY=4 HANDLERS_TO_BUILD="basic-lambda-concurrent" make test-rie-lmi'
+	@echo '  test-dockerized-concurrent Run concurrent LMI test scenarios'
```
jlizen: nit: formatting, the rest of the lines can push right

.env
```diff
@@ -11,3 +11,6 @@ OUTPUT_DIR=test/dockerized/tasks
 
 # Max concurrent Lambda invocations for LMI mode
 RIE_MAX_CONCURRENCY=4
+
+# Branch of containerized-test-runner-for-aws-lambda to clone
+TEST_RUNNER_BRANCH=dmelfi/multi-concurrency-test
```
jlizen: is this intended to be checked in? is there a way to get this into main instead?

.github/workflows/dockerized-test.yml
```diff
@@ -44,4 +44,37 @@ jobs:
...
+        ls -la test/dockerized/tasks/
+
+    - name: Build base test image with RIE and custom entrypoint
+      run: docker build -t local/test-base -f Dockerfile.test .
+
+    - name: Create Docker network for concurrent tests
+      run: docker network create concurrent-test-net
+
+    - name: Run concurrent scenarios
+      uses: aws/containerized-test-runner-for-aws-lambda@dmelfi/multi-concurrency-test
```
jlizen: same feedback, it would be better to get this over to main when possible

darklight3it: yeah I know. Unfortunately I am missing some features in that repo. And since it's something we would like to check in more detail I opted for this solution. @maxday knows about this.

.env
darklight3it: moved to main.

.github/workflows/dockerized-test.yml
darklight3it: moved to main.

Makefile
darklight3it: fixed

Issue #1029, opened 2025-08-18
[lambda_http] Add support for VPC Lattice lambda target groups


andrewpatto:
VPC Lattice can target lambdas - with events that are 95% the same as the existing support for API gateway.

It would be good to also support Lattice.

https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html

Related to #1028

darklight3it:
I think we can close this.

andrewpatto:
This was specifically about adding in support for VPC Lattice into the lambda-http codebase - unless I am missing a PR I don't think that has been done?

andrewpatto:
> We are able to handle requests from:
> [API Gateway](https://docs.aws.amazon.com/apigateway/latest/developerguide/welcome.html) REST, HTTP and WebSockets API lambda integrations
> AWS [ALB](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/introduction.html)
> AWS [Lambda function URLs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-urls.html)

Given the similarity to api gateway - I was preparing a PR that mirrors that for VPC Lattice (using the event structures that just got added)

jlizen:
Still glad to help review the lattice impl whever you have time, @andrewpatto  :)

andrewpatto:
No probs - is 90% done sitting on my computer. Just need to broaden out the test cases. Soon!

PR #1136, opened 2026-04-08
Add VPC Lattice as an upstream source for lambda-http

andrewpatto:
📬 *Issue #, if available:*

#1029 

✍️ *Description of changes:*

Added VPC Lattice support (and test cases) to lambda-http.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 This package makes it easy to run AWS Lambda Functions written in Rust. This workspace includes multiple crates:
 
 - [![Docs](https://docs.rs/lambda_runtime/badge.svg)](https://docs.rs/lambda_runtime) **`lambda-runtime`** is a library that provides a Lambda runtime for applications written in Rust.
-- [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library that makes it easy to write API Gateway proxy event focused Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library for writing HTTP Lambda functions in Rust, with support for upstream API Gateway, ALB, Lambda Function URLs, and VPC Lattice.
 - [![Docs](https://docs.rs/lambda-extension/badge.svg)](https://docs.rs/lambda-extension) **`lambda-extension`** is a library that makes it easy to write Lambda Runtime Extensions in Rust.
 - [![Docs](https://docs.rs/aws_lambda_events/badge.svg)](https://docs.rs/aws_lambda_events) **`lambda-events`** is a library with strongly-typed Lambda event structs in Rust.
 - [![Docs](https://docs.rs/lambda_runtime_api_client/badge.svg)](https://docs.rs/lambda_runtime_api_client) **`lambda-runtime-api-client`** is a shared library between the lambda runtime and lambda extension libraries that includes a common API client to talk with the AWS Lambda Runtime API.
--- a/lambda-events/src/event/vpc_lattice/v2.rs
+++ b/lambda-events/src/event/vpc_lattice/v2.rs
@@ -1,7 +1,4 @@
-use crate::{
-    custom_serde::{deserialize_headers, deserialize_nullish, http_method, serialize_multi_value_headers},
-    encodings::Body,
-};
+use crate::custom_serde::{deserialize_headers, deserialize_nullish, http_method, serialize_multi_value_headers};
 #[cfg(feature = "builders")]
 use bon::Builder;
 use http::{HeaderMap, Method};
@@ -47,7 +44,7 @@ pub struct VpcLatticeRequestV2 {
 
     /// The request body
     #[serde(default)]
-    pub body: Option<Body>,
+    pub body: Option<String>,
 
     /// Whether the body is base64 encoded
     #[serde(default, deserialize_with = "deserialize_nullish")]
--- a/lambda-http/README.md
+++ b/lambda-http/README.md
@@ -2,7 +2,9 @@
 
 [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http)
 
-**`lambda-http`** is an abstraction that takes payloads from different services and turns them into http objects, making it easy to write API Gateway proxy event focused Lambda functions in Rust.
+**`lambda-http`** is an abstraction that takes HTTP-shaped payloads from different AWS services and turns them
+into standard `http` objects, making it easy to write Lambda functions in Rust that serve HTTP traffic
+regardless of the upstream trigger.
 
 lambda-http handler is made of:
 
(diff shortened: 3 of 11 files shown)


lambda-events/src/event/vpc_lattice/v2.rs
```diff
@@ -47,7 +44,7 @@ pub struct VpcLatticeRequestV2 {
 
     /// The request body
     #[serde(default)]
-    pub body: Option<Body>,
+    pub body: Option<String>,
```
andrewpatto, line 47: This was a mistake by me in the original event PR.  This makes the type consistent with all other request event types. The Body abstraction is meant for responses (I think)

lambda-http/README.md
```diff
@@ -2,7 +2,9 @@
 
 [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http)
 
-**`lambda-http`** is an abstraction that takes payloads from different services and turns them into http objects, making it easy to write API Gateway proxy event focused Lambda functions in Rust.
+**`lambda-http`** is an abstraction that takes HTTP-shaped payloads from different AWS services and turns them
+into standard `http` objects, making it easy to write Lambda functions in Rust that serve HTTP traffic
+regardless of the upstream trigger.
```
andrewpatto, line 7: clarifying the overall purpose of the library - I feel it is more centered on HTTP shape payloads and not API gateway proxy events..

README.md
```diff
@@ -5,7 +5,7 @@
 This package makes it easy to run AWS Lambda Functions written in Rust. This workspace includes multiple crates:
 
 - [![Docs](https://docs.rs/lambda_runtime/badge.svg)](https://docs.rs/lambda_runtime) **`lambda-runtime`** is a library that provides a Lambda runtime for applications written in Rust.
-- [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library that makes it easy to write API Gateway proxy event focused Lambda functions in Rust.
+- [![Docs](https://docs.rs/lambda_http/badge.svg)](https://docs.rs/lambda_http) **`lambda-http`** is a library for writing HTTP Lambda functions in Rust, with support for upstream API Gateway, ALB, Lambda Function URLs, and VPC Lattice.
```
andrewpatto, line 8: clarify the top-level description of what the library does (but happy if this is tidied/shortened). Possibly does not need to enumerate all upstream services.

andrewpatto:
I have not attempted to fix the CHANGELOG entries for either library as I was unsure exactly when/how these are done.

jlizen:
Awesome!! Adding to my review queue, might take a moment

lambda-events/src/event/vpc_lattice/v2.rs
darklight3it: Looking at the public documentation ([here](https://docs.aws.amazon.com/vpc-lattice/latest/ug/lambda-functions.html#receive-event-from-service)) I don't think it was a mistake — base64 payloads are entirely possible for VPC Lattice events.

The `Body` abstraction is meant to automatically handle content in requests and responses that can be both base64 or string. 

I understand that `lambda-http` handles the conversion automatically when bridging to `http::Request<Body>`, but I would prefer the event struct to be declarative and represent the real possible types of the payload. 

This has also the added benefit of simplifying the code in `request.rs`.

@jlizen wdyt?

darklight3it (commented):
Thanks for your contribution ❤️ . 

I left only one comment about your choice of changing the type. Let's dicuss that before approval.

lambda-http/README.md
jlizen: good improvement

lambda-events/src/event/vpc_lattice/v2.rs
jlizen: Andrew is correct here - unfortunately, we don't have support for parsing maybe-base64 incoming requests via the `Body` type, since it only implements `visit_str` and always produces `Body::Text(string)`. 

The old behavior was arguably wrong - it will never actually deserialize into `Body::Binary`, so it is a more complicated, misleading version of simply `String`. Sadly serde doesn't let you compare sibling fields, and some sort of "attempt to base64 decode and fallback" is highly fragile.

I think the risk of breakage is low (Andrew added this type directly for usage in this PR, unlikely to affect other consumers) and it is very misleading currently. So, we can consider it a bug and fix it.

It would be great to add a helper that had nicer handling:
```
  impl VpcLatticeRequestV2 {
      pub fn decoded_body(&self) -> Option<Body> {
          self.body.as_deref().map(|b|
  Body::from_maybe_encoded(self.is_base64_encoded, b))
      }
  }
```
But that's somewhat separate scope than this change.

We should also probably update the `Body` documentation to be clearer about this usage (it is for SERIALIZING agnostically, not deserializing).

jlizen: In other words: I'm aligned with changing the event struct to use `Option<String>` as a bugfix, with optional additions:
- fix the doc comments on `Body` to be clearer about usage with deserializing vs serializing
- helper on VpcLatticeRequestV2 to do the sibling-based decoded body, which we could use in `lambda-http` conversion if desired

README.md
jlizen: Seems good

lambda-http/src/request.rs
```diff
@@ -773,6 +825,189 @@ mod tests {
         );
     }
 
+    #[test]
```
jlizen, line 829: these tests need gating behind cfg(feature = "vpc_lattice")

lambda-http/src/request.rs
```diff
@@ -337,6 +349,41 @@ fn into_websocket_request(ag: ApiGatewayWebsocketProxyRequest) -> http::Request<
...
+                .map_or_else(Body::default, |b| Body::from_maybe_encoded(base64, b)),
+        )
+        .expect("failed to build request");
+
+    // no builder method that sets headers in batch
+    let _ = std::mem::replace(req.headers_mut(), vlr.headers);
+    let _ = std::mem::replace(req.method_mut(), http_method.unwrap_or(http::Method::GET));
+
+    req
+}
```
jlizen, line 386: nit: whitespace

lambda-http/src/request.rs
jlizen: needs the vpc_lattice cfg flag too

jlizen (commented):
LGTM overall. I agree with the doc changes. Commented on the discussion of `Body`->`String`.

Can we also have tests for:
- method: `None` defaults to GET
- POST with a body
- VPC lattice events aren't deserialized as api gateway v2
The two's shapes overlap significantly, it's not immediately apparent if there is a conflicting structural constraint or not

A relevant test would be something like:
```
  #[test]
  fn vpc_lattice_event_does_not_match_apigw_v2() {
      let data =
  include_bytes!("../../lambda-events/src/fixtures/example-vpc-la
  ttice-v2-request.json");
      let result =
  serde_json::from_slice::<ApiGatewayV2httpRequest>(data);
      assert!(result.is_err(), "VPC Lattice event should not
  deserialize as APIGW V2");
  }
```

(That said, I dislike this approach for deserializing things in order ^^ This library is a bit crufty... but no need for scope creep here)

lambda-events/src/event/vpc_lattice/v2.rs
darklight3it: Ok I think the helper method is a good compromise.

darklight3it (approved):

jlizen:
Cool, sounds like we are aligned on approach! Whenever you have time @andrewpatto , my only blocking concerns are:
- fixing the config gating on tests
- having the test that the vpc lattice event actually is being handled as such rather than api gateway (otherwise we'll want to re-order the deserialization cascade)

The rest are nice-to-haves.

jlizen:
@andrewpatto i don't mind throwing a quick commit on top to address those testing concerns, if you're busy

andrewpatto:
Hi yes @jlizen - happy for you to commit on top. I spend most of my days in python genomics code - so sometimes finding the time to fire up the rust environment for this gets away from me (despite best intentions). Thanks everyone for the reviews.

jlizen:
@andrewpatto glad to, can you turn on "maintainers are allowed to edit this pull request"?

andrewpatto:
> @andrewpatto glad to, can you turn on "maintainers are allowed to edit this pull request"?

For some reason I'm not seeing that option on this PR??

andrewpatto:
Apparently because the fork comes from an organisation owned fork - that option does not appear. I given you enough permissions into our org fork repo @jlizen that it may work now

jlizen (approved):
Thanks for the contribution!

Does this fully close out #1029 ?

andrewpatto:
> Thanks for the contribution!
> 
> Does this fully close out #1029 ?

Yes.. thanks

Issue #1137, opened 2026-04-14
Missing Builder derive for MillisecondTimestamp?


DirkRusche-Otto:
I was trying to instantiate a `KafkaRecord` and the builder exposes a `timestamp` method that lets me pass a `MillisecondTimestamp`. But unfortunately I couldn't find a way to construct the `MillisecondTimestamp` struct (without using Deserialize - which feels a bit awkward to use). 

So I guess there is a `#[cfg_attr(feature = "builders", derive(Builder))]` missing?

jlizen:
Yeah, looks like they are needed here:
https://github.com/aws/aws-lambda-rust-runtime/blob/main/lambda-events/src/encodings/time.rs

Another option is to just chain through chrono setters for `from_systemtime_and_offset` or other ergonomic constructors. Technically the non exhaustive signals that we might add future fields, but that seems unlikely to me, and they would surely be optional.

PR welcome!

DirkRusche-Otto:
Thanks for your quick response. I can prepare a PR, can you give me more precise instructions what solution you would prefer?

jlizen:
We definitely want `derive(Builder)`, like you suggested.

If you find the ergonomics for setting it poor, I don't mind `MillisecondTimestamp::from_systemtime_and_offset()` as a supplementary change. Up to you if you want to add it, I'd accept it.

If you don't mind adding the same for builder / optionally the direct constructor to the other types (`SecondTimestamp`, etc), I think they need it too.

DirkRusche-Otto:
Here you go https://github.com/aws/aws-lambda-rust-runtime/pull/1138

PR #1138, opened 2026-04-16
chore: add builder derive for time structs

DirkRusche-Otto:
📬 *Issue #, if available: #1137*

✍️ *Description of changes:*

Added `Builder` derive for the time structs (`MillisecondTimestamp`, `SecondTimestamp`, `SecondDuration`, `MinuteDuration`).

### DIFF
--- a/lambda-events/src/encodings/time.rs
+++ b/lambda-events/src/encodings/time.rs
@@ -7,7 +7,6 @@ use serde::{
 use std::ops::{Deref, DerefMut};
 
 /// Timestamp with millisecond precision.
-#[non_exhaustive]
 #[derive(Clone, Default, Debug, Deserialize, Eq, PartialEq, Serialize)]
 pub struct MillisecondTimestamp(
     #[serde(deserialize_with = "deserialize_milliseconds")]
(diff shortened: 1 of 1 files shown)


jlizen (approved):
Much appreciated! Will try to get a release out this week

jlizen (changes requested):
Oof, I missed that these are tuple structs, which bon doesn't support.

Given the types here, I am comfortable just removing the `#[non_exhaustive]` entirely. That addresses your use case, right @DirkRusche-Otto ?

I don't see a scenario where we will need to add new fields. It was likely add in a bulk change without looking closely.

DirkRusche-Otto:
> Oof, I missed that these are tuple structs, which bon doesn't support.
> 
> Given the types here, I am comfortable just removing the `#[non_exhaustive]` entirely. That addresses your use case, right @DirkRusche-Otto ?
> 
> I don't see a scenario where we will need to add new fields. It was likely add in a bulk change without looking closely.

Done.

jlizen (approved):
Thanks!

DirkRusche-Otto:
> Much appreciated! Will try to get a release out this week

Thank you again for getting to a solution so quickly. Very appreciated, that unblocks us now from upgrading to the newest version.

PR #1143, opened 2026-05-06
fix(lambda-http): into_api_gateway_v2_request joins cookies with "; "

### LINKED ISSUE #1142: `into_api_gateway_v2_request` joins cookies with `";"` instead of `"; "`, violating RFC 6265 and breaking strict parsers
## Describe the bug
 
`into_api_gateway_v2_request` reconstructs the `Cookie` header by joining the `cookies` array from the API Gateway v2 event payload with `";"` (no space). This violates RFC 6265 §4.2.1, which mandates `"; "` (semicolon followed by a space) as the cookie-pair separator. The result is a non-compliant `Cookie` header that breaks any downstream cookie parser that follows the spec strictly [like `BetterAuth`](https://github.com/better-auth/better-auth/issues/9465).
 
## Affected code
 
[`lambda-http/src/request.rs`](https://github.com/aws/aws-lambda-rust-runtime/blob/e397be4ab03645fec8474ee8a26461ff035dd051/lambda-http/src/request.rs)
 
```rust
if let Some(cookies) = ag.cookies {
    if let Ok(header_value) = HeaderValue::from_str(&cookies.join(";")) {
        headers.insert(http::header::COOKIE, header_value);
    }
}
```
 
The join produces:
 
```
cookie1=value1;cookie2=value2
```
 
instead of the RFC-compliant:
 
```
cookie1=value1; cookie2=value2
```
 
Note: the broken behavior is also **explicitly asserted in the test suite**, meaning the non-compliant output is enforced as correct:
 
```rust
assert_eq!(cookie_header, Ok("cookie1=value1;cookie2=value2"));
```
 
This test assertion would need to be updated alongside the fix.
 
## RFC 6265 reference
 
[RFC 6265 §4.2.1](https://www.rfc-editor.org/rfc/rfc6265.html) defines the
grammar as:
 
```
cookie-header = "Cookie:" OWS cookie-string OWS
cookie-string = cookie-pair *( ";" SP cookie-pair )
```
 
`SP` is a mandatory single space (`%x20`). The `";"` separator without a following space is not a valid cookie-string per the spec.
 
## Real-world impact
 
I discovered this issue while deploying a **Next.js application on AWS Lambda behind API Gateway v2**, using
[`aws-lambda-web-adapter`](https://github.com/aws/aws-lambda-web-adapter).
 
The [`better-auth`](https://github.com/better-auth/better-auth) library's `getSessionCookie` helper parses the raw `Cookie` header by splitting on `"; "`. Because `lambda_http` joins cookies with `";"` only, the entire cookie string is treated as a single unparseable token, causing `getSessionCookie` to return `null` even when the session cookie is present.
  
I have submitted a corresponding fix to `better-auth` to make their parser lenient (https://github.com/better-auth/better-auth/issues/9465), but the root cause is in `lambda_http`: the header should be constructed correctly here.
 
## Expected behavior
 
The `Cookie` header reconstructed from `ag.cookies` should join pairs with `"; "` to produce a spec-compliant value:
 
```
cookie1=value1; cookie2=value2
```
 
## Environment
 
- `lambda_http`: current HEAD (`e397be4`)
- Trigger: AWS API Gateway HTTP API (v2 payload format)
- Runtime: AWS Lambda (Node.js via `aws-lambda-web-adapter`, but the bug is in the Rust `lambda_http` crate itself)
- Framework: Next.js 16 (App Router)

sbougerel:
`into_api_gateway_v2_request` reconstructs the Cookie header by joining the cookies array from the API Gateway v2 event payload with `";"` (no space). This violates RFC 6265 §4.2.1, which mandates `"; "` (semicolon followed by a space) as the cookie-pair separator. The result is a non-compliant Cookie header that breaks any downstream cookie parser that follows the spec strictly like BetterAuth.

📬 *Issue #, if available:*

Closes https://github.com/aws/aws-lambda-rust-runtime/issues/1142

✍️ *Description of changes:*

The `Cookie` header reconstructed from `into_api_gateway_v2_request` now joins pairs with `"; "` to produce a spec-compliant value:
 
```
cookie1=value1; cookie2=value2
```

The corresponding test was updated.

### DIFF
--- a/lambda-http/src/request.rs
+++ b/lambda-http/src/request.rs
@@ -147,7 +147,7 @@ fn into_api_gateway_v2_request(ag: ApiGatewayV2httpRequest) -> http::Request<Bod
 
     let mut headers = ag.headers;
     if let Some(cookies) = ag.cookies {
-        if let Ok(header_value) = HeaderValue::from_str(&cookies.join(";")) {
+        if let Ok(header_value) = HeaderValue::from_str(&cookies.join("; ")) {
             headers.insert(http::header::COOKIE, header_value);
         }
     }
@@ -570,7 +570,7 @@ mod tests {
 
         assert_eq!(req.method(), "POST");
         assert_eq!(req.uri(), "https://id.execute-api.us-east-1.amazonaws.com/my/path?parameter1=value1&parameter1=value2&parameter2=value");
-        assert_eq!(cookie_header, Ok("cookie1=value1;cookie2=value2"));
+        assert_eq!(cookie_header, Ok("cookie1=value1; cookie2=value2"));
 
         // Ensure this is an APIGWv2 request
         let req_context = req.request_context_ref().expect("Request is missing RequestContext");


darklight3it (approved):
LGTM 🚀

jlizen (approved):

PR #1154, opened 2026-07-19
docs: show custom event handler usage

### LINKED ISSUE #1153: docs: make the custom event object example in the README show handler usage
Ref: https://github.com/aws/aws-lambda-rust-runtime#custom-event-objects

Currently we show how to define a custom event object. But, our usage is not realistic, it just deserializes it in `main()`.

Instead we should show the pattern of actually passing the type to a handler, and printing the payload inside or something.

floze-the-genius:
## Summary
- replace the custom event JSON deserialization in `main` with a typed Lambda handler
- log the received payload and return the typed response

Fixes #1153

## Validation
- `git diff --check`


jlizen (approved):
Wow, you're quick! Appreciate the contribution.

PR #1159, opened 2026-07-30
feat: add lambda-runtime-invocation-id header

### LINKED ISSUE #1155: Add Lambda-Runtime-Invocation-Id header support for cross-wiring protection
## Summary

Lambda now sends a `Lambda-Runtime-Invocation-Id` header on `/runtime/invocation/next` responses. Runtimes should echo this header back on `/runtime/invocation/{requestId}/response` and `/runtime/invocation/{requestId}/error` to enable RAPID to detect and reject stale responses from timed-out invocations.

## Problem

When an invoke times out, the runtime process continues running. If a new invoke arrives with the same `requestId` (customer-provided or retried by upstream), RAPID accepts it. The still-running old invocation eventually posts its response, and RAPID delivers the wrong response to the new invoke (cross-wiring).

This affects both On-Demand and Lambda Managed Instances (LMI), and Rust is a supported LMI runtime.

## What needs to change

1. **Parse** `Lambda-Runtime-Invocation-Id` from the `/next` response headers (optional — may be absent with older RAPID)
2. **Store** it alongside the invocation context
3. **Echo** it as a request header on `/response` and `/error`
4. If the header is absent from `/next`, don't send it back (backward compat)

## Backward Compatibility

- Old RAPID (doesn't send the header) + New RIC (no header to echo) → no change ✅
- New RAPID (sends the header) + Old RIC (doesn't echo) → RAPID skips validation ✅
- New RAPID + New RIC → header echoed, RAPID validates match ✅

## Reference Implementations

- Python RIC: https://github.com/aws/aws-lambda-python-runtime-interface-client/pull/214

darklight3it:
## Summary

Add `Lambda-Runtime-Invocation-Id` header support for cross-wiring protection.

The RIC now echoes the invocation ID received from RAPID on `/next` back on `/response` and `/error`, enabling RAPID to detect and reject stale responses from timed-out invocations.

## Problem

On Lambda Managed Instances (LMI) and On-Demand (OD), when an invoke times out, the runtime process continues running in the background. If a new invoke arrives with the same `requestId`, RAPID accepts it. The still-running old invocation eventually posts its response, and RAPID matches it to the new invoke — delivering the wrong response (cross-wiring).

## Solution

RAPID sends a unique per-invoke nonce via `Lambda-Runtime-Invocation-Id` header on `/next`. The runtime echoes it back on `/response` and `/error`. RAPID validates the match before accepting the response.

## Backward Compatibility

Fully backward compatible in both directions:
- If RAPID doesn't send the header → RIC doesn't see it → doesn't echo → no behavior change
- If RIC doesn't echo it (old version) → RAPID skips validation → no behavior change

### Rate-limited malformed-header logging

Added an internal  generic `RateLimiter` for runtime different uses. In this case the request was to rate limit warning caused by malformed  `Lambda-Runtime-Invocation-Id` headers.

The limiter:

* Uses a monotonic `Instant` and configurable `Duration`.
* Is protected by a standard mutex because the check-and-update operation must be atomic across concurrent Tokio tasks.
* Is process-local, so a static limiter is shared across warm invocations in  the same Lambda execution environment.
* Resets its timestamp and clears mutex poisoning if the mutex is poisoned, since the limiter state is disposable and must not crash the runtime.
* Keeps rate-limiting independent from `tracing`, allowing a future `tracing-subscriber` integration without coupling the runtime behavior to a  specific subscriber.

Malformed invocation-ID headers are ignored rather than converted with `from_utf8_lossy`, because the value may be echoed back as an HTTP header and must remain a valid ASCII header value. A rate-limited warning is emitted with the configured interval included as structured log metadata.

## Testing

- added unit test
- added a new dockerized test working with the new RIE.


lambda-runtime/src/types.rs
```diff
@@ -85,6 +92,11 @@ pub struct Context {
     /// Includes information such as the function name, memory allocation,
     /// version, and log streams.
     pub env_config: RefConfig,
+    /// The invocation ID assigned by the Lambda runtime for cross-wiring protection.
+    /// Echoed back on `/response` and `/error` to allow RAPID to reject stale responses
+    /// from timed-out invocations. `None` when running against older RAPID versions
+    /// that don't send this header.
+    pub invocation_id: Option<String>,
```
vip-amzn: shouldn't be used in context.

vip-amzn: https://github.com/aws/aws-lambda-python-runtime-interface-client/blob/main/awslambdaric/lambda_runtime_client.py#L170

Should be directly passed into invocation request

examples/invocation-id-concurrent/src/main.rs
```diff
@@ -0,0 +1,129 @@
...
+    }
+    Ok(())
+}
+
+pub(crate) async fn my_handler(event: LambdaEvent<Request>) -> Result<Response, HandlerError> {
+    if event.payload.sleep > 0 {
+        tokio::time::sleep(tokio::time::Duration::from_secs(event.payload.sleep.into())).await;
+    }
+
+    let resp = Response {
```
vip-amzn: Echo the marker only. The invocation-id should not be available here. it is
 internal RIC- RAPID id and the runtime echoes it automatically.
```
    Ok(Response {
        from: event.payload.command,
    })
```

And then 

```
async fn handler_echoes_marker() {
        let event = LambdaEvent {
            payload: Request { command: "invoke-B".into(), sleep: 0 },
            context: Context::default(),
        };
        let result = my_handler(event).await.unwrap();
        assert_eq!(result, Response { from: "invoke-B".into() });
    }
```

test/dockerized/scenarios/concurrent_scenarios.py
```diff
@@ -62,3 +71,28 @@ def get_concurrent_scenarios():
     ))
 
     return scenarios
+
+
+def get_invocation_id_scenarios():
```
vip-amzn, line 75: This is not testing the invocation id cross wiring situation. the assertion is on request id, which is going to be the same. But we want to check that response is coming appropriately.

```
batches = [
    [Request.create(
        payload={"command": "invoke-A", "sleep": TIMEOUT + 2},
        assertions=[{"transform": ".errorType", "error": "Sandbox.Timedout"}],
        headers={"X-Amzn-RequestId": SAME_REQUEST_ID},
    )],
    [Request.create(
        payload={"command": "invoke-B", "sleep": TIMEOUT - 1},
        assertions={"response": {"from": "invoke-B"}},   
        headers={"X-Amzn-RequestId": SAME_REQUEST_ID},
    )],
]
```

lambda-runtime/src/types.rs
darklight3it: you are right.

darklight3it: solved

examples/invocation-id-concurrent/src/main.rs
darklight3it: solved

vip-amzn (approved):

lambda-runtime/src/requests.rs
```diff
@@ -129,7 +138,12 @@ where
                 let body = serde_json::to_vec(&body)?;
                 let body = Body::from(body);
 
-                let req = build_request().method(Method::POST).uri(uri).body(body)?;
+                let mut req = build_request().method(Method::POST).uri(uri).body(body)?;
+
+                if let Some(id) = self.invocation_id {
+                    req.headers_mut().insert(LAMBDA_RUNTIME_INVOCATION_ID, id.parse()?);
```
jlizen, line 144: This will error out and crash the runtime if a malformed invocation id header is sent.

We originally decode with `String::from_utf8_lossy()`, but that replaces non-utf8 bytes with `U+FFFD` which is anyway not ascii. So then this parse will fail.

I did a quick check and didn't find any other round trips, this new code is the only place impacted.

I know we control the sender but we should be defensive against malformed inputs anyway. I would suggest a rate-limited log warning if we have bad bytes.

examples/invocation-id-concurrent/src/main.rs
```diff
@@ -0,0 +1,96 @@
...
+// { "command": "do something" }
+
+use lambda_runtime::{service_fn, tracing, Diagnostic, Error, LambdaEvent};
+use serde::{Deserialize, Serialize};
+
+#[derive(Deserialize)]
+struct Request {
+    #[serde(rename = "command")]
+    _command: String,
+    sleep: u32,
```
jlizen, line 11: This should have `#[serde(default)]`, otherwise the example command in the header will actually have a deserialization failure.

test/dockerized/scenarios/concurrent_scenarios.py
jlizen: this is stale now i think?

test/dockerized/scenarios/concurrent_scenarios.py
```diff
@@ -9,8 +9,11 @@
...
 
 HANDLER = "basic-lambda-concurrent"
+INVOCATION_ID_HANDLER = "invocation-id-concurrent"
 IMAGE = os.environ.get("TEST_IMAGE", "local/test-base")
+SAME_REQUEST_ID = "shared-request-id"
 DEFAULT_CONCURRENCY = 10
+TIMEOUT = 5
 
 
 def _make_env(concurrency: int = DEFAULT_CONCURRENCY) -> dict:
```
jlizen: this should accept a handler input too, right now it makes it seem like everything is the basic concurrent

lambda-runtime/src/requests.rs
jlizen, line 144: nit: we use `insert()` here but `append()` on the streaming branch, I would make them consistent

lambda-runtime/src/layers/api_response.rs
```diff
@@ -183,11 +200,47 @@ where
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{constants::LAMBDA_RUNTIME_INVOCATION_ID, runtime::LambdaInvocation, Context};
+    use http::{HeaderValue, Response};
+    use serde_json::json;
+    use tower::{service_fn, Service};
+
+    #[tokio::test]
+    async fn forwards_invocation_id_from_next_response_headers() {
```
jlizen, line 235: It would be good to have a test covering the deserialization error path as well

jlizen (changes requested):
Core approach is good, but some small tweaks.

Also: currently we log `Lambda function timeout!` for any 410, but we now will have a 410 if a stale response is rejected. We should tweak the message.

test/dockerized/scenarios/concurrent_scenarios.py
jlizen: IE, there are multiple lambda handlers now?

examples/invocation-id-concurrent/src/main.rs
darklight3it: done

lambda-runtime/src/layers/api_response.rs
darklight3it: done

lambda-runtime/src/requests.rs
darklight3it: done

darklight3it: Done! I also added a general `rate_limiter` class. I made it very generic so it can be reused in the runtime.

We can probably open an issue to see if we can use it in other places. And evaluate integration with logging mechanism.

Another thing I would think about is to offer this capabilities to our reexported logging to offer the possibility for the customer to use our own logging and a very convenient rate limiter.

test/dockerized/scenarios/concurrent_scenarios.py
darklight3it: done

test/dockerized/scenarios/concurrent_scenarios.py
darklight3it: done

lambda-runtime/src/requests.rs
jlizen: We can do it without locking, refer to: https://github.com/dial9-rs/dial9/blob/64b8e9ff6af3fde0e13cbcd31b6e7c6b5891913f/dial9-core/src/rate_limit.rs#L15

(thought we already had this, sorry)

I would recommend against re-exporting this functionality externally as it is a bit hacky and outside the lambda-runtime's domain, I would rather see this land in `tracing` if we wanted it.

jlizen: But yeah I would just copy over dial9's approach wholesale

examples/invocation-id-concurrent/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+lambda_runtime = { path = "../../lambda-runtime", features = ["concurrency-tokio"] }
+serde = "1.0.219"
+tokio = { version = "1", features = ["macros", "rt", "time"] }
+
+[dev-dependencies]
+serde_json = "=1.0.151"
```
jlizen: is there a reason for the exact bound? a bit odd

lambda-runtime/src/layers/api_response.rs
```diff
@@ -123,9 +127,31 @@ where
...
+        // that don't send this header.
+        let invocation_id = match req.parts.headers.get(LAMBDA_RUNTIME_INVOCATION_ID) {
+            Some(value) => match value.to_str() {
+                Ok(value) => Some(value.to_owned()),
+                Err(error) => {
+                    rate_limited!(MALFORMED_INVOCATION_ID_LOG_INTERVAL, {
+                        warn!(
+                            error = ?error,
+                            rate_limit_interval_ms = MALFORMED_INVOCATION_ID_LOG_INTERVAL.as_millis(),
+                            "Ignoring malformed Lambda runtime invocation ID header; this warning is rate limited"
```
jlizen: nit: i might not include "this warning is rate limited"

jlizen (approved):

examples/invocation-id-concurrent/Cargo.toml
darklight3it: no, fixing it

lambda-runtime/src/layers/api_response.rs
darklight3it: ok

PR #1164, opened 2026-08-20
Add Kafka partial batch failure response types

FalkWoldmann:
📬 *Issue #, if available:*

✍️ *Description of changes:*
Add support to handle batch item failures for kafka trigger. See https://docs.aws.amazon.com/lambda/latest/dg/kafka-retry-configurations.html for more info

### DIFF
--- a/lambda-events/src/event/kafka/mod.rs
+++ b/lambda-events/src/event/kafka/mod.rs
@@ -55,6 +55,90 @@ pub struct KafkaRecord {
     pub other: serde_json::Map<String, Value>,
 }
 
+/// `KafkaEventResponse` is the outer structure to report batch item failures for `KafkaEvent`.
+#[non_exhaustive]
+#[cfg_attr(feature = "builders", derive(Builder))]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct KafkaEventResponse {
+    pub batch_item_failures: Vec<KafkaBatchItemFailure>,
+    /// Catchall to catch any additional fields that were present but not explicitly defined by this struct.
+    /// Enabled with Cargo feature `catch-all-fields`.
+    /// If `catch-all-fields` is disabled, any additional fields that are present will be ignored.
+    #[cfg(feature = "catch-all-fields")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "catch-all-fields")))]
+    #[serde(flatten)]
+    #[cfg_attr(feature = "builders", builder(default))]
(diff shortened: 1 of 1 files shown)


lambda-events/src/event/kafka/mod.rs
```diff
@@ -55,6 +55,90 @@ pub struct KafkaRecord {
...
+/// `KafkaItemIdentifier` identifies a Kafka record for a partial batch response.
+#[non_exhaustive]
+#[cfg_attr(feature = "builders", derive(Builder))]
+#[derive(Debug, Default, Clone, Eq, PartialEq, Deserialize, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct KafkaItemIdentifier {
+    /// The topic-partition key from the Kafka event's `records` map.
+    pub partition: String,
+    /// The Kafka record offset.
+    pub offset: i64,
```
darklight3it, line 131: `partition` and `offset` must match the record that failed in the current Kafka batch. An invalid topic-partition key or offset can cause Lambda to retry the entire batch, so please document the expected format—for example, partition: "topic-3"—and clarify that offset must be the original record offset.

https://docs.aws.amazon.com/lambda/latest/dg/kafka-retry-configurations.html#kafka-partial-batch-response

lambda-events/src/event/kafka/mod.rs
```diff
@@ -55,6 +55,90 @@ pub struct KafkaRecord {
...
+        self.batch_item_failures.push(KafkaBatchItemFailure {
+            item_identifier,
+            ..Default::default()
+        });
+    }
+
+    /// Set all failed Kafka item identifiers in the batch response.
+    ///
+    /// This replaces any previously registered failures.
+    pub fn set_failures<I>(&mut self, item_identifiers: I)
```
darklight3it, line 91: The method documentation says that `set_failures` replaces previously registered failures, but the current test only calls it on an empty response. Could you update the test to add an initial failure first, call set_failures, and verify that only the new failures remain?

lambda-events/src/event/kafka/mod.rs
```diff
@@ -68,4 +152,54 @@ mod test {
         let reparsed: KafkaEvent = serde_json::from_slice(output.as_bytes()).unwrap();
         assert_eq!(parsed, reparsed);
     }
+
+    #[test]
+    #[cfg(feature = "kafka")]
+    fn kafka_event_response_serializes_item_identifiers() {
```
darklight3it, line 158: Could you add a test that parses the expected Kafka partial-batch response JSON and verifies the resulting `KafkaEventResponse`? A serialization round-trip test would also help catch incorrect serde field names or types.

PR #1171, opened 2026-09-19
Add #[serde(default)] to `to` field of SimpleEmailCommonHeaders

Supreeeme:
✍️ *Description of changes:*
An SES event can be missing the `to` field, and without this can cause deserialization to fail. I can reproduce this, for example, by sending an email from Outlook to SES where all the recipients are only in the Bcc field.


lambda-events/src/event/ses/mod.rs
```diff
@@ -136,6 +136,7 @@ pub struct SimpleEmailHeader {
 #[serde(rename_all = "camelCase")]
 pub struct SimpleEmailCommonHeaders {
     pub from: Vec<String>,
+    #[serde(default)]
```
darklight3it: Nice catch! Can you please add coverage for your own case so this is documented?

PR #1172, opened 2026-09-20
Document AWS-LC jitter entropy cold-start cost

### LINKED ISSUE #1101: Consider documenting AWS_LC_SYS_NO_JITTER_ENTROPY
Context: https://github.com/aws/aws-lc-rs/issues/899

Docs: https://aws.github.io/aws-lc-rs/resources.html#entropy-configuration

Recent versions of `AWS_LC` enable a new entropy source, CPU jitter, which is somewhat expensive to initialize. It adds ~50ms of latency to startup in their testing, though I've seen signs of a decent amount higher on small instance sizes.  CPU jitter is enabled by default, but can be opted out of.

Since lambdas are frequently cold-start latency sensitive, this might be relevant to our userbase. For context, java lambdas using snapstart, have it disabled by default, based on a built in AWS LC check.

We don't directly use AWS_LC, but it is widely used by lambda applications as aws-sdks default to using it as the TLS backend (along with reqwest and much of the rest of the async ecosystem). So, we can expect most lambdas to be impacted by this behavior.

The workaround is to inject an environment variable via eg your `.cargo/config.toml`:
```
[env]
AWS_LC_SYS_NO_JITTER_ENTROPY=1
```

The implication of disabling this is, you would only have 1-2 other entropy sources (OS source + CPU source). CPU source availability varies based on environment. You generally want >= 2 entropy sources, hence the new default.

However, on casual examination, I think pretty much any hardware that aws lambda runs on, should anyway have cpu source? (I have not looked deeply though, some might not, this needs more research). 

It seems like this is something our users would appreciate knowing about. Not sure exactly where though. Perhaps it belongs in the aws-side lambda docs instead?

danielgaskins:
📬 *Issue #, if available:*

Closes #1101

✍️ *Description of changes:*

Adds a short note beside the build instructions about AWS-LC's CPU jitter entropy collection and its effect on Lambda cold starts.

The note shows the build-time setting for opting out, names the minimum `aws-lc-sys` version, and explains the security tradeoff. It also links to the AWS SDK announcement and AWS-LC's configuration documentation for readers who want the full story.

Validation:

- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- Verified both documentation links return HTTP 200

No Rust tests were run because this only changes the README.


jlizen (approved):
Appreciate it!

